PR #35, opened 2023-12-26
Fix mod compatibility check

krmeet:
When checking mod compatibilities, the original strings are used and in turn they are modified by strtok_r.
This PR makes it use copies of the strings instead.

(there is probably a better way to do this but i know next to nothing about c)


Agent-11:
Thank you

krmeet:
accident :3

krmeet:
making a new pr for dev branch

PR #39, opened 2023-12-28
Improve Lua error handling

krmeet:
This fixes script error handling and makes Lua logs more readable by removing the C source file names and long timestamps.
Previously, any kind of Lua error would output a tostring error. Load errors would only output the tostring error, runtime errors would output the actual error AND the tostring error.


Agent-11:
Did you mean to make this a draft PR?

krmeet:
yeah just making sure i haven't completely broken anything

PR #40, opened 2023-12-28
force render rooms, effectively disabling the room system

Dominicentek:
self explanatory


Agent-11:
... Why?

eros71-dev:
This will cause clipping issues with the castle inside, and possibly HMC, not sure about BBH but may not in that specific case
imagen

Isaac0-dev:
don't change the level geos, better to just disable the functionality in `geo_switch_area` if the level is a vanilla level.
also @eros71-dev  your linked image doesn't work

eros71-dev:
I guess GitHub didn't like me, it was supposed to load just fine according to the preview, but basically you can see part of the wall "pillars" next to the stairs in the center of the castle inside when entering the way to go to the basement.

eros71-dev:
imagen

PR #62, opened 2024-01-21
Added a custom hud font and a recolorable version and added dynos support to the aliased font

xLuigiGamerx:
image

All custom characters were drawn by me (yes every single one):

image

I am planning to make a menu font selector after this (when I figure it out), but for now, enjoy a recolorable hud font with all unicodes that the other fonts have :D


Agent-11:
Looks good

PR #75, opened 2024-02-25
fixed two interp functions not saving their resolution

Cooliokid956:
`djui_hud_render_texture_tile_interpolated()` and `djui_hud_render_rect_interpolated()` would not save their current resolution, resulting in interpolated textiles and rects rendering at larger or smaller scales than expected

### DIFF
--- a/src/pc/djui/djui_hud_utils.c
+++ b/src/pc/djui/djui_hud_utils.c
@@ -517,6 +517,7 @@ void djui_hud_render_texture_tile_interpolated(struct TextureInfo* texInfo, f32 
     interp->width = texInfo->width;
     interp->height = texInfo->height;
     interp->z = savedZ;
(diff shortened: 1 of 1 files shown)


Agent-11:
Peachy.

Agent-11:
I'm surprised you made this PR, I thought you would hate me after that commit I made on Character Select.

Cooliokid956:
the pain wore off the next day 😌

Agent-11:
I'm sorry about that, I shouldn't have slandered you publicly like that. Even though I thought it was funny and went hard, I also think it was ultimately petty of me. I guess I was just excited to be a contributor to CS and put it towards something Not That Great. So, I apologize.

Agent-11:
I don't hate you, I just think you're hard to understand sometimes and according to squishy not the easiest to work with at times. either

PR #95, opened 2024-03-17
Fix Error With Curl.h on debian & ubuntu.

ThePerfectMario64:


Agent-11:
I forgot to add the curl dependency to the README, but there are more distros listed there besides Ubuntu & Debian. I will do this myself.

ThePerfectMario64:
Cool.

PR #113, opened 2024-03-30
Display native language name in languages menu

tulpenkiste:
Small QoL feature that displays the native language name within the language menu for those that don't know the English name for their language.

Draft as the implementation could be better.

Screenshots:
image
image


Agent-11:
This is indeed abysmal.

tulpenkiste:
Ok, improved it to be less abysmal (no more lots of else ifs), I think its good for review now.

Agent-11:
This is still not a good implementation imo... Not to mention the variable name `language_hell`.

tulpenkiste:
> This is still not a good implementation imo... Not to mention the variable name `language_hell`.

Yeah, I agree. It definitely needs some work. I'm going to re-draft this while I think of a better system for this.

AgentXLP:
Seems to have gone stale.

PR #126, opened 2024-04-21
Add hooks for attacking objects

### LINKED ISSUE #125: [suggestion] Hook for Object vs Object / Player vs Object
I'm creating a mod that requires these hooks, and it seems like there doesn't exist any. (I've also tried looking through interact hooks, they're too inconsistent)

Hex4dec:
Fixes #125.


Hex4dec:
This could use improvement, I'll probably make a hook with both Mario and an object as parameters. Don't merge this yet.

Hex4dec:
This seems pretty much ready now!

AgentXLP:
Added to the dev repo

PR #145, opened 2024-05-24
Add HOOK_BEFORE_MARIO_HEALTH_UPDATE

Mechstreme:
### Called once per frame before Mario's health is updated.
My mod requires these hooks to detect when mario's heal/hurt counter is above zero before these values are updated. closest options I could find for this are HOOK_BEFORE_PHYS_STEP and HOOK_ALLOW_HAZARD_SURFACE, but they don't always update every frame


AgentXLP:
What would the purpose of this one be?

Mechstreme:
it can be used to override or track Mario's health or the heal/hurt counters before they update
```lua
-- Tracks all of the frames where mario's hurt counter is active & prints the hurt counter in chat
---@param m MarioState
function before_mario_health_update(m)
    if m.hurtCounter == 0 then return end
    djui_chat_message_create("-"..m.hurtCounter)
end
hook_event(HOOK_BEFORE_MARIO_HEALTH_UPDATE,before_mario_health_update)
```
while such is possible with `HOOK_MARIO_UPDATE`/`HOOK_BEFORE_MARIO_UPDATE`, both hooks still are called _after_ mario's health is updated (even though `HOOK_BEFORE_MARIO_UPDATE` is called outside of execute_mario_action), which causes the tracking to be late by one frame which can be kind of inconvenient

Isaac0-dev:
`HOOK_BEFORE_MARIO_UPDATE` is called before mario's health is updated.

```
bhv_mario_update
│
├── HOOK_BEFORE_MARIO_UPDATE called
│
├── execute_mario_action
│   └── update_mario_health
│       └── proposed HOOK_BEFORE_MARIO_HEALTH_UPDATE called
│
└── HOOK_MARIO_UPDATE called
```
I don't see why you can't run your example Lua code with the hook `HOOK_BEFORE_MARIO_UPDATE`

AgentXLP:
Yeah I agree with Isaac

PR #176, opened 2024-07-06
Threaded Audio Support 

theclashingfritz:
Initial support for threaded audio, Won't work with EU/SH regions (But those are currently broken anyways)

Testing is needed to check for crashes


Isaac0-dev:
awesome i can listen to music on the crash screen

PR #180, opened 2024-07-07
Fixed Spanish grammar

supahfox:
Some strings are bad grammatically


supahfox:
For example: "expresiones de odio"  translates as "hate expressions" and not "hate speech"

AgentXLP:
Retarget to the `dev` branch please

supahfox:
done

eros71-dev:
Yeah some lines are definitely better with this one, I'm only worried about the "Intensidad de vibración" string, while it does sound better, I'm not sure if it fits entirely without being cut off, if it does though then yeah it's all for the better.

AgentXLP:
I think it'll fit

PR #185, opened 2024-07-08
Support IPv6 for hosting and joining

theclashingfritz:


Isaac0-dev:
what are the limitations of this?

theclashingfritz:
There shouldn't be any more limitations compared to normal - The only limitation will the OS itself (IE: Some Linux distros disable dual-stack hosting by default which means it will only use IPv6) but that's not something we can do much about without having two sockets.

PR #188, opened 2024-07-09
Fix Discord invites for paths with spaces

Flower35:
Reworked the platform-dependent executable paths code.

Testing required for cases when the game is installed in different directories, with Discord Activity enabled:
- Windows
    - simple paths (like `C:\Games\sm64coopdx\`)
    - paths with spaces and unicode characters (like `C:\users\Użytkownik\Desktop\My Games\Super Mario 64 Coop DX\`)
- Linux
    - simple paths (like `/home/user/Games/sm64coopdx/`)
    - paths with spaces (like `/home/user/Desktop/My Games/Super Mario 64 Coop DX/`)
- macOS

I can only confirm that this compiles for Linux and Windows, but I cannot test on macOS and I cannot test the Discord Invites feature at the moment.

### DIFF
--- a/data/dynos.h
+++ b/data/dynos.h
@@ -29,7 +29,7 @@ extern "C" {
 #endif
 
 #define DYNOS_VERSION           "1.0"
-#define DYNOS_EXE_FOLDER        sys_exe_path()
+#define DYNOS_EXE_FOLDER        sys_exe_path_dir()
 #define DYNOS_USER_FOLDER       fs_get_write_path("")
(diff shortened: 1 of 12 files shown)


Isaac0-dev:
> I can only confirm that this compiles for Linux and Windows, but I cannot test on macOS and I cannot test the Discord Invites feature at the moment.

Can anyone check this?
We need to know:
- Do Discord invites work on this PR, specifically when coop is in a folder which has spaces in its name
- Does this work wtih macOS

PR #195, opened 2024-07-10
Add `djui_attempting_to_open_playerlist`

### LINKED ISSUE #175: Add hook to render after playerlist
Currently, there's no hook, or any way, to render on top of the playerlist. Having a way to do this would allow modders to replace/edit the playerlist, which would come in handy for certain mods.

EmeraldLoc:
Resolves #175

Since you can already disable the playerlist via `gServerSettings.enablePlayerList`, there was only one step to making #175 possible, that being allowing lua to detect if the playerlist was attempting to be opened, as outlined by Isaac.

I'm not fond of the implementation done here, however I couldn't think of a better implementation. Firstly, I create a global variable, `gAttemptingToOpenPlayerlist`. I create that variable in `djui_panel_playerlist.c` and define it in the corresponding .h file (could use a different location, since it isn't used in that c file). Then, in `smlua_misc_utils.c`, I create a function called `djui_attempting_to_open_playerlist` that just returns `gAttemptingToOpenPlayerlist`. This variable is set in the `djui_interactable_on_key_down ` and `djui_interactable_on_key_up` functions, where the playerlist visibility is also set.

Ideally there'd be a way I can detect what input is being pressed when the function `djui_attempting_to_open_playerlist` is being called, however I didn't find a way.

With this function, it's pretty easy to do what #175  was aiming to do. By first disabling the playerlist using `gServerSettings.enablePlayerList`, you can simply check if `djui_attempting_to_open_playerlist` is true. If it is, show the playerlist.

### DIFF
--- a/autogen/lua_definitions/functions.lua
+++ b/autogen/lua_definitions/functions.lua
@@ -8136,6 +8136,11 @@ function deref_s32_pointer(pointer)
 end
 
 --- @return boolean
+function djui_attempting_to_open_playerlist()
+    -- ...
+end
+
+--- @return boolean
 function djui_is_playerlist_open()
     -- ...
 end
--- a/docs/lua/functions-5.md
+++ b/docs/lua/functions-5.md
@@ -1953,6 +1953,24 @@
 
 <br />
 
+## [djui_attempting_to_open_playerlist](#djui_attempting_to_open_playerlist)
+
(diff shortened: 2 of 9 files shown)


AgentXLP:
Hope this doesn't have any strange side effects, though I doubt it will.

PR #198, opened 2024-07-11
Update sockets and structs for full IPv6 support

### LINKED ISSUE #172: Hosting does not listen on IPv6
**Describe the bug**
Hosting servers using Direct Connection only listens on IPv4, making hosting private servers impossible for those on ISP's that does IPv4 CGNAT.

**To Reproduce**
1. On Main Menu, click on "Host"
2. Set "Network System" to "Direct Connection"
3. Start Hosting

**Expected behavior**
Hosting should listen to both IPv4 and IPv6, or making the user choose which protocol to use on the hosting menu, defaulting to dual-stack IPv4+IPv6

**Screenshots**
listen

RoddyDev:
Resolves #172 
- This updates sockets to use AF_INET6 with dual-stack support instead of AF_INET, and all IP address structs from sockaddr_in to sockaddr_in6. That means IPv4 addresses are treated as IPv6-mapped (Example: ::ffff:192.168.1.69) for compatibility.

- Direct Connection input box was also updated for this change. It now supports direct IPv4, direct IPv6 addresses and hostnames. IPv6 addresses must be enclosed in square brackets [IPv6]:port to be able to separate IP from port, otherwise it will be treated as an IPv4 address or hostname. Would be better if IP and port had it's own input boxes, though.

- resolve_domain() now uses getaddrinfo() to get IPv4 and IPv6 addresses from a hostname, replacing the deprecated gethostbyname(). The only limitation in my code for resolve_domain() is that it still only grabs the first address it sees and returns. But getaddrinfo() is smart enough that it prioritizes IPv4 for those without IPv6 connectivity, so it shouldn't be a problem for now.

Trying to connect to this updated version from v1.0 connects normally on Windows, showing "Version mismatch".
Trying to connect to this updated version from v1.0 causes a crash on Linux, but it connects normally when using the same updated version.
The updated version was only tested on Linux. Couldn't get the compiler to work on Windows, even after following the compilation guide. Might need to get someone to test it for me on Windows before merging.


RoddyDev:
https://github.com/coop-deluxe/sm64coopdx/assets/45442056/7cddbc78-57b0-4d07-a22e-d59fa3f79f1e

Video of it in action (very low quality to fit in 10 MB)

RoddyDev:
Finally got the compiler to work on Windows and can confirm that it can connect normally to any server hosted on either Windows or Linux systems. I'm unable to test on macOS but it should also work the same way.

https://github.com/user-attachments/assets/ba6352e1-32b2-4cc9-bba9-5a4cd3870c8f

Here's a Windows client connecting to a server hosted on Linux.

EmeraldLoc:
Tested this on macOS, works fine!

RoddyDev:
> Tested this on macOS, works fine!

Thank you so much for testing.

RoddyDev:
I don't think there's anything else to add here anymore. Everything works perfectly in all platforms. The only requirement is that IPv6 needs to be enabled in your network adapter settings. (Does not require actual IPv6 connectivity + it's enabled by default in all operating systems today). Maybe we should add a fallback to AF_INET socket or show an error on screen if binding to an AF_INET6 socket fails (which means IPv6 is disabled on the system). Coopnet is untouched and also works perfectly normal.
I don't know if the edits in the translation files are really necessary, so feel free to edit these back to what it was if you want.
For now, I call this final. Waiting for further testing from the maintainers.

Isaac0-dev:
imo there should still be a way for ipv4 to be used, what is the current behavior if ipv6 fails?

RoddyDev:
> imo there should still be a way for ipv4 to be used, what is the current behavior if ipv6 fails?

When hosting a server using Direct Connection, it creates an AF_INET6 socket with IPV6_V6ONLY disabled so it does have IPv4+IPv6 support when binding to all addresses (in6addr_any). In my testings, neither Windows nor Linux complains when IPv6 is disabled on the network adapter and can still host.

When joining, getaddrinfo() returns an address depending if IPv6 connection is working or not. If it is, return IPv6 addresses first, if not, return IPv4 addresses first.

EmeraldLoc:
> neither Windows nor Linux complains when IPv6 is disabled on the network adapter and can still host.

can confirm macOS also doesn’t complain, and connects just fine.

AgentXLP:
You only have the English and Portuguese language translations, I'd need you to get translations for other languages before I can merge this.

RoddyDev:
Uploaded the translations. Cannot guarantee that the translations are accurate except English, Portuguese and German, because these three were translated from native speakers.

RoddyDev:
If the translations are the problem that this isn't being merged, I can just revert them to what it was before. I just wanted to let the user know that both protocols are supported.

AgentXLP:
I think I'll merge this now

PR #205, opened 2024-07-13
Update 4 example mods

Mechstreme:
Mario-Run.lua
  - Script was broken because it tried to check if mario's forward speed is below _and_ above a certain value; this is no longer the case

hud.lua
  - Added example functions for mouse position, rotation, filtering types, world-to-screen projection and rendering elements behind the hud via `HOOK_ON_HUD_RENDER_BEHIND`
  
water-level.lua
  - Now uses `get_water_level` and `set_water_level` instead of `get_environment_region` and `set_environment_region`
  - Added error handling for `on_set_command` to check if input is a number
  
Moonjump.lua
  - Updated a comment

### DIFF
--- a/docs/lua/examples/Mario-Run.lua
+++ b/docs/lua/examples/Mario-Run.lua
@@ -1,23 +1,16 @@
 -- name: Mario RUN!
 -- description: Mario is constantly running.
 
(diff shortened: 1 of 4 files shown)


AgentXLP:
Thank you sir

PR #208, opened 2024-07-14
Add rotation interpolation

Cooliokid956:
A new function, `djui_hud_set_rotation_interpolated`, has been added, which allows rotation and pivot point interpolation for interpolated DJUI elements. The function takes `s32`s for the angles instead of `s16`s to allow for wide rotations.

### DIFF
--- a/src/pc/djui/djui_hud_utils.c
+++ b/src/pc/djui/djui_hud_utils.c
@@ -415,7 +415,6 @@ void djui_hud_print_text_interpolated(const char* message, f32 prevX, f32 prevY,
     interp->z = savedZ;
     interp->resolution = sResolution;
     interp->rotation = sRotation;
-    
 }
 
 static inline bool is_power_of_two(u32 n) {
--- a/src/pc/djui/djui_hud_utils.c
+++ b/src/pc/djui/djui_hud_utils.c
@@ -27,7 +27,7 @@
 static enum HudUtilsResolution sResolution = RESOLUTION_DJUI;
 static enum HudUtilsFilter sFilter = FILTER_NEAREST;
 static enum DjuiFontType sFont = FONT_NORMAL;
-static struct HudUtilsRotation sRotation = { 0, 0, 0 };
+static struct HudUtilsRotation sRotation = { 0, 0, 0, 0, 0, 0 };
 static struct DjuiColor sColor = { 255, 255, 255, 255 };
 static struct DjuiColor sRefColor = { 255, 255, 255, 255 };
 static bool sLegacy = false;
@@ -101,6 +101,7 @@ struct InterpHud {
     f32 width;
     f32 height;
     enum HudUtilsResolution resolution;
+    struct HudUtilsRotation rotation;
 };
 static struct InterpHud sInterpHuds[MAX_INTERP_HUD] = { 0 };
 static u16 sInterpHudCount = 0;
@@ -114,6 +115,7 @@ void patch_djui_hud(f32 delta) {
     f32 savedZ = gDjuiHudUtilsZ;
     Gfx* savedHeadPos = gDisplayListHead;
     enum HudUtilsResolution savedResolution = sResolution;
+    struct HudUtilsRotation savedRotation = sRotation;
     for (u16 i = 0; i < sInterpHudCount; i++) {
         struct InterpHud* interp = &sInterpHuds[i];
         f32 x = delta_interpolate_f32(interp->prevX, interp->x, delta);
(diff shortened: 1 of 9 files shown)


src/pc/djui/djui_hud_utils.c
```diff
@@ -131,14 +134,29 @@ void patch_djui_hud(f32 delta) {
...
 
-        // translate scale
-        f32 translatedW = scaleW;
-        f32 translatedH = scaleH;
-        djui_hud_size_translate(&translatedW);
-        djui_hud_size_translate(&translatedH);
+       // rotate
+       f32 translatedW = scaleW;
+       f32 translatedH = scaleH;
+       s32 rotation = delta_interpolate_s32(sRotation.rotation-sRotation.rotationDiff, sRotation.rotation, delta);
```
Isaac0-dev: please use spaces in between the `-` for readability.

src/pc/djui/djui_hud_utils.c
```diff
@@ -131,14 +134,29 @@ void patch_djui_hud(f32 delta) {
...
+           f32 pivotTranslationX = translatedW * pivotX;
+           f32 pivotTranslationY = translatedH * pivotY;
+           create_dl_translation_matrix(DJUI_MTX_NOPUSH, +pivotTranslationX, -pivotTranslationY, 0);
+           create_dl_rotation_matrix(DJUI_MTX_NOPUSH, rotation, 0, 0, 1);
+           create_dl_translation_matrix(DJUI_MTX_NOPUSH, -pivotTranslationX, +pivotTranslationY, 0);
+       }
+
+        // scale
+     // djui_hud_size_translate(&translatedW);
+     // djui_hud_size_translate(&translatedH);
```
Isaac0-dev: don't leave this please, i assume this wasn't intentional

src/pc/djui/djui_hud_utils.c
```diff
@@ -386,6 +416,8 @@ void djui_hud_print_text_interpolated(const char* message, f32 prevX, f32 prevY,
     interp->height = font->defaultFontScale;
     interp->z = savedZ;
     interp->resolution = sResolution;
+    interp->rotation = sRotation;
+
```
Isaac0-dev: pointless new line?

src/pc/djui/djui_hud_utils.h
```diff
@@ -24,7 +24,10 @@ enum DjuiFontType {
 };
 
 struct HudUtilsRotation {
-    f32 rotation;
+    s32 rotationDiff;
+    f32 prevPivotX;
+    f32 prevPivotY;
+    s32 rotation;
```
Isaac0-dev: this order seems inconsistent. `s32 rotation;` maybe put it at the beggining of the struct just above `s32 rotationDiff;`

Isaac0-dev (commented):
its interesting that peachy made `rotation` a float, but it doesn't seem to be necessary because of the format of sm64 angles...
here are some intentionally nitpicky things that really don't matter.
only bother to make these changes if you actually want to, or someone requests that they are made

src/pc/djui/djui_hud_utils.c
AgentXLP: I agree

src/pc/djui/djui_hud_utils.c
AgentXLP: I agree

src/pc/djui/djui_hud_utils.c
AgentXLP: I agree

src/pc/djui/djui_hud_utils.h
AgentXLP: I agree

src/pc/djui/djui_hud_utils.c
Cooliokid956: while fixing this, i found that that entire section was offtab
image

thanks

src/pc/djui/djui_hud_utils.c
Cooliokid956: assuming that first line wasn't pointless since this happens after its removal
image

Cooliokid956:
..oh... I think I see why they were made floats

Cooliokid956:
> its interesting that peachy made `rotation` a float, but it doesn't seem to be necessary because of the format of sm64 angles... 

The reason that the angles were made `float`s was because the internal rotation function used (`guRotate`) works with degrees instead of SM64 units.. using `int`s instead led to [jitter at low angular velocities](https://github.com/coop-deluxe/sm64coopdx/pull/208/commits/8dd38e67ca44606175107e79306815e7a4c74d9a), but that's been fixed now. They are back to using `float`s as they should.

This Pull Request is ready for merging

Isaac0-dev (approved):

PR #213, opened 2024-07-15
`HOOK_ON_NAMETAGS_RENDER`

EmeraldLoc:
Adds `HOOK_ON_NAMETAGS_RENDER`. Has the playerIndex as a parameter, with a string as a return value. If the string is empty, no tag is rendered, if there's no return value, or the return value is nil, it renders the name as normal. Returning a custom string renders that custom string instead of the player's name.


src/pc/lua/smlua_hooks.c
```diff
@@ -690,6 +691,36 @@ void smlua_call_event_hooks_int_params_ret_int(enum LuaHookedEventType hookType,
     }
 }
 
+void smlua_call_event_hooks_int_params_ret_string(enum LuaHookedEventType hookType, s32 param, char** returnValue) {
```
AgentXLP, line 694: Shouldn't this be `char*` not `char**`?

docs/lua/guides/hooks.md
```diff
@@ -137,6 +137,7 @@ The lua functions sent to `hook_event()` will be automatically called by SM64 wh
 | HOOK_ON_ATTACK_OBJECT | Called when a player attacks an object. May be double-fired in some cases, you'll need to write special code for this | [MarioState](structs.md#MarioState) attacker, [Object](structs.md#Object) victim, `integer` interactionId |
 | HOOK_ON_LANGUAGE_CHANGED | Called when the language is changed | `string` language |
 | HOOK_ON_MODS_LOADED | Called directly after every mod file is loaded in by smlua | None |
+| HOOK_ON_NAMETAGS_RENDER | Called when nametags are rendered. Return a `string` to change what renders on the nametag, return an empty `string` to render nothing. | `string` playerIndex |
```
AgentXLP, line 140: Thank you for not forgetting hooks.md

src/pc/lua/smlua_hooks.c
Isaac0-dev: no its fine, because over in `nametags.c`, `hookedString` is a pointer.
thats because the hook passes on a memory buffer that is allocated by Lua when the string is converted (`smlua_to_string`).

Isaac0-dev: actually on second thoughts, you're right.
the call to `smlua_call_event_hooks_int_params_ret_string` should not reference `hookedString` as a double pointer
we don't need a double pointer we only need a reference to the Lua string buffer, not the pointer that created it
(its likely to work either way, but this isn't necessary and its confusing, except that its not dereferenced)

EmeraldLoc: Can you explain what you mean a bit more? My C understanding is not that good, and I'm not finding anything else online. If you're curious, I read to do that from [this guy on Stack Overflow](https://stackoverflow.com/a/14331455), which made sense in my head.

Isaac0-dev:
can you give an example for what you might need this hook for?

EmeraldLoc:
For me I’d like to use it for titles before the name and show the lives on it, and also for showing roles. Having a hook for this instead of checking on launch if the name tags are in or off then recreating the name tags seems a whole lot easier for new modders and experienced ones alike, so ye.

PR #216, opened 2024-07-17
a bunch of fixes

Isaac0-dev:
Here's a bunch of fixes for issues that have been found in the latest release

- some changes to frame delaying to hopefully fix some frame timing inconsistencies
- fixed a bug where the pause menu could be duplicated and spammed
- fixed a bug where the chat box was open, but was out of focus and could not be refocused (temporary solution for this bug was to open the djui pause menu)
- stopped updating cap flags for remote players, to fix an issue where all players in the level can lose their cap thanks to one person losing theirs
- palette editor cutscene improvements:
	- more reliable way of ensuring cap in hand does not happen
	- show the press Z to toggle cap text on palette editor too
	- hide the press Z to toggle cap text when in an incompatible action
- crash and bug fixes for the miniaudio implementation:
	- fixed a bug where sample copies cannot be created because loading audio directly from memory buffers was added after adding the miniaudio implementation
	- do not allow mods to fully destroy mod audio sounds to avoid memory bugs
	- made some mod audio fields invisible and immutable to prevent mods from tampering with things they shouldn't
- made the mod warning popup only show for the `DEVELOPMENT` compile flag. This is because casual players do not need to be bothered by mod warnings. The message is still displayed in the djui console however regardless of compile flags.

### DIFF
--- a/autogen/convert_structs.py
+++ b/autogen/convert_structs.py
@@ -86,6 +86,7 @@ override_field_invisible = {
     "NetworkPlayer": [ "gag", "moderator"],
     "GraphNode": [ "_guard1", "_guard2" ],
     "Object": [ "firstSurface" ],
+    "ModAudio": [ "sound", "decoder", "buffer", "bufferSize", "sampleCopiesTail" ],
 }
 
 override_field_deprecated = {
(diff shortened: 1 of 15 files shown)


AgentXLP:
Looks good

PR #221, opened 2024-07-18
Fix bettercam zoom & pan level affecting main menu

Mechstreme:
if you zoom out to the furthest distance with bettercam, the main menu's camera gets zoomed in
the pan level, if enabled, can also affect the main menu's camera
this PR aims to fix both of these issues.

https://github.com/user-attachments/assets/36c5cffa-d837-41f8-bf81-53d7ed1bafbb


AgentXLP:
thats an awesome bug

PR #240, opened 2024-07-23
Excoop Aliased Font in CoopDX

JairoThePlumber:
I thought of adding the excoop aliased font for those who want to use them, now for now I only add the font itself, not sure how it could be added to the Djui Font option, but for now it's a extra font that anyone can use for there mods (despite it used the same files from dx aliased)


JairoThePlumber:
> I thought of adding the excoop aliased font for those who want to use them, now for now I only add the font itself, not sure how it could be added to the Djui Font option, but for now it's a extra font that anyone can use for there mods (despite it used the same files from dx aliased)

PR #247, opened 2024-07-27
Mumble Positional Audio Integration

jordanvr:
Hi,

I've adapted the positional audio link plugin Mumble provide into sm64 coop dx. The mumble connection relies on sharing memory so I figured it would be more appropriate to integrate it into the project proper rather than as a mod.

the plugin details can be found here: https://www.mumble.info/documentation/developer/positional-audio/link-plugin/

image

I've tested it on Windows and OSX, and verified that the audio positioning works as expected by testing with two players connected to mumble. 

Let me know if there is anything else you need for the PR, otherwise happy to see it become part of the project!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -506,6 +506,8 @@ ifeq ($(DISCORD_SDK),1)
   SRC_DIRS += src/pc/discord
 endif
 
+SRC_DIRS += src/pc/mumble
+
 ULTRA_SRC_DIRS := lib/src lib/src/math lib/asm lib/data
 ULTRA_BIN_DIRS := lib/bin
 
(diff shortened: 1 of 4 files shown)


EmeraldLoc:
Looks cool! Wanted to lyk you should merge into the `dev` branch rather then `main`.

AgentXLP:
How does mumble hook into this? Does it read the memory of the process?

jordanvr:
> How does mumble hook into this? Does it read the memory of the process?

This technique uses [Named Shared Memory](https://learn.microsoft.com/en-us/windows/win32/memory/creating-named-shared-memory), so yes Mumble reads this defined section of memory, but it as it is shared there are not the normal issues around reading the memory of a another process.

There is another technique that Mumble plugins can use that involves identifying the offsets of position data in ram and using those, which does straight up read the memory of a game's process.

jordanvr:
I've just added an improvement around how the context is handled around rooms. I've left a comment in the code, but essentially the current room is not necessarily a room and shouldn't be used for context changes.

PR #249, opened 2024-07-28
Excoop Aliased Font

JairoThePlumber:
I accidentally close the pull request by mistake when I try to add a video of how it would look, I have to redo the pull request again, but I make sure to check is all the files are ready and making sure the texture is ready as well.

For the Language, not sure all of them need to be called "Excoop" but if you find a better name for different Languages, i'll update the name.

Anyway, here's the Video:
https://github.com/user-attachments/assets/518da019-f5bb-4c7b-b7d0-a7589160c405


AgentXLP:
The current aliased font is fine. If people don't like it then they can use a DynOS pack that has the old one.

PR #259, opened 2024-08-06
fixes, clean up and some new features

Isaac0-dev:


src/pc/configfile.c
AgentXLP, line 1: Let's Not do this

robertkirkman: what if you do it but just don't do the `SecretConfigOption` thing, looks better to me as a normal option

Isaac0-dev:
finally decided to create a list of changes, hope thats helpful.

Isaac0-dev:
I've pushed the changes I made which prevents the use-after-free bug. 
This is working as intended now. Now I will try to explain the bug a bit:
In all builds of coopdx 1.0, there is a bug that allows modders to trigger use-after-free behavior. 
Its as simple as grabbing any collision surface pointer in Lua, and holding onto that pointer even after the object itself is freed. If you wait until a level warp, the surface pool is cleared, freeing all static surfaces in the pool. This allows Lua to hold a pointer to a memory block which has been freed. Lua can then read and write from that pointer whenever it wants and can therefore cause crashes if that memory pointer is used up again later. 
This wasn't possible in previous coop versions because surfaces were never freed, just being leaked instead. Its very good that this was fixed, but it's given way to a new bug, which needs to be handled differently.
Currently, the only place I'm aware of that could have this bug is the surface pool, but I'm sure there are probably more Lua exposed objects that could be dynamically allocated, and freed in the time Lua is active. 
If any more situations are found like the surface pool, simply swapping free function calls with my custom `smlua_free` should fix the issue.

src/pc/configfile.c
Isaac0-dev: > Let's Not do this

what change do you mean, and why?

PR #280, opened 2024-08-17
F3DEX3 implementation [WIP, Help needed]

eros71-dev:
Currently trying to add F3DEX3 for the extra features and performance boosts, unfortunately the new rom extraction system doesn't allow it because some constants for the models use ROM_ASSET_LOAD_TEXTURE and the like, which the gbi.h doesn't like.


eros71-dev:
Here's an example of the errors being thrown:

include/PR/gbi.h:3743:5: error: initializer element is not constant
 3743 |     (unsigned int)(i)                       \
      |     ^

ThePerfectMario64:
Why'd you remove all the light defines?

eros71-dev:
The original repo had a script called fixlights.py that apparently is needed to be ran to remove some things that f3dex3 doesn't use/need(?), running it did that, to be fair I'm also looking into it because the script doesn't even run properly without erroring out sometimes. I didn't have time to revert the actors and levels folder yet before I look into it.

ThePerfectMario64:
re-add the light's just change the direction?(idk what the menu means by direction) from 0x28 to 0x49 to fix lighting

ThePerfectMario64:
you forgot to change 0x28 to 0x49

ThePerfectMario64:
on all lights

eros71-dev:
To be fair, I'm fairly certain f3dex3 does not even need those anyways, and the HackerSM64 repo seems to confirm that as well, I'm just gonna see if I figure out why the fixlights script failed sometimes, so I made that commit in case I need to revert to it once again because I broke something

eros71-dev:
Some updates, got it to "run"

imagen

eros71-dev:
Also, important, more or less, if we want these new features, we'll have to work on what this says (the occlusion plane may not be a must to be honest.)
imagen

Isaac0-dev:
this pr was opened for the wrong branch. Due to inactivity I'm going to close this branch.

PR #289, opened 2024-09-02
Fix Compiler Bug.

ThePlayerRolo:


src/pc/djui/djui_panel_mod_menu.c
```diff
@@ -88,6 +88,7 @@ static void djui_panel_mod_menu_mod_create_element(struct DjuiBase* parent, int
             break;
         }
         case MOD_MENU_ELEMENT_MAX:
+        break;
```
AgentXLP: Indent this `break;`

docs/lua/examples/Custom Text Example.lua
AgentXLP, line 1: This file already exists, git weirdness?

ThePlayerRolo: Yup, Idk why it wants to remerge it.

PR #294, opened 2024-09-02
Fix building on linux

ThePerfectMario64:


ThePerfectMario64:
better version of this #295

PR #295, opened 2024-09-02
Fix build on Linux

TheEssem:
The newly added Mumble support fails to build on Linux due to the lack of the `libc.h` header (my assumption is that the non-Windows includes were mainly tested on macOS, where that header does exist). This replaces it with an include for the `unistd.h` header, which works on both macOS and Linux and is where the `getuid` function is actually defined.


ThePerfectMario64:
bro i already did this

ThePerfectMario64:
but this sleeker

TheEssem:
> bro i already did this

Apologies, I didn't notice your PR.

PR #310, opened 2024-09-08
Metal as it's own recolor option

ThePerfectMario64:
fixed max lights
fixed the config setting
fixed recolor equaling to emblem


ThePerfectMario64:
I think i messed up the palettes, The color is equaling to their cap (by that i mean it just mirrors the light color) i think i may switch it to white for default palette

AgentXLP:
11 yo's first attempt at C

ThePerfectMario64:
😒

PR #314, opened 2024-09-09
Fix `sequences_offsets` file reference

### LINKED ISSUE #186: fatal error: 'sound/sequences_offsets.inc.c' file not found #include "sound/sequences_offsets.inc.c"
as in title

happens whenever I try to manualy compile it for my intel Mac on Macos 10.15.7

Expected to fully compile it with no issues.

Screen Shot 2024-07-08 at 7 36 47 PM

**Desktop (please complete the following information):**
 - OS: MacOS X 
 - Browser: Chrome
 - Version: 10.15.7
I did all the steps on this: https://github.com/coop-deluxe/sm64coopdx/wiki/Compiling-(macOS-Intel)

AloXado320:
Fixes #186 , `sequences_offsets.inc.c` was referencing the wrong file and `tools/assemble_sound.py` only generates that file while processing `sequences.bin`, since that rule wasn't reached, compilation fails

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1327,9 +1327,6 @@ $(SOUND_BIN_DIR)/tbl_header: $(SOUND_BIN_DIR)/sound_data.ctl
 $(SOUND_BIN_DIR)/samples_offsets.inc.c: $(SOUND_BIN_DIR)/sound_data.ctl
 	@true
 
-$(SOUND_BIN_DIR)/sequences_offsets.inc.c: $(SOUND_BIN_DIR)/sound_data.ctl
-	@true
-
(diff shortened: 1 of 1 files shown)


robertkirkman:
Awesome I talked about that [here](https://github.com/coop-deluxe/sm64coopdx/issues/276#issuecomment-2292190707) and I'm working on a more robust solution to the problem than just fixing the single error, since the error became too-easily possible due to a complex series of other factors, but in the mean time it's a good idea to merge this PR as a solution for now since it is taking a long time to make a more aggressive solution to prevent similar errors in the future.

PR #321, opened 2024-09-15
Implemented smlua_text_utils_dialog_get

EliteMasterEric:
- Implemented `smlua_text_utils_dialog_get` as a new Lua API function.
  - This function returns a table providing the information for a given line of dialog.
  - This allows for mods to create custom dialog interfaces, and for mods to perform inline replacements on the text of existing dialog (for example, replacing one name with another)
  - This function respects any previous replacements that have been performed using `smlua_text_utils_dialog_replace`.
- Added a hard-coded limit of 1024 characters to `smlua_text_utils_dialog_replace`.
  - This should not conflict with the vanilla game, as the longest dialog is the Lakitu roughly 780 characters.
- Added a Lua log warning to `smlua_text_utils_dialog_replace` which displays if the provided string exceeds the hard-coded length limit of `1024`.
- Added a Lua log warning to `smlua_text_utils_dialog_replace` which displays if the target dialog ID is invalid (outside the range of 1-170).

### DIFF
--- a/src/pc/lua/utils/smlua_text_utils.h
+++ b/src/pc/lua/utils/smlua_text_utils.h
@@ -4,6 +4,8 @@
 #include "types.h"
 #include "dialog_ids.h"
 
+#define DIALOG_MAXIMUM_LENGTH 1024
+
 #define MAX_ACTS 6
 #define MAX_ACTS_AND_100_COINS 7
 
@@ -25,6 +27,7 @@ extern struct CourseName *gReplacedActNameTable[];
 void smlua_text_utils_init(void);
 void smlua_text_utils_shutdown(void);
 void smlua_text_utils_reset_all(void);
+struct DialogEntry* smlua_text_utils_dialog_get(enum DialogId dialogId);
 void smlua_text_utils_dialog_replace(enum DialogId dialogId, u32 unused, s8 linesPerBox, s16 leftOffset, s16 width, const char* str);
 void smlua_text_utils_course_acts_replace(s16 courseNum, const char* courseName, const char* act1, const char* act2, const char* act3, const char* act4, const char* act5, const char* act6);
 void smlua_text_utils_secret_star_replace(s16 courseNum, const char* courseName);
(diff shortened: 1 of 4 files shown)


Squishy6094 (approved):
The madman picked up C and Lua did it in 2 hours,, I plan on implementing this in Character Select for replacing dialog from "Mario" to your current character.

src/pc/lua/utils/smlua_text_utils.h
```diff
@@ -40,4 +43,6 @@ void smlua_text_utils_castle_secret_stars_replace(const char* name);
 void smlua_text_utils_extra_text_replace(s16 index, const char* text);
 const char* smlua_text_utils_get_language(void);
 
+void convert_string_sm64_to_ascii(char *strAscii, const u8 *str64);
```
Isaac0-dev, line 46: `convert_string_sm64_to_ascii` does not belong to `smlua_text_utils.c`
it is defined in `level_info.c`

Isaac0-dev (changes requested):
Limiting the dialog length may cause mod incompatibilities if a mod has a massive string.
I feel like it'd be best to use autogen instead of manually creating `smlua_func_smlua_text_utils_dialog_get`.
This would require exposing the struct `DialogEntry`, and allowing cobjects to handle strings.

src/pc/lua/utils/smlua_text_utils.h
EliteMasterEric: I can correct this 👍

EliteMasterEric:
> Limiting the dialog length may cause mod incompatibilities if a mod has a massive string.

The reason I limited dialogue length is because I had to convert the internal dialog string into a `char*` that could be passed to Lua; this requires using `convert_string_sm64_to_ascii`, which requires a fixed length buffer.

> I feel like it'd be best to use autogen instead of manually creating `smlua_func_smlua_text_utils_dialog_get`. This would require exposing the struct `DialogEntry`, and allowing cobjects to handle strings.

I previously used autogen and then switched to manual in order to return a table containing all the values. I am new to Lua C and do not know how to expose a struct or allow cobjects to handle the strings being passed to Lua

Isaac0-dev:
Please feel free to reopen this pull request when you have implemented the requested changes.

PR #335, opened 2024-09-23
Fix Linux compilation issues with mumble.c

Multi-Volt:
Found this issue on my system when trying to compile the latest version (v 1.0.3) on my Nobara 40 system.

src/pc/mumble/mumble.c:21:18: fatal error: libc.h: No such file or directory
   21 |         #include <libc.h>
      |                  ^~~~~~~~
compilation terminated.


Isaac0-dev:
this is already fixed on the dev branch

PR #345, opened 2024-10-02
Trig functions moved to smlua_math_utils.c for autodoc and macros reused for relevant functions

Multi-Volt:
This PR is more of a suggestion than a finalized ultra-professional implementation.

As it currently stands sins(), coss(), and atan2s() are all functions exposed to the game's Lua API. Despite this, they are ignored by the autogen for documentation as the macros are declared as explicit lua functions in smlua_functions.c. Normally, for a few functions this would be okay (still not ideal) however, smlua_math_utils.c already features re-definitions (as functions) for the min, max, and sqr macros from math_util.h.

This PR seeks to unify these two clashing ideas by integrating the macros into smlua_math_utils.c by creating functional wrappers for all relevant macros which are friendly with the autodoc generation. This prevents the copying of already written code while also allowing for the documentation to be properly generated for sins(), and coss(). In addition to this, atan2s() was allowed in the autogen script to be included in the function documentation as its already exposed due to its prototype (and thus does not need an explicit declaration in smlua_functions.c).

I would love suggestions for a better implementation and am willing to contribute and fix any issues with this solution.


Isaac0-dev:
those functions can be found in `manual.lua`, which is how IDEs can see the definition of these functions

Multi-Volt:
Updated the PR to remove the redundant documentation in `manual.lua`. Doing a simple string search it doesn't look like this file is dynamically updated via autogen so this should be fine. 

Let me know if there is anything else or if I can add clarification anywhere.

src/pc/lua/utils/smlua_math_utils.c
```diff
@@ -1,28 +1,37 @@
...
 
 f32 maxf(f32 a, f32 b) {
-    return ((a) > (b) ? (a) : (b));
+    return (f32)max(a,b);
 }
 
-s32 sqr(s32 x) {
-    return x * x;
+s32 (sqr)(s32 x) {
+    return (s32)sqr(x);
```
Isaac0-dev: i don't think redefining these is a problem, but if it really is, it'd be better to just expose the functions that are in `math_util.c`.

Issue #349, opened 2024-10-05
Make dives launch players farther depending on how fast you dived at them


TheMan6900:
It just makes more sense, and would be little more fun.

Issue #343, opened 2024-09-29
Increased damage for punch and slide kick


SausRelicsSSJ2:
The damage punching deals to other players is upped to 3 and slide kick is upped to 2. I feel this is necessary because it gives people an incentive to try other moves besides the far faster and more spammable moves, normal dive and jumpkick, for the same amount of damage as these attacks currently.

TheMan6900:
Could also make these moves deal more knockback.

yoyeet961:
https://github.com/coop-deluxe/sm64coopdx/issues/344#issuecomment-2385552322

Issue #264, opened 2024-08-08
A way to access player's coopnet id's (via lua)


TheMan6900:
would be useful for mods.

TheMan6900:
a way for lua to access player ids.

frog8412:
well im sure that already exists i just dont know how to do it

PR #353, opened 2024-10-07
Add some suggestions from issues

Sunketchupm:
Each commit is a different suggestion. Specifically, issues #343, #349, and #264. One of the suggestions doesn't come from an issue but still something I heard about and wanted to make.


Sunketchupm:
Forgot to test, will retract until I fix any issues that come around

PR #354, opened 2024-10-07
Add some suggestions from issues

Sunketchupm:
Each commit is a different suggestion. Specifically, issues #343, #349, and #264. One of the suggestions doesn't come from an issue but still something I heard about and wanted to make.


Cooliokid956:
is there a reason why the coopnet id function was limited to only the local player's id? the issue didn't narrow it down to only the local id

Sunketchupm:
The reason is that this is based almost entirely on the get discord id function.
There's no real reason otherwise though, so I'll change that.

Cooliokid956:
if you didn't notice, the original discord id function was undeprecated a while ago, now discord ids are networked

src/pc/lua/utils/smlua_misc_utils.c
```diff
@@ -426,6 +426,14 @@ const char* get_local_discord_id(void) {
 #endif
 }
 
+const char* get_coopnet_id(s8 localIndex) {
```
Isaac0-dev, line 430: this function will allow mods to see ip addresses if direct connection is used, which isn't really ideal.

src/game/interaction.c
```diff
@@ -1327,10 +1328,12 @@ static u8 resolve_player_collision(struct MarioState* m, struct MarioState* m2)
...
 
 u8 determine_player_damage_value(u32 interaction) {
-    if (interaction & INT_GROUND_POUND_OR_TWIRL) { return 3; }
+    if (interaction & INT_GROUND_POUND) { return 4; }
+    if (interaction & (INT_TWIRL | INT_PUNCH | INT_TRIP)) { return 3; }
     if (interaction & INT_KICK) { return 2; }
-    if (interaction & INT_ATTACK_SLIDE) { return 1; }
-    return 2;
+    if (interaction & INT_SLIDE_KICK) { return 2; }
+    if (interaction & INT_FAST_ATTACK_OR_SHELL) { return 1; }
```
Isaac0-dev: this line does nothing

Isaac0-dev (changes requested):

src/game/interaction.c
Sunketchupm: Which line?

Isaac0-dev: `if (interaction & INT_FAST_ATTACK_OR_SHELL) { return 1; }`

Sunketchupm:
Surprised no one mentioned the pvp damage changes that were suggested. I talked with saus and theman, both of whom do a lot of pvp, and they think this is good. However I would like some other opinions on it since it will change pvp a lot.

EmeraldLoc:
PvP changes are generally good. Punches should 100% do more damage, so it’s good it does that now. The only thing that I feel like could be iffy is the slide kick, but that’ll take some tinkering. Also I don’t know how I feel about the dive thing, while it does make sense, dives having good kickback feels pretty broken, once again takes some testing.

Issue #150, opened 2024-06-04
Piranha Plant's Hitbox Is Completely Inaccurate to Every Other Version of the Game


sher-b:
To put it simply, piranha plant hitboxes are just inaccurate to what it's meant to be. In Excoop/Coopdx, the hitbox is much smaller in comparison to the sm64-ex fork, alongside these versions:
N64
Emulator
sm64-port-master
sm64-ex

**How to Replicate**
There really isn't much you need to do to reproduce this. Open up EXCoop/CoopDX and create a vanilla server. Collect enough stars to enter Whomp's Fortress and walk to the piranha plant near the grey stairs. Using the bricks in the ground, attempt to kill the piranha plant from two bricks away. It will not work. 

Then open up any other version of SM64 and follow the same steps as before. You will bounce on the piranha plant from two bricks away if you follow it correctly. 

What's meant to happen is pretty clear from the replication steps. You're meant to easily kill the piranha plant from much further than you normally can in both coop versions. There really isn't much to this.

**Visual Proof**
Here are some screenshots taken on the first frame of impact to better exemplify what I mean:

These are screenshots taken on N64, Emulator, and PC Port respectively.
![image](https://github.com/coop-deluxe/sm64coopdx/assets/171748633/33d8e1b6-e64b-46bc-9d9c-3776f0f1fa0e)
![image](https://github.com/coop-deluxe/sm64coopdx/assets/171748633/392f63f8-aa27-46fe-b485-10479fc3e2c7)
![image](https://github.com/coop-deluxe/sm64coopdx/assets/171748633/82cb929d-5b29-4e14-8d1b-7e8aaefacd3a)

And this is a screenshot from EXCoop and CoopDX respectively.
![image](https://github.com/coop-deluxe/sm64coopdx/assets/171748633/56dad75e-a10d-45b4-84af-07b165853663)
![image](https://github.com/coop-deluxe/sm64coopdx/assets/171748633/b30cfd0f-49cd-4779-8977-56cd1d62ebb0)

Videos will be attached to show proper gameplay if it wasn't already obvious. Emulator and EXCoop will be attached in the same video as this was what was provided on the discord server.

**The order of videos goes as follows: N64, PC Port, EXCoop/Emulator, CoopDX**

https://github.com/coop-deluxe/sm64coopdx/assets/171748633/edba33ab-067f-4edb-8991-70cb7e3681be
[...]

PR #355, opened 2024-10-07
Fix piranha plant hitboxes

Sunketchupm:
Bug reported in #150

### DIFF
--- a/src/game/behaviors/piranha_plant.inc.c
+++ b/src/game/behaviors/piranha_plant.inc.c
@@ -122,9 +122,6 @@ void piranha_plant_act_woken_up(void) {
      */
     o->oDamageOrCoinValue = 3;
 #endif
-    // The piranha plant still normally has the larger hitbox up to this point
-    cur_obj_set_hitbox_radius_and_height(250.0f, 200.0f);
-    cur_obj_set_hurtbox_radius_and_height(150.0f, 100.0f);
 
     if (o->oTimer == 0)
         stop_secondary_music(50);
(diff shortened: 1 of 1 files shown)


src/game/behaviors/piranha_plant.inc.c
```diff
@@ -122,6 +122,10 @@ void piranha_plant_act_woken_up(void) {
      */
     o->oDamageOrCoinValue = 3;
 #endif
+    // The piranha plant still normally has the larger hitbox up to this point
+    cur_obj_set_hitbox_radius_and_height(250.0f, 200.0f);
+    cur_obj_set_hurtbox_radius_and_height(150.0f, 100.0f);
```
Isaac0-dev: why is this necessary? the original sm64 decomp didn't have this, so how come its necessary to force the hitbox? that'll mess with mods and its not vanilla behavior.
if this is the best work around for this bug, thats fine, I just wonder why the bug even exists in the first place

Sunketchupm: In vanilla sm64, the hitbox gets set while sleeping and that doesn't change until the plant starts biting.
In coop, the biting hitbox is set every frame. I don't know why it does this, but I assume it's to force a hitbox to exist no matter what. Regardless, since the biting hitbox is set every frame, when the action becomes start waking up, it will instead use the biting hitbox rather than the sleeping hitbox.

Isaac0-dev: I'd suggest its a bug in itself. This has been a common issue with the behaviors in coop, where certain parts of their behaviors are always as though there is always a player nearby. This is just slightly broken logic that can be fixed

Sunketchupm: The hitbox being set to biting every frame isn't a bug. It's set directly in the loop code.

Isaac0-dev: yeah i know what you mean i'm saying there's probably a bug in the code that was changed for syncing the piranha plants

Isaac0-dev (approved):

PR #356, opened 2024-10-08
Add ban-kick buttons to playerlist

Cooliokid956:
Adds two buttons for banning/kicking when hovering over a player list entry.

# Kick
Click once to bring up a confirmation button, click again to kick.
image
image
image

# Ban
Click once to bring up a confirmation button, click again to ban.
image

# Perm Ban
Click the ban button, then the Kick button will transform into the Perm Ban button. Once clicked, click again to confirm. You may also click Go Back (or stop hovering over the player) to reset the buttons.
image
image

### DIFF
--- a/src/pc/djui/djui_cursor.c
+++ b/src/pc/djui/djui_cursor.c
@@ -1,5 +1,6 @@
 #include "djui.h"
 #include "djui_panel.h"
+#include "djui_panel_playerlist.h"
 #include "pc/controller/controller_mouse.h"
 #include "pc/gfx/gfx_window_manager_api.h"
 #include "pc/pc_main.h"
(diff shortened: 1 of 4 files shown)


DayMario64:
Super peak!

DayMario64:
Is the blood included (i wish)

DayMario64:
I suggest making the player list toggleable so you don't have to hold tab down and add roles. Aka owner, admin, player etc.

EmeraldLoc:
I'd be curious on what other players think, however for me, I don't like having the mouse be unusable for the camera when the playerlist is open. Perhaps if mouse look is on, you could have the user press a button to activate the mouse, but that might be a bit too complicated for new players.

Isaac0-dev:
While I think its a great idea to have such an easy way to kick and ban, I also feel like it'd be pretty weird having the cursor active while the player list is up.

djoslin0:
oh cool, this was something I never got around to

iZePlayzYT:
i would put it in here ("Quick Access Moderation Menu" is just a placeholder name), so you can ban and kick, etc. players (maybe also display their ID's). so you can navigate to there with just your controller without needing to use the keyboard or mouse:
image

Pixelsz64:
> i would put it in here ("Quick Access Moderation Menu" is just a placeholder name), so you can ban and kick, etc. players (maybe also display their ID's). so you can navigate to there with just your controller without needing to use the keyboard or mouse: image

Agreed this would be very helpful and cool

PR #357, opened 2024-10-08
[WIP] Rework PvP Balance

Sunketchupm:
Do not merge until fully complete.
After the last PR I made that increased the damage of many attacks, there were some people that thought that change didn't really make sense. After talking with some people, namely saus, phenominal sire, and theman, we've decided to change around some attacks, in a different way. I would like to get some more input for this, which is why I'm making a PR despite this being unfinished. Mod compatibility and maybe even some obscure vanilla behavior may be affected.

As of the creation of this PR (October 7th), this is the state of all attacks:
Ground pounds and twirls deal 3 damage.
Jump kicks, slidekicks, and breakdances deal 2 damage.
Punches (and all other attacks) deal 1 damage.
Punches are made unique in that instead of being given the usual 30 frames of invincTimer upon being hit, it instead only gives 12.


EmeraldLoc:
I dont agree with the fact that breakdances only do 2 damage, they should def do 3. Hitting a breakdance has got to be one of the hardest moves, cuz you have to get crouched, be slow enough, and hope the other player isn't diving. Twirls being 3 feels pretty broken, since once you're in the action it's, compared to a GP and a breakdance, easy to land a hit, 2 would be a good value imo. Punches depends, if the I frames allows you to hit all 3 punches, then yea sure, otherwise though that's back to square one. If Mario could hit all 3 with a proper timing that'd be dope.

Sunketchupm:
Twirls deal 3 because twirls and ground pounds share the same interaction type. I should be able to separate them though.
As much as I'd love to make punches be able to hit their entire combo, the entirety of the knockback action makes it impossible even with 0 knockback and 0 invincibility. If you can figure out a way to ignore a knockback's invincibility for punches, I'll add that and rework it.

EmeraldLoc:
> As much as I'd love to make punches be able to hit their entire combo, the entirety of the knockback action makes it impossible even with 0 knockback and 0 invincibility.

This is what I thought, was hoping the timing could work out. It would most likely take some ugly code to get done. The most streamlined approach I can think of is creating a whole new action dedicated to this, otherwise though the kb action has a specific action flag for invulnerability that'd need to be overwritten, but it can't be overwritten when hit by a bully or hit with a dive. As for twirls, yea that should be easy, simply one line above the action flag check check if Mario's in the twirl action, since there's only one (if you dont include the landing action).

Cooliokid956:
if you make a pr that isn't ready to pull yet, you can mark it as a draft

Sunketchupm:
> This is what I thought, was hoping the timing could work out. It would most likely take some ugly code to get done. 

I managed to not only make this work, but with pretty clean code as well. Here is a video demonstrating.
https://github.com/user-attachments/assets/32be81a3-a3ee-43b6-af18-6ee3165b924d

EmeraldLoc:
Oh dang, nice! Yea that's not bad at all! Thats def the way to go, that looks dope :D

Sunketchupm:
Lots of changes since yesterday if anyone is interested in reading them

EmeraldLoc:
Why are we making slide kicks not use speed to check if they beat a dive? Maybe you could buff the slide kick by a tad my checking if the speed difference is within a range (perhaps 5 or so). You could also check the vertical velocity to see if the slide kick is heading up to kick Mario directly in the face, or if Mario is diving downwards into Mario’s body. Just my opinion ofc

Sunketchupm:
The new purpose of the slidekick is to be able to attack a dive from standstill.
The speed a slidekick reaches from standstill and on flat ground before sliding is around 35 (as starting up a slidekick immediately bumps Mario to 32 speed). The speed a ground dive reaches on flat ground is around 49. The speed a single jump dive reaches on flat ground is 52.
Finding a range that can reasonably deal with dives and jump dives is very difficult and this is not even accounting for the vertical speed being a factor the speed check or uneven terrain (which affects slidekicks more than all other movement options). Since it is rather hard to hit a dive in the first place, and whatever balance that could be found for slidekicks would likely be favoring them, I don't think it is worthwhile to make a check at all.
If there is enough testing done that shows slidekicks to be too strong due to no speed check, then I'll look into it more strongly.

Sunketchupm:
This PR is just about ready now except for different languages. I added the option to switch between what the current coop pvp is and this version of the coop pvp.

src/game/interaction.c
```diff
@@ -1450,10 +1472,14 @@ u32 interact_player_pvp(struct MarioState* attacker, struct MarioState* victim)
...
     // determine if slide attack should be ignored
+    // Ground pounds will always be able to hit
     if ((interaction & INT_ATTACK_SLIDE) || player_is_sliding(cVictim)) {
         // determine the difference in velocities
-        Vec3f velDiff;
-        vec3f_dif(velDiff, attacker->vel, cVictim->vel);
+        //Vec3f velDiff;
+        //vec3f_dif(velDiff, attacker->vel, cVictim->vel);
+        if (gServerSettings.pvpType == PLAYER_PVP_REVAMPED && attacker->action == ACT_GROUND_POUND) {
+            goto ignoreSlideCheck;
```
Isaac0-dev: preferably don't use gotos

PR #363, opened 2024-10-11
Fix early hits

Sunketchupm:
Sometimes, the player can ground pound a pole before even landing or punch/kick an enemy at high speeds which will kill the enemy but not give bounce back to the player.
Mod incompatibilities are guaranteed, so it's best to look at how to make this incompatible with as little mods as possible.
In case it's hard to understand what exactly this is fixing, see the videos.

Punching through a goomba (in vanilla sm64 you would get hit by the goomba)
https://github.com/user-attachments/assets/4d215de9-091e-4e93-969d-99386ec0a2e3
Double pounding the pole (this seems to only happen in multiplayer)
https://github.com/user-attachments/assets/7796d91b-21cb-4b1e-a869-9f358522a02b


src/game/interaction.c
```diff
@@ -158,7 +158,8 @@ u32 determine_interaction(struct MarioState *m, struct Object *o) {
 
     if (interaction == 0 && action & ACT_FLAG_ATTACKING) {
         u32 flags = (MARIO_PUNCHING | MARIO_KICKING | MARIO_TRIPPING);
-        if (m->flags & flags) {
+        if ((action == ACT_PUNCHING || action == ACT_MOVE_PUNCHING || action == ACT_JUMP_KICK) ||
```
Isaac0-dev, line 161: I'm not sure about this change because actions are supposed to asign the flags, rather than a punch always creating a hit for example.

Sunketchupm: I'd agree but vanilla sm64 does this

Isaac0-dev: oh i see you're right, sorry about that https://github.com/n64decomp/sm64/blob/9921382a68bb0c865e5e45eb594d9c64db59b1af/src/game/interaction.c#L187

PR #375, opened 2024-10-16
Fix the branch bloat found in the atan2s and atan2_lookup functions

Multi-Volt:
Hello,

Currently, both the `atan2s` and `atan2_lookup` functions contain unnecessary branching that can be significantly optimized. This proposal implements changes aimed at reducing branch complexity through the use of bit manipulation, ternary operators where appropriate, and static constant tables. These optimizations not only simplify the code but also result in a minor performance improvement, likely attributed to enhanced CPU caching behavior due to the static tables as well as move towards inlining.

Before PR Changes:
image

After PR Changes:
image

### DIFF
--- a/src/engine/math_util.c
+++ b/src/engine/math_util.c
@@ -822,7 +822,7 @@ f32 approach_f32(f32 current, f32 target, f32 inc, f32 dec) {
  * Helper function for atan2s. Does a look up of the arctangent of y/x assuming
  * the resulting angle is in range [0, 0x2000] (1/8 of a circle).
  */
-static inline u16 atan2_lookup(f32 y, f32 x) {
+static u16 atan2_lookup(f32 y, f32 x) {
     s16 idx = (s16)(y / x * 1024.0f + 0.5f);
     idx = (idx >= 0 && idx < 0x401) ? idx : 0;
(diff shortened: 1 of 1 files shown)


robertkirkman:
It works on ARM. Good job! I tested using Arch Linux ARM aarch64 and [the instance of call to `atan2s()` used to calculate the knockback direction of players hit by bob-ombs in Arena](https://github.com/coop-deluxe/sm64coopdx/blob/main/mods/arena/arena-proj-bobomb.lua#L83)

Multi-Volt:
> It works on ARM. Good job! I tested using Arch Linux ARM aarch64 and [the instance of call to `atan2s()` used to calculate the knockback direction of players hit by bob-ombs in Arena](https://github.com/coop-deluxe/sm64coopdx/blob/main/mods/arena/arena-proj-bobomb.lua#L83)

Awesome! Thank you for testing my changes, I am glad it works fine on ARM! That means it should work fine on the RPI which is one of our targets.

theclashingfritz:
This is really cool to see, Optimizations of the games code is always welcome, Any and all performance gained is a boon and gives more freedom for things such as mods. I haven't tested this yet but I look forward to it.

PR #377, opened 2024-10-18
Optimization of mtxf_lookat()

Multi-Volt:
Hello,

While reviewing the current implementation of `mtxf_lookat()` in `math_util.c`, I identified some areas for optimization. Specifically, I noticed redundant calculations in the cross product operations, as well as the absence of checks for very small or zero xz distances. Any testing for these changes would be appreciated, below are my results.

Before Changes:
Screenshot from 2024-10-17 23-05-42

After Changes:
Screenshot from 2024-10-17 21-22-44

### DIFF
--- a/src/engine/math_util.c
+++ b/src/engine/math_util.c
@@ -303,7 +303,8 @@ void mtxf_lookat(Mat4 mtx, Vec3f from, Vec3f to, s16 roll) {
     mtx[2][2] = forward[2];
     mtx[3][2] = -(from[0] * forward[0] + from[1] * forward[1] + from[2] * forward[2]);
 
-    mtx[0][3] = mtx[1][3] = mtx[2][3] = mtx[3][3] = 0.0f;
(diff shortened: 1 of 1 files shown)


Cooliokid956:
isn't that [3][3] supposed to be 1? or does this cause no errors

Multi-Volt:
> isn't that [3][3] supposed to be 1? or does this cause no errors 

Good find, I didn't notice that I lumped that together. In the build I never saw any differences but I went ahead and fixed it just in case.

Multi-Volt:
> the result seems to be different from the original `mtxf_lookat` if `roll` is not 0. from what i remember, vanilla sm64 doesn't use roll much but accuracy is still important

I fixed my code, I ended up forgetting a single negative sign on the normalization for the xzdist! Now the calculations appear to match for any roll.

Isaac0-dev (approved):
well done this function indeed accurate and faster in my testing.

PR #379, opened 2024-10-18
add controller.buttonReleased

Drahnokks:
- Add the `controller.buttonReleased` flag set (was tested)

- Optimized the operation to get `controller->buttonPressed` (tested too)
from : `controller->controllerData->button & (controller->controllerData->button ^ controller->buttonDown)`
to : `(~controller->buttonDown & controller->controllerData->button)`
since doing `X AND (X XOR Y`) it's the same thing as `NOT Y AND X` but simplified

### DIFF
--- a/autogen/lua_definitions/structs.lua
+++ b/autogen/lua_definitions/structs.lua
@@ -537,6 +537,7 @@
 --- @class Controller
 --- @field public buttonDown integer
 --- @field public buttonPressed integer
+--- @field public buttonReleased integer
 --- @field public extStickX integer
 --- @field public extStickY integer
 --- @field public port integer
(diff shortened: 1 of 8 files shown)


Isaac0-dev:
what is `buttonReleased` for? I'd assume for mods? If so, mods can already figure that out

Isaac0-dev:
sorry, accidentally pressed close

Isaac0-dev (approved):
I assume this is intended for mods, while its easy for mods to do this themselves, I'm sure it'll save many people some time.

Drahnokks:
> I assume this is intended for mods, while its easy for mods to do this themselves, I'm sure it'll save many people some time.

Yep it was exactly the purpose!

PR #381, opened 2024-10-19
Revamp Djui Profilers 

theclashingfritz:
This PR ports both profilers to DJUI, Does some revamps, And fixes some bugs relating to them. 

image


Bloxxel64:
Oh wow, this looks super well done! And all the extra information will probably help a lot with optimization.

PR #391, opened 2024-10-28
Make Lua audio samples silent for specific sound configuration

Flower35:
Lua audio samples (music streams and sound effects) are no longed playing, while the game window has no focus and the config option "Mute When Window Unfocused" is set to True.

This behaves similar to the check in `buffer_audio()` from `src/pc/pc_main.c`.

You might still hear a sample that is already playing while the player clicks out of the window, but no more audio samples will be audible as long as the focus is lost.

### DIFF
--- a/src/pc/lua/utils/smlua_audio_utils.c
+++ b/src/pc/lua/utils/smlua_audio_utils.c
@@ -336,9 +336,13 @@ void audio_stream_play(struct ModAudio* audio, bool restart, f32 volume) {
     if (!audio_sanity_check(audio, true, "play stream")) {
         return;
     }
-    f32 masterVolume = (f32)configMasterVolume / 127.0f * (f32)gLuaVolumeMaster / 127.0f;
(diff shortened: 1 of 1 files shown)


AgentXLP (approved):

Isaac0-dev:
this looks like it will stop all audio streams and samples without starting them again when you tab back
that would cause some mods to have their music completely stopped, which will sound like a bug.

AgentXLP:
This just makes it so when audio sample/stream play is called while the window is unfocused, it sets the volume to 0. More efficiently, it could just not be played at all, but this solution still works.
Ideally, the audio would be able to change back and forth in realtime, but that would require either updating the volume every frame, or having it update the audio when the window focus status changes and setup an event for that in the window managers. Maybe that would be something worth looking into.

Isaac0-dev:
having it update when the tabbing in would be best because the audio could be restored to its original volume

AgentXLP:
This will work for now. We can make it better in the future.

PR #490, opened 2024-11-12
Minor vertex processing optimizations

khang06:
~8% FPS increase in Star Revenge 1.5's lobby

Before:
sm64coopdx_Nl9WZgJ8of

After:
sm64coopdx_1pRzz83xWC

### DIFF
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -6,6 +6,10 @@
 #include <stdbool.h>
 #include <assert.h>
 
+#ifdef __SSE__
+#include <xmmintrin.h>
+#endif
+
 #define STB_IMAGE_IMPLEMENTATION
 #include <stb/stb_image.h>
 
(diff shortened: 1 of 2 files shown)


AgentXLP:
Very nice

src/pc/gfx/gfx_pc.c
```diff
@@ -760,19 +764,43 @@ static void gfx_sp_pop_matrix(uint32_t count) {
 }
 
 static float gfx_adjust_x_for_aspect_ratio(float x) {
-    return x * (4.0f / 3.0f) / ((float)gfx_current_dimensions.width / (float)gfx_current_dimensions.height);
+    return x * gfx_current_dimensions.x_adjust_ratio;
 }
 
 static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, const Vtx *vertices, bool luaVertexColor) {
+    float globalLightCached[2][3] = {
```
Isaac0-dev: `globalLightCached` is only used within a condition, may as well only do this math inside that condition.

Isaac0-dev (commented):
I can confirm that I get higher framerates on my computer.
Thanks for the pr, this looks good so far!

PR #516, opened 2024-11-24
Improve romhack camera Y axis smoothness

aglab2:
This patch improves "default" romhack camera Y axis smoothness. Camera behaves similarly to HackerSM64 parallel lakitu camera.


AgentXLP (approved):
Looks good

PR #518, opened 2024-11-24
Fix HOOK_ON_SEQ_LOAD

Blockyyy:
accidentally made it so you couldn't return 0, making it so you were unable to mute a sequence,
so i changed the return value to be -1 by default instead of 0, allowing you to mute sequences properly


AgentXLP (approved):
Looks good

PR #522, opened 2024-11-25
add set_room_override

Blockyyy:
lets you override the displayed room, do set_room_override(-1) in order to get back to normal

https://github.com/user-attachments/assets/18430915-a0e5-446d-a966-b6784378a439

### DIFF
--- a/src/game/object_helpers.c
+++ b/src/game/object_helpers.c
@@ -185,6 +185,26 @@ Gfx *geo_switch_anim_state(s32 callContext, struct GraphNode *node) {
     return NULL;
 }
 
+s16 gRoomOverride = -1;
+
+void set_room_override(s16 room) {
+    gRoomOverride = room;
+}
+
+s16 get_room_at_pos(f32 x, f32 y, f32 z) {
+    struct Surface *surface = NULL;
+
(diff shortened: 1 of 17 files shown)


src/game/object_helpers.c
```diff
@@ -185,6 +185,26 @@ Gfx *geo_switch_anim_state(s32 callContext, struct GraphNode *node) {
     return NULL;
 }
 
+s16 gRoomOverride = -1;
+
+void set_room_override(s16 room) {
+    gRoomOverride = room;
+}
+
+s16 get_room_at_pos(f32 x, f32 y, f32 z) {
```
Isaac0-dev: this function seems to only be useful in very specific situations. it would probably be more useful if you could enable `gFindFloorIncludeSurfaceIntangible` for your own ray casts, since you can already just get the room parameter from the surface you find.

Blockyyy: can't raycasts already get intangible surface types?, i'll probably just remove this function

Isaac0-dev: having `gFindFloorIncludeSurfaceIntangible` true does change the behavior of `find_floor` but its fine if you just remove the function too.

PR #528, opened 2024-11-27
Added djui_menu_get_excoop_theme()

xLuigiGamerx:
# `djui_menu_get_excoop_theme()`
Returns `configExCoopTheme` to check if someone is using it.


Isaac0-dev:
the ex-coop theme is a hidden easter egg and isn't meant to be exposed to Lua

PR #530, opened 2024-11-27
Fix frame rate

PeachyPeachSM64:
Fix inconsistent and inaccurate frame rate.
Still loses a bit of accuracy at high frame rates and breaks above 600 FPS due to `WAPI.delay` using a delta time in milliseconds.

Comparison:
fps

### DIFF
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -90,7 +90,8 @@ f32 gRenderingDelta = 0;
 #define FRAMERATE 30
 static const f64 sFrameTime = (1.0 / ((double)FRAMERATE));
 static f64 sFrameTargetTime = 0;
-static f64 sFrameTimeStart;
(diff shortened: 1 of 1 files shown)


AgentXLP:
Wdym breaks above 600 fps?

PeachyPeachSM64:
delay becomes 0 and the frame rate is no longer limited

djoslin0:
this one was super tricky, good work for figuring it out

Isaac0-dev:
it looks like sdl doesn't provide any ways to delay by microseconds, but would it be worth using other built in functions to delay by microseconds for more accuracy?

djoslin0:
making a frame limiter work past ~244 fps seems excessive to me, at that point just run uncapped

AgentXLP:
True

PR #508, opened 2024-11-22
more detailed mod description & added chat colors codes

iZePlayzYT:

PR #531, opened 2024-11-27
added chat color codes like in minecraft (&1, &e, &r, etc.)

iZePlayzYT:


kermeow:
This maybe is over-complicating things, especially when it's replacing all matches for those strings with the already existing colour codes.

Also correct me if I'm wrong but I don't think `#fff` is the right colour for reset.

iZePlayzYT:
1) there has been previous discussion, the new PR is because i was asked to split both changes in seperate PRs: https://github.com/coop-deluxe/sm64coopdx/pull/508

2) its more simplifying it, because writing just &1 is way faster than \#xxx\ if you want to quickly highlight something in chat. regarding the &r in minecraft it would reset the color to fff and also resets any style formatting codes like &k &m &o &l which are not existing in coopdx yet so its just resetting the color to fff currently.
 
3) it looks like this image

iZePlayzYT:
> This maybe is over-complicating things, especially when it's replacing all matches for those strings with the already existing colour codes.
> 
> Also correct me if I'm wrong but I don't think `#fff` is the right colour for reset.

wait did you mean that the text in chat without formatting is not exactly fff?

AgentXLP:
Sorry but the color codes were already simplified and I don't think we need another system for them.

iZePlayzYT:
-.-

PR #532, opened 2024-11-27
more detailed mod description

iZePlayzYT:


iZePlayzYT:
The reason for this PR is, that you can differentiate the mods if you have multiple versions of a mod with the same names. Also it shows what mods are incompatible. for example: 
image

iZePlayzYT:
Also why @AgentXLP

AgentXLP:
I think there is a better way to do this that is also localized.

PR #533, opened 2024-11-27
added new start parameter --coopnet

iZePlayzYT:


AgentXLP:
I don't agree with this implementation, there should only be server and client network types, adding a CoopNet type generates confusion and just isn't right

AgentXLP:
Then find a better way to implement it

PR #534, opened 2024-11-27
added norender mode

iZePlayzYT:


iZePlayzYT:
Exactly. This helps a lot if you want to just host without wasting to much resources

AgentXLP:
`make HEADLESS=1`?

iZePlayzYT:
headless does not work with coopnet, also the performance mode does allow you to access the chat to enter commands and you can change stuff without restarting since you have full access to the Menu/UI

AgentXLP:
I see

AgentXLP:
There is probably a better name you can pick for this I imagine

iZePlayzYT:
do you have any idea?

Isaac0-dev:
better to improve the headless server than introduce this dodgy method of running the game

iZePlayzYT:
@Isaac0-dev improving headless does not allow you to access the chat and the menus. this is way better than headless in many use-cases. also @AgentXLP approved it before.

AgentXLP:
I approved it because the code looked fine, Isaac merges PRs as of now

AgentXLP:
I do agree though that the method is a bit dodgy

iZePlayzYT:
i thought about it once again and have to agree that the parameter is a bad solution. but the ingame setting would be perfect for many reasons and cant be replaced with headless in some use cases. for example: if somebody who is currently hosting a lobby decides to play something else they would are forced to shut down the lobby because his computer is low-end and keeping coopdx running would waste to much resources. instead of shutting the game down they could disable rendering while not ingame without closing the lobby

PR #535, opened 2024-11-28
added config setting in "DISPLAY" to only render the game when focused

iZePlayzYT:


src/game/area.c
```diff
@@ -437,8 +438,10 @@ void play_transition_after_delay(s16 transType, s16 time, u8 red, u8 green, u8 b
 void render_game(void) {
     dynos_update_gfx();
     if (gCurrentArea != NULL && !gWarpTransition.pauseRendering) {
-        geo_process_root(gCurrentArea->unk04, D_8032CE74, D_8032CE78, gFBSetColor);
-
+        if (WAPI.has_focus() || !configNoRenderFocusLoss) {
+            geo_process_root(gCurrentArea->unk04, D_8032CE74, D_8032CE78, gFBSetColor);
```
PeachyPeachSM64, line 442: This function is not responsible for rendering the game, this is `gfx_run`.
A part of the game engine is updated during `geo_process_root` (like the famous HOLP and some Mario logic too), as well as some hooks like `HOOK_ON_OBJECT_RENDER`.

iZePlayzYT: But the HUD and UI should be still visible

Isaac0-dev: could disable interpolation and rendering instead, but i don't know why the hud needs to be visible out of focus

iZePlayzYT: If you for example want to read the chat or Star count in the meantime on a second screen without to much resources getting wasted or getting distracted by the visuals

iZePlayzYT: But maybe instead of a checkbox we could use one of those djui elements with multiple options and add a "Hud/UI Only" a "Disable Everything" and "Off" option

iZePlayzYT:
i removed the checkbox and replaced it with a selection box "Render When Unfocused" with the options "Everything", "HUD and UI" and "Nothing". When "Nothing" is selected it will prevent the execution of the "gfx_run" method. Which improves the performance even more when unfocused @PeachyPeachSM64 @Isaac0-dev

iZePlayzYT:
Why?

AgentXLP:
Overcomplicated option that should probably only be available in the dev build anyway

iZePlayzYT:
> Overcomplicated option that should probably only be available in the dev build anyway

ok i will change it

AgentXLP:
Don't worry about it

iZePlayzYT:
> Don't worry about it

But reduces ram usage a lot when the game is minimized. This is very useful

iZePlayzYT:
Currently its fully rendering the game even if it's not even visible

PR #536, opened 2024-11-28
Added djui_menu_get_rainbow_string_color()

xLuigiGamerx:
```lua
--- @param color integer
--- @return string
function djui_menu_get_rainbow_string_color(color)
    -- ...
end
```
Takes an integer 0-3:
Red = 0
Green = 1
Blue = 2
Yellow = 3
Returns a string with a hexcolor.


AgentXLP:
should you make an enum for the color value so you dont have to just put an integer

AgentXLP:
???

xLuigiGamerx:
> you should make an enum for the color value so you dont have to just put an integer
That's a good idea, I tried to do it but couldn't figure it out.

PR #537, opened 2024-11-28
Fix frame rate (again)

PeachyPeachSM64:
Take 2

Fix the issues introduced by the previous commit.

Computes delta time based on the remaining number of frames to draw during the current second.
More stable and doesn't break anymore at very high frame rates.
Downside is that 30 FPS had to be handled separately due to race condition.

image


Isaac0-dev:
buttery smooth

PR #544, opened 2024-12-06
Use proper controller names in settings

eros71-dev:
Make the controller names show up in the selection box instead of things like 0, 1, etc... Now it will show names like "PS4 Controller", if provided by SDL2.

### DIFF
--- a/src/pc/djui/djui_panel_controls.c
+++ b/src/pc/djui/djui_panel_controls.c
@@ -52,14 +52,7 @@ void djui_panel_controls_create(struct DjuiBase* caller) {
                     // Then we can append the number so it fits in the slot
 
                     // Should we look into making scrolling text for this?
-                    if (strlen(gamepadChoices[i]) > 9) {
-                        strncpy(newName, gamepadChoices[i], 9);
-                        newName[9] = '\0';
-                        strcat(newName, "...");
-                        snprintf(newName, sizeof(newName), "%s (%d)", newName, count);
-                    } else {
-                        snprintf(newName, sizeof(newName), "%s (%d)", gamepadChoices[i], count);
-                    }
+                    snprintf(newName, sizeof(newName), "%.9s... (%d)", gamepadChoices[i], count);
                     
                     // Remove the old string and replace it with the new one
                     free(gamepadChoices[i]);
--- a/src/pc/djui/djui_panel_controls.c
+++ b/src/pc/djui/djui_panel_controls.c
@@ -25,34 +25,55 @@ void djui_panel_controls_create(struct DjuiBase* caller) {
         djui_checkbox_create(body, DLANG(MISC, USE_STANDARD_KEY_BINDINGS_CHAT), &configUseStandardKeyBindingsChat, NULL);
 
 #ifdef HAVE_SDL2
-    int numJoys = SDL_NumJoysticks();
-    if (numJoys == 0) { numJoys = 1; }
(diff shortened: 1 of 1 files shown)


PeachyPeachSM64:
not sure if it's a github thing but the indentation looks off

Flower35:
> not sure if it's a github thing but the indentation looks off

It is off by one level (4 spaces) to the left, compared to the previous code block.

eros71-dev:
Since git was messing me up a little with my stashed f3dex3 changes I ended up copying that block somewhere else, resetting the repo locally, and then pasted it back so there's a chance I forgot to clean up the indentation afterwards.

I'll change it in a bit.

src/pc/djui/djui_panel_controls.c
```diff
@@ -25,21 +25,34 @@ void djui_panel_controls_create(struct DjuiBase* caller) {
...
+                char newName[256];
+                snprintf(newName, sizeof(newName), "%s (%d)", gamepadChoices[i], count);
+                free(gamepadChoices[i]);
+                gamepadChoices[i] = strdup(newName);
             }
-            sprintf(&gamepadChoicesLong[index], "%d", i);
-            gamepadChoices[i] = &gamepadChoicesLong[index];
         }
-        djui_selectionbox_create(body, DLANG(CONTROLS, GAMEPAD), gamepadChoices, numJoys, &configGamepadNumber, NULL);
+    }
```
PeachyPeachSM64: Can crash due to a use-after-free bug + djui choices are never updated
This block should be moved before `djui_selectionbox_create`

PeachyPeachSM64 (changes requested):

eros71-dev:
Alright, everything should be okay now, from what I see, although I still need someone to make sure the name trimming for repeated controllers works properly.

src/pc/djui/djui_panel_controls.c
```diff
@@ -28,18 +28,53 @@ void djui_panel_controls_create(struct DjuiBase* caller) {
...
+                    
+                    // If the name is bigger than 9 characters, we need to truncate it first
+                    // Then we can append the number so it fits in the slot
+
+                    // Should we look into making scrolling text for this?
+                    if (strlen(gamepadChoices[i]) > 9) {
+                        strncpy(newName, gamepadChoices[i], 9);
+                        newName[9] = '\0';
+                        strcat(newName, "...");
+                        snprintf(newName, sizeof(newName), "%s (%d)", newName, count);
```
PeachyPeachSM64: using `newName` in `snprintf` as both the destination and an argument is undefined behavior in C
but you don't need to do that, printf functions have a specific format for trimming:
```
snprintf(newName, sizeof(newName), "%.9s... (%d)", gamepadChoices[i], count);
```

PeachyPeachSM64 (changes requested):

PeachyPeachSM64:
also I like the idea of scrolling text

src/pc/djui/djui_panel_controls.c
eros71-dev: Alright, changed it, hopefully I did it right heh

PeachyPeachSM64: keep the if statement, otherwise you'll have the ellipsis even for names that have less than 9 characters

eros71-dev: Ah, right my bad

PeachyPeachSM64:
looks good to me

PR #545, opened 2024-12-07
Add Function Descriptions to The Auto-Documentation System

Multi-Volt:
Hello,

This PR introduces a description system for autodoc. The idea is straightforward: whenever the autodoc system detects a suitable function header or declaration, it will look for a comment above it that includes a description. Descriptions are indicated by two markers, `|description|` and `|descriptionEnd|`, and any text between these markers will have leading and trailing whitespace removed.

A description can appear in multiple formats, such as:

```
/*|description| ... |descriptionEnd|*/
```

```
/*|description
...
|descriptionEnd|*/
```

```
/*
|description|
...
|descriptionEnd|
*/
```

...and many more variations.

To demonstrate this new system, I have described all of the functions in `math_util.h` and `smlua_math_utils.h` to the best of my ability. Both of these are included in the `extract_functions.py` file. Additionally, this PR removes some remnants from an earlier PR in `math_util.h` that were no longer needed.

To better handle more complex inputs, I’ve also modified the function detection in `extract_functions.py` so that it now keeps track of both formatted and raw text inputs. This change ensures that the proper function descriptions are found and associated correctly. I am still relatively new to Python so any review and testing is greatly appreciated.


PeachyPeachSM64:
Shouldn't it be better to omit the `### Description` block completely instead of writing "No description available" for the undocumented functions?

AgentXLP:
I will work on this

PR #551, opened 2024-12-09
Revert/fix recent romhack camera changes

Sunketchupm:
Fixed the issue with the romhack camera being abnormally high.

The direction the camera moves with romhack camera before this change was actually the correct direction.


kingthememer:
the romhack cam direction is actually effected by the freecam invert settings
it was always like this but the direction was accidentally flipped within the romhack cam code so it was fixed
if you want the romhack cam how it was before just invert/uninvert the x axis in the freecam setting

Sunketchupm:
I would argue about this affecting vanilla camera but I just learned that invert X does not affect vanilla camera. I'll just revert my own direction change and keep just the height fix.

Issue #538, opened 2024-11-29
[Suggestion] Add Camera Collision Toggle for Free/Mouse Cam


sher-b:
Another suggestion to bridge the gap between using classic cam and free cam. It can stay off by default and be toggled either through the camera menu or by enabling some romhack camera code or smth. 

This suggestion is purely here because of kaizos. I've given suggestions regarding free cam and kaizo before, and having the option to remove camera collision when needed similar to using romhack cam would make technical play a lot easier for people who would rather learn to use free/mouse cam at a higher skill level.

sher-b:
> This would be lovely for romhacks when using freecam, however it would require mods to disable it to prevent unfair advantages in gamemodes or others

that should honestly be a given, it's already a thing with the standard romhack cam anyway. just have it work similarly to how you'd toggle romhack cam and it should work fine.

Issue #497, opened 2024-11-14
[Suggestion] An Aglab Camera For Free/Mouse Cam Players


sher-b:
Aglab cam is a feature found in romhacks and has been transferred to CoopDX. It provides more freedom in choosing precise angles and manipulating the camera to better suit technical play. As one of the only mouse cam users, I want to suggest a "reverse aglab cam" for free/mouse camera, accessible using the dpad. 

Just as aglab allows classic camera players to access precise angles on a whim, this "reverse aglab" would allow free/mouse cam players to access the opposite of that: cardinal and ordinal directions. Here is a general idea of how the binds would work:
- dpad-left: shift the camera to a cardinal direction going counter-clockwise
- dpad-right: shifts the camera to a cardinal direction going clockwise
- dpad-up: locks the camera at its current angle (toggleable)
- dpad-down: unlocks the ordinal directions to be accessed via dpad-left/right (toggleable)

This option would be accessible either through the camera options in the main menu or as a function within CoopDX's library of functions (or some third option that would work better).

The point of this addition is specifically for sm64 tech that is either impractical or outright impossible with free cam/mouse cam to be more convenient and actually possible without tearing out your hair. Although you could say that one can switch to classic cam to achieve these feats, you could say the same for aglab cam as well as this addition is meant to even the playing field between classic cam and free/mouse cam at a more experienced level. 

I've personally played through a lot of kaizos within ex-coop and CoopDX for a couple years now, and I can say for certain that there are many things that cannot be done solely because of mouse cam's limitations. I remember a C-up cam feature in the game before it was removed due to crashes, so clearly there's some effort being put into making free cam feel like a more worthy alternative to classic cam. I just want to give more perspective on how free cam as a whole affects kaizo play and SM64 tech as a whole, and having this added in will appeal to the small set of free cam users that love kaizos.

PR #555, opened 2024-12-11
More settings to the Camera settings menu

Sunketchupm:
Sherbie's suggestions from #497 and #538, as well as suggestions from others on discord.
Add 2 new camera settings to free camera, being camera collision and dpad usage.
Add an entire new camera submenu for romhack camera, which are now no longer exclusive to romhacks.
I'm well aware that the code here isn't well made so I will need some feedback to make it better.

No other languages are supported right now.


mods/hide-and-seek.lua
```diff
@@ -227,17 +228,7 @@ local function mario_update(m)
...
-    else
-        puX = math_ceil((-8192 + m.pos.x) / 65536)
-    end
-    if m.pos.z >= 0 then
-        puZ = math_floor((8192 + m.pos.z) / 65536)
-    else
-        puZ = math_ceil((-8192 + m.pos.z) / 65536)
-    end
-    if puX ~= 0 or puZ ~= 0 then
+    if m.playerIndex == 0 and (m.pos.x > 32767 or m.pos.x < -32768 or m.pos.z > 32767 or m.pos.z < -32768) then
```
Isaac0-dev: rather than `32768` and `32767` use `0x8000` (which is `32768` in hex).
when replacing `32767`, put `>=` to keep the same logic.
this is represents 180 degrees in sm64 angles, and hex is typically how we write it

src/game/camera.c
```diff
@@ -3109,17 +3110,32 @@ void update_lakitu(struct Camera *c) {
...
+        set_camera_mode(c, c->defMode, 0);
+        gOverrideRomhackCamera = RCO_NONE;
+        return;
+    }
     if (gIsDemoActive) { return; }
 
-    if (gOverrideRomhackCamera == RCO_ALL_EXCEPT_BOWSER) {
-        if (gCurrLevelNum == LEVEL_BOWSER_1 || gCurrLevelNum == LEVEL_BOWSER_2 || gCurrLevelNum == LEVEL_BOWSER_3) {
+    if ((!(gOverrideRomhackCamera == RCO_ALL_INCLUDING_VANILLA || gOverrideRomhackCamera == RCO_ALL_VANILLA_EXCEPT_BOWSER)) &&
+            dynos_level_is_vanilla_level(gCurrLevelNum)) {
```
Isaac0-dev: this line doesn't look great here. prefer
```C
if (gOverrideRomhackCamera != RCO_ALL_INCLUDING_VANILLA && gOverrideRomhackCamera != RCO_ALL_VANILLA_EXCEPT_BOWSER && dynos_level_is_vanilla_level(gCurrLevelNum)) {
```

Isaac0-dev (commented):
here's some feedback. 
another general comment, indentation looks more clean with mutiline conditions when formatted like this:
```C
if ((gOverrideRomhackCamera != RCO_ALL_INCLUDING_VANILLA && gOverrideRomhackCamera != RCO_ALL_VANILLA_EXCEPT_BOWSER) &&
    (dynos_level_is_vanilla_level(gCurrLevelNum))) {
```

Sunketchupm:
Just wanted to make a comment here to say that this should probably be reviewed again

Isaac0-dev:
sorry, just trying to get a build for testing

Isaac0-dev:
I tested it and I'm happy with it, but we don't have other languages still. do you mind please using DeepL to add translations? If someone thinks their language was translated incorrectly they can fix that for us later
@Sunketchupm

Isaac0-dev:
starting to feel like it shouldn't even be called romhack camera if its going to be an option available for enabling at any time

Sunketchupm:
I don't have any name ideas myself

Sunketchupm:
Finished the translations.
As for the name, the name might work since it's the camera best suited for romhacks (as romhacks past 2018 or so were made with aglab camera in mind, which is similar enough to this camera).

Isaac0-dev:
oh I only said that because its now a bit of a weird thing we have about coop, but its fine. if someone else thinks it should be changed they can fix that because i dont mind it

PR #566, opened 2024-12-15
Solve crashing for disconnected players

Flower35:
This PR adresses the following issues:
- show a pop-up message (disconnected, host is down) in case `gNetworkSystem->initialize()` fails
    - for example, if IPv6 is not configured properly for "Direct Connection".
- prevent sudden crashes with empty mod list panel when `(gNetworkType == NT_NONE)`
- allow player to quit the game to main menu while `(gNetworkType == NT_NONE)`

### DIFF
--- a/src/pc/djui/djui_panel_playerlist.c
+++ b/src/pc/djui/djui_panel_playerlist.c
@@ -64,8 +64,12 @@ static void playerlist_update_row(u8 i, struct NetworkPlayer *np) {
 
 void djui_panel_playerlist_on_render_pre(UNUSED struct DjuiBase* base, UNUSED bool* skipRender) {
     if (gDjuiInMainMenu || gNetworkType == NT_NONE) {
-        djui_base_set_visible(&gDjuiPlayerList->base, false);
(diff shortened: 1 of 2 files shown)


Isaac0-dev:
imo if `gNetworkType` is `NT_NONE` during the game, thats a bug

Flower35:
> imo if `gNetworkType` is `NT_NONE` during the game, thats a bug

It can happen when `network_init()` fails in "Direct Connection" mode due to `socket()` returning "Error 97: Address family (IPv6) not supported by protocol".

https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/pc/network/network.c#L144-L155

```
[ERROR] socket_linux.c: socket failed with error 97
[ERROR] network.c: failed to initialize network system
```

---

But yes, that leaves the game in unplayable state with weird things happening (`djui_chat` is not initialized, all message boxes automatically close for some reason, and mods are not loaded).

So perhaps the player should be immediately kicked out of the game to the main menu?

AgentXLP:
If you could kick the player out I would really appreciate it

PR #572, opened 2024-12-17
All Current Star Road Changes

64BitMouse:
This is basically every change so far but in one commit

- You now get sent to Zero Life after dying with a life count below Zero instead of at Zero
- Tuxie (Race Penguin) No longer clips through geometry when finishing the race in Colossal Candy Clutter
- Set Number of lives from 4 to 3 to be more accurate
- Implemented a fully functional Hard Mode as the original hack also had one (Hold L and Host a lobby OR select "Hard Mode" in the mod menu to activate it, uses Backup Slot)
- Koopa the Quick in both Koopa Canyon and Mushroom Mountain Town no longer softlocks himself
- Added a signpost in Castle Grounds hinting at Hard Mode
- Fixed a strange warp node in Zero Life to warp you to the start warp
- Moved a Goomba farther from the Warp Pipe in Hidden Palace Finale to prevent him from hitting you immediately (Also makes Hard Mode Possible)
- Fixed a faulty warp in Piranha Plant Pond
- Added name coloring to the mod description for easier readability for credits
- Altered the "SMSRSpaceBox" object's behavior to have a home using SET_HOME() so that Star 2, "Climb Through the Machinery", in Starlight Runway was casually possible (was previously just metal boxes with a different model)


AgentXLP:
I think this looks good!

PR #575, opened 2024-12-19
Add HOOK_HEALED_MARIO and HOOK_HURT_MARIO

Sunketchupm:
Not very useful hooks but someone might find it useful


PeachyPeachSM64:
that would be nice to also have the value of the heal/damage taken as a parameter

Sunketchupm:
In all cases you already know how much has been healed/taken damage and are able to read how much heal/hurt counter is left. I don't think it's necessary.

PR #584, opened 2024-12-26
Autogen correctly handle pointers

kermeow:
Back again with ANOTHER pull request!!!!!

This time the autogen was broken and nobody noticed this for however long until now!
Now they are handled correctly with the correct LOTs for CObjects.

I also have temporarily removed Mtx from being recognised by autogen since it's only used in one obscure place (thanks to my other pull request) and isn't actually implemented yet as a LOT.

throwMatrix is revolutionising coop modding


AgentXLP:
I'm pretty sure its called throwMatrix because it's used to transform Mario when he's being thrown by King Bob-omb or one of those purple ball motherfuckers

AgentXLP:
CHUCKYA

PR #585, opened 2024-12-27
[Experimental] Reduce the size of Windows builds by cleaning the MinGW debug info

Flower35:
Experimental update to help reduce the size of Windows builds.
- removes the "spam" (around 50 KB) of `"Built by MSYS2 project"` strings from the executable (these get deduplicated on Linux, but not on Windows)
- significantly drops the size of `"coop.map"` by leaving only the function labels (other info is ignored on the crash screen anyways)
- add a separate database of symbols (global and static function labels + global and static variable offsets)

Pros:
- drops the size of `"sm64coopdx.exe"` from around 72 MB to 33 MB
- drop the size of `"coop.map"` from around 12 MB to 3 MB
- adds separate symbol database for `x64dbg`, around 6 MB
- total saved disk space: `((72 - 33) + (12 - 3) - 6) = 42 MB`

Cons:
- the build would fail on 32-bit targets
  - (I think these are not officially supported at the moment, but could rework the Python script)
- in case someone uses `addr2line` or `gdb`, these will no longer work
  - (stepping through the source code would still only make sense if the repository path is identical for other devs)
  - (while working with DEVELOPMENT or DEBUG builds, the `"sm64coopdx.exe.bak"` copy can stil be used) 
- `cv2pdb` tool might be used instead (for debugging with Visual Studio or `x64dbg`), but it is not a part of MSYS2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -869,6 +869,7 @@ ifeq ($(WINDOWS_BUILD),1)
   ifeq ($(CROSS),)
     LDFLAGS += -no-pie
   endif
+  LDFLAGS += -T windows.ld
 else ifeq ($(TARGET_RPI),1)
   LDFLAGS := $(OPT_FLAGS) -lm $(BACKEND_LDFLAGS) -no-pie
 else ifeq ($(OSX_BUILD),1)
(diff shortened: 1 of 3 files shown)


AgentXLP:
You should make this only happen if the `DEVELOPMENT` flag is disabled, it would still be useful to use GDB

PR #586, opened 2024-12-28
Hard Mode Water Bug

64BitMouse:
Fixed an issue with water not killing you when taking damage underwater


AgentXLP:
I typed in the fix and co-authored you so you got credited

PR #591, opened 2024-12-29
This adds a new hook HOOK_MARIO_OVERRIDE_GEOMETRY_INPUTS

Amy54Desu:
This adds HOOK_MARIO_OVERRIDE_GEOMETRY_INPUTS as a new hook with a return value, return false to get override the update_mario_geometry_inputs. This has a lot of applications, such as PROPERLY walking on water, (since floorHeight is reset in this function) as well as swimming in the air (as m.waterLevel is set here, BEFORE_MARIO_UPDATE is called before this function, and MARIO_UPDATE is after), as well as custom NULL floor handlers (so you don't always have to do the coop teleport to other player)
sm64coopdx_FAETe6IJCC
sm64coopdx_k8UqW2qLU9


PeachyPeachSM64:
please do not insert new hooks in the middle of the existing list
it will break compatibility with some existing mods

AgentXLP:
No hooks.md update

Amy54Desu:
> please do not insert new hooks in the middle of the existing list it will break compatibility with some existing mods

alright, I'm puttin it at the end

Amy54Desu:
> No hooks.md update

sorry, I didn't hear you over the big battle I'm having with autogen.sh

Amy54Desu:
> please do not insert new hooks in the middle of the existing list it will break compatibility with some existing mods

should be OK now

Issue #564, opened 2024-12-14
Exiting to Castle Advances Gameplay by a Single Frame Before Freezing Again in Singleplayer


sher-b:
Normally, all gameplay is paused when pausing and exiting to castle in the original game. 
https://github.com/user-attachments/assets/25cc0146-6422-4b0f-9a71-43bd6c9de24b

Although in CoopDX, gameplay moves just a single frame forward before actually pausing gameplay properly.
https://github.com/user-attachments/assets/06a41f8b-31a1-457a-ba41-032f4c71430b

To better visualize this, pay attention to the red coin in the CoopDX clip.

Before:
![image](https://github.com/user-attachments/assets/2b4b072a-b93c-4927-87c9-afbce4a1ea1c)

After:
![image](https://github.com/user-attachments/assets/0dd8a1cd-4dc0-4215-acf4-313ca42b9061)

And to show that this affects the player as well, watch this clip:
https://github.com/user-attachments/assets/8c21dac8-a17c-42ef-9bbf-4f9eef87a48c

AgentXLP:
Weird

Flower35:
Alright, found the reason why extra frame is processed on the pause screen.

First, this is executed when player presses START and pausing is allowed:
https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/game/level_update.c#L1318-L1323

Then, the game enters 2nd game loop iteration.

Now, `network_check_singleplayer_pause()` still returns `FALSE`, because `(gMenuMode == (-1))`:
https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/pc/network/network_utils.c#L56-L58

And when this code in `update_level()` is reached, the game executes:
(1) `changeLevel = play_mode_normal();`
(2) `sCurrPlayMode` has not changed from `PLAY_MODE_PAUSED`
(3) `changeLevel = play_mode_paused();`
(4) `{ set_menu_mode(RENDER_PAUSE_SCREEN); }`
https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/game/level_update.c#L1702-L1710

(normally this is so that the gameplay continues, while the pause screen is rendered in multiplayer, or when one of the loaded mods has property `pausable: false`)

---

[...]

PR #596, opened 2024-12-31
Stop the gameplay from advancing a single frame on the pause screen in singleplayer

Flower35:
Addresses the issue: https://github.com/coop-deluxe/sm64coopdx/issues/564

`network_check_singleplayer_pause()` now returns `true` for the moment when player wants to pause in singleplayer, and the pause screen should be displayed on the next frame -- when `pressed_pause()` in `play_mode_normal()` returned `true`, but `play_mode_paused()` was not called yet to change `gMenuMode`.

---

### Technical details (code flow before the bugfix)

First, this code executed when player presses START and pausing is allowed:
https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/game/level_update.c#L1318-L1323

Then, the game enters 2nd game loop iteration.

Now, `network_check_singleplayer_pause()` still returns `FALSE`, because `(gMenuMode == (-1))`:
https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/pc/network/network_utils.c#L56-L58

And when this code in `update_level()` is reached, the game executes:
(1) `changeLevel = play_mode_normal();`
(2) `sCurrPlayMode` has not changed from `PLAY_MODE_PAUSED`
(3) `changeLevel = play_mode_paused();`
(4) `{ set_menu_mode(RENDER_PAUSE_SCREEN); }`
https://github.com/coop-deluxe/sm64coopdx/blob/f85b8419afc6266ac0af22c5723eebe3effa1f7d/src/game/level_update.c#L1702-L1710

(normally this is so that the gameplay continues, while the pause screen is rendered in multiplayer, or when one of the loaded mods has property `pausable: false`)

---

### DIFF
--- a/src/pc/network/network_utils.c
+++ b/src/pc/network/network_utils.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
 #include "network_utils.h"
+#include "game/camera.h"
 #include "game/level_update.h"
 #include "game/mario_misc.h"
 #include "pc/mods/mods.h"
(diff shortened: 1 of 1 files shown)


AgentXLP (approved):
Looks good

PR #597, opened 2024-12-31
Synchronize instructions in `network_init()` (server) with `network_receive_join()` (client)

Flower35:
Synchronizes the order of operations when a new game starts:
- `network_player_connected(NPT_LOCAL, ...)` called before `smlua_init()`
  - so that, for example, `gNetworkPlayers[0]` has valid fields then Lua scripts are initialized
  - previously, `gNetworkPlayers[0]` was empty on the server side
- `djui_chat_box_create()` called before `smlua_init()`
  - allows `djui_chat_message_create()` to be called when Lua mod is initialized, or in `HOOK_ON_MODS_LOADED`
  - previously it would only work on the client side, because server-side chat-box was created after loading all mod files
- `gChangeLevelTransition` was set twice for no reason
  - first conditionally in `network_init(NT_SERVER)`
  - then unconditionally in `djui_panel_do_host()`

### DIFF
--- a/src/pc/djui/djui_panel_host_message.c
+++ b/src/pc/djui/djui_panel_host_message.c
@@ -2,7 +2,6 @@
 #include "djui.h"
 #include "djui_panel.h"
 #include "djui_panel_menu.h"
-#include "djui_panel_modlist.h"
 #include "pc/network/network.h"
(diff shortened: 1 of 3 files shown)


AgentXLP (approved):
Looks good

PR #606, opened 2025-01-03
Fix smlua memory leaks

PeachyPeachSM64:
make smlua great again

this could potentially affect performances, but should get rid of lag/freeze issues

details in comments


AgentXLP:
Good

PR #607, opened 2025-01-04
smlua improvements

Isaac0-dev:
changes to smlua to improve performance, and reduce memory usage. 
returned to using a pool for cobjects but this time in the c api.


kermeow:
this doesn't build without development flag
is it supposed to be a draft?

src/pc/lua/smlua_cobject.c
```diff
@@ -324,26 +328,28 @@ struct LuaObjectField* smlua_get_custom_field(lua_State* L, u32 lot, int keyInde
 /////////////////////
 
 static int smlua__get_field(lua_State* L) {
-    LUA_STACK_CHECK_BEGIN();
+    LUA_STACK_CHECK_BEGIN_NUM(1);
 
-    CObject *cobj = lua_touserdata(L, 1);
+    const CObject *cobj = lua_topointer(L, 1);
```
kermeow: Should this not be `lua_touserdata` like before?
`lua_topointer` accepts more types than we should be and there is already no typechecking.

src/pc/lua/smlua_cobject.c
```diff
@@ -406,18 +410,19 @@ static int smlua__get_field(lua_State* L) {
...
 
+    LUA_STACK_CHECK_END();
     return 1;
 }
 
 static int smlua__set_field(lua_State* L) {
     LUA_STACK_CHECK_BEGIN();
 
-    CObject *cobj = lua_touserdata(L, 1);
+    const CObject *cobj = lua_topointer(L, 1);
```
kermeow: Same question about `lua_touserdata`

src/pc/lua/smlua_cobject.c
```diff
@@ -496,37 +501,27 @@ static int smlua__set_field(lua_State* L) {
...
-            }
-        }
-    }
-    return 0;
-}
-
 static int smlua_cpointer_get(lua_State* L) {
-    CPointer *cptr = lua_touserdata(L, 1);
-    const char *key = smlua_to_string(L, 2);
+    const CPointer *cptr = lua_topointer(L, 1);
```
kermeow: Same thing with `lua_touserdata` and same null check issue.

src/pc/lua/smlua_cobject.c
kermeow: Adding to this, maybe this is a good PR to add the aforementioned missing typecheck (and null check) to fix the otherwise quick and easy segfault.

src/pc/lua/smlua.h
```diff
@@ -24,7 +24,8 @@
 #define LOG_LUA_LINE_WARNING(...)  { if (!gLuaActiveMod->showedScriptWarning) { gLuaActiveMod->showedScriptWarning = true; smlua_mod_warning(); snprintf(gDjuiConsoleTmpBuffer, CONSOLE_MAX_TMP_BUFFER, __VA_ARGS__), sys_swap_backslashes(gDjuiConsoleTmpBuffer), djui_console_message_create(gDjuiConsoleTmpBuffer, CONSOLE_MESSAGE_WARNING); } }
 
 #ifdef DEVELOPMENT
-#define LUA_STACK_CHECK_BEGIN() int __LUA_STACK_TOP = lua_gettop(gLuaState)
+#define LUA_STACK_CHECK_BEGIN_NUM(n) int __LUA_STACK_TOP = lua_gettop(gLuaState) + (n)
```
kermeow, line 27: Build failure comes from this only being defined in development builds

src/pc/lua/smlua_cobject.c
Isaac0-dev: its an optimization. we can assume the pointer will always be userdata since Lua can't pass in a different type, so we can skip the type check.

kermeow: This is not true and there is no benefit to using `lua_topointer` over `lua_touserdata`. Each function has a completely different use-case. You even continued to use `lua_touserdata` in other methods for CObjects and CPointers.
The only real difference is that `lua_topointer` also accepts functions, threads, and tables.

Additionally, Lua can actually pass in any type.
If a table or function (which are accepted by `lua_topointer`) are passed, this code attempts to access their respective GCObject (from Lua source). Since the data at the same offset as a CObject's LOT is not a valid LOT, this causes a segfault.
If any other type is passed, `lua_topointer` and `lua_touserdata` would both return NULL, also causing a segfault.

kermeow (approved):
Awesome
God bless America

PR #614, opened 2025-01-06
Star Road Blueberry Bully Fix

64BitMouse:
Changed the bhv name in script.c and bhv_overrides.lua to match behaviordata.c


AgentXLP:
Fixed on my branch locally

PR #616, opened 2025-01-07
Allow only valid versions to trigger update notices

Cooliokid956:


AgentXLP:
No longer necessary now that the version number is read from the GitHub

Isaac0-dev:
it would still be a good idea to do this

PR #617, opened 2025-01-07
Fixed MacOS Resource Path Location

KirbyKid256:
Wanted to play this with my dad so I compiled it for MacOS since we don't have many Windows computers. I noticed the path to get the mods and stuff was different from what was created by the MakeFile, so I fixed that. I also added a new function called `sys_resource_path()` in place of some `sys_exe_path_dir()` instances to grab resources like DynOS, mods, and languages. I have not tested this on a Windows 32 bit or Linux, although I assume it still works.

Side Note: When compiling for MacOS with these changes on the dev branch, `--build-id=none` in the MakeFile caused an error so I removed it to get it to compile. However, it was fine to keep in when I tested it on Windows. I don't know if it causes an error on Linux or continues compiling, but it errored on my MacOS Intel.

### DIFF
--- a/src/pc/platform.c
+++ b/src/pc/platform.c
@@ -332,19 +332,19 @@ const char *sys_user_path(void) {
 
 const char *sys_resource_path(void)
 {
-    #ifdef __APPLE__ // Kinda lazy, but I don't know how to add CoreFoundation.framework
+#ifdef __APPLE__ // Kinda lazy, but I don't know how to add CoreFoundation.framework
     static char path[SYS_MAX_PATH];
     if ('\0' != path[0]) { return path; }
 
     const char *exeDir = sys_exe_path_dir();
     char *lastSeparator = strrchr(exeDir, '/');
     if (lastSeparator != NULL) {
-        const char folder[10] = "/Resources";
+        const char folder[] = "/Resources";
         size_t count = (size_t)(lastSeparator - exeDir);
         strncpy(path, exeDir, count);
-        return strcat(path, folder);
+        return strncat(path, folder, sizeof(path) - 1 - count);
     }
-    #endif
+#endif
 
     return sys_exe_path_dir();
 }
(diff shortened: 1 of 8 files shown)


src/pc/platform.c
```diff
@@ -326,6 +330,25 @@ const char *sys_user_path(void) {
     return path;
 }
 
+const char *sys_resource_path(void)
+{
+    #ifdef __APPLE__ // Kinda lazy, but I don't know how to add CoreFoundation.framework
```
PeachyPeachSM64: don't indent #ifdef/#endif for readability

src/pc/platform.c
```diff
@@ -326,6 +330,25 @@ const char *sys_user_path(void) {
...
+    static char path[SYS_MAX_PATH];
+    if ('\0' != path[0]) { return path; }
+
+    const char *exeDir = sys_exe_path_dir();
+    char *lastSeparator = strrchr(exeDir, '/');
+    if (lastSeparator != NULL) {
+        const char folder[10] = "/Resources";
+        size_t count = (size_t)(lastSeparator - exeDir);
+        strncpy(path, exeDir, count);
+        return strcat(path, folder);
```
PeachyPeachSM64: Don't specify the length of const strings, let the compiler compute it. Some versions of GCC refuse to compile since you don't account for the null terminator ("/Resources" is actually 11 chars long), which leads to undefined behavior.
Also `strcat` is unsafe. Prefer using `strncat(path, folder, sizeof(path) - 1 - count)`.

PeachyPeachSM64 (changes requested):

src/pc/platform.c
KirbyKid256: For this one, would I just remove the 10 from `folder[10]` to result in just `folder[]` or should I just use `SYS_MAX_PATH` in place of 10?
Also thanks for the help

PeachyPeachSM64: just remove the 10

PR #626, opened 2025-01-12
Fix a linker error with macOS

EmeraldLoc:
During the linking stage of compilation, an error is thrown as `--build-id` appears to not exist in macOS. I've simply gone ahead and removed the added ld flags if the project is being compiled on macOS. The game appears to launch and run fine with this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -990,7 +990,9 @@ endif
 IS_DEV_OR_DEBUG := $(or $(filter 1,$(DEVELOPMENT)),$(filter 1,$(DEBUG)),0)
 ifeq ($(IS_DEV_OR_DEBUG),0)
   CFLAGS += -fno-ident -fno-common -ffile-prefix-map=$(PWD)=. -D__DATE__="\"\"" -D__TIME__="\"\"" -Wno-builtin-macro-redefined
-  LDFLAGS += -Wl,--build-id=none
+  ifeq ($(OSX_BUILD),0)
+    LDFLAGS += -Wl,--build-id=none
+  endif
 endif
 
 # Prevent a crash with -sopt


Isaac0-dev:
but the reason that line is there is to attempt to keep manually built executables as predictable as possible. will mac builds consistently produce the same executable every time even with this change?

EmeraldLoc:
Thats good to know. How are you comparing the binary? When I go and compile coop, put that into a SHA-256 convertor, save that, then delete only the executable file, and recompile that file, it results in the same SHA-256. Running make clean or distclean and then recompiling it and comparing the hashes results in them being different, thus not accomplishing the goal you have in mind. However. their may be metadata related to time and other info that would cause this to happen, so I don't know if that is a reliable way to compare.

Isaac0-dev:
yeah that's how I test the builds are the same, and it seems that this change breaks the intended result.. the hope is to get the same build every time you make clean

EmeraldLoc:
Alright, so apparently macOS' linker does not include a build id by default anyways, so something else is causing the issue (I think?). I've been unable to reproduce 2 of the same executables, they're always different. The only solution that might work (that I haven't tried as I don't want to deal with uninstalling it if it turns out it doesn't work) is getting a GNU linker that allows the build id to be passed, however this would be rather inconvenient (but doable if necessary) for us Mac users. Pretty stumped on this one other than that though unfortunately.

Isaac0-dev:
@EmeraldLoc can you try using `-Wl,-no_uuid` and `-Wl,-x` separately, and see if either produce a consistent build? if neither do alone, i guess try both combined (`-Wl,-no_uuid -Wl,-x`)?
i'm hoping only one of those is needed though. also don't forget to run `make clean` between builds to ensure its a proper test.

EmeraldLoc:
Running 1make clean1 in between each test, trying the flags `-Wl,-no_uuid`, `-Wl,-x`, and `-Wl,-no_uuid -Wl,-x`, none of which appear to output the same SHA-256 binary. My workflow is compiling the app, then, without doing any changes to the location of the binary, dragging the binary into the SHA256 checker at [this site](https://emn178.github.io/online-tools/sha256_checksum.html), and with that compiling my next binary, and comparing the 2 outputs of that sha256 checker. Dragging in the same binary results in the same output, and dragging in a copied binary results in the same output, but dragging in a differently compiled binary results in a different output.

Isaac0-dev:
this will mean that people can't build on macos manually and have coopnet compatibility, but i guess it's a small price to pay.

PR #631, opened 2025-01-14
added different rendering options for when game window is not focused

iZePlayzYT:
this is like the "Mute When Window Unfocused" audio setting but for the graphics. it highly reduces the ram, cpu and gpu usage while unfocused.


iZePlayzYT:
this is like the "Mute When Window Unfocused" audio setting but for the graphics. it highly reduces the ram, cpu and gpu usage while unfocused

Isaac0-dev (commented):
I'm really skeptical about adding this because some mods rely on the rendering hooks being called.

iZePlayzYT:
@Isaac0-dev do you have any ideas how i can improve it?

Isaac0-dev:
still run `geo_process_root` as normal? perhaps interpolation could be disabled instead, since that wouldn't change how often hooks are called

PR #633, opened 2025-01-14
Raspberry Pi CoopNet Support

coderpro1234-2:
I have added the necessary library file to compile sm64coopdx with CoopNet support for Raspberry Pi (arm64 and arm devices). I have also added automatic detection for these devices, and it builds for the Raspberry Pi if it finds a suitable device. I have tested this on a Raspberry Pi 5 and Raspberry PI 4, and it compiles without any errors, and CoopNet works on them.


Isaac0-dev:
where was the source code found for these libraries?

coderpro1234-2:
The source for the CoopNet libraries is from your repository (https://github.com/Isaac0-dev/coopnet) and the source for the libjuice is from the official repository (https://github.com/paullouisageneau/libjuice). I have then compiled these for both arm64 and arm computers (for the Raspberry Pi)

martenmario64:
ive managed to play sm64ex-coop with wine on a raspberry pi 4

Makefile
```diff
@@ -966,9 +976,9 @@ ifeq ($(COOPNET),1)
     endif
   else ifeq ($(TARGET_RPI),1)
     ifneq (,$(findstring aarch64,$(machine)))
-      LDFLAGS += -Llib/coopnet/linux -l:libcoopnet-arm64.a -l:libjuice.a
+      LDFLAGS += -Llib/coopnet/linux -l:libcoopnet-arm64.a -l:libjuice-arm64.a
     else
-      LDFLAGS += -Llib/coopnet/linux -l:libcoopnet-arm.a -l:libjuice.a
+      LDFLAGS += -Llib/coopnet/linux -l:libcoopnet-arm.a -l:libjuice-arm.a
```
theofficialgman, line 981: this detection will not work on most arm32 systems as these systems generally use 64bit kernels these days even if they have 32bit userspace. this will result in the findstring aarch64 being true and adding the link flags for the arm64 library even on arm32 userspace.

theofficialgman: A more robust method employed in nearly every makefile that needs to do 32 vs 64 detection after first determining the base architecture (arm) would be to check SIZEOF_VOID_P (typically by running the compiler)

Isaac0-dev: feel free to make a pull request for that

PR #634, opened 2025-01-18
Fixed and clarified misleading --coopnet argument description

iZePlayzYT:


Cooliokid956:
it is stylized CoopNet

iZePlayzYT:
> it is stylized CoopNet

fixed now

PR #637, opened 2025-01-19
New config option: Skip DynOS binary file generation on startup

Cooliokid956:
it is off by default, but the common player may want to turn it on to improve loading times


Cooliokid956:
just a few more then i'll get to bed

PR #639, opened 2025-01-19
Headless mode as a CLI option

Cooliokid956:
the `HEADLESS` compile flag has been entirely replaced with the `--headless` argument,
What happens when you enable Headless mode?
- Windows console opens if it hadn't already
- Render/Window APIs are overwritten
- Info-level logs are enabled

Tested, works with both `--server <port>` and `--coopnet`

this also adds two spaces after `--console`


Isaac0-dev:
imo the compile flag should remain, the headless argument is fine, just both should be available

PR #642, opened 2025-01-20
Dynamic Custom Model Allocation

theclashingfritz:
Added support for dynamic custom model allocation. The limit is also raised and hopefully mods will no longer have an issue with it.

Please do bug test this for any issues.


Isaac0-dev (commented):
i think using the growing array functions that we have would be beneficial here

FluffaMario:
This'd actually be really cool for mod devs to be honest since some hosts tend to overload the mod list with hundreds of character mods, had to have a stern talking to with my older brother about this because he thought it would work like GMod and ended up crashing his own game from it a few times

PR #659, opened 2025-02-12
[WIP] FreeBSD compatibility

Krush206:
Some small adjustments for compatibility with FreeBSD systems. The game loads, but crashes upon hosting a LAN server.
```
Assertion failed: ((idx <= 255 + 1) && "upvalue index too large"), function index2addr, file lapi.c, line 76.
Abort (core dumped)
```
Result is identical for both, GCC and Clang.


Krush206:
This is done. Can confirm Lua is working and SIGSEGV is eliminated.

NOTE: Discord and CoopNet aren't included. I displaced CoopNet with [ZeroTier One](https://zerotier.com), and Discord's SDK doesn't generate a valid FreeBSD executable.

Isaac0-dev:
why did you change Lua version?

Krush206:
Unlike Lua 5.3.5, Lua 5.4.7 doesn't throw the
```
Assertion failed: ((idx <= 255 + 1) && "upvalue index too large"), function index2addr, file lapi.c, line 76.
Abort (core dumped)
```
error. Updating is what solved the issue, though I wonder the error lies in `luaconf.h`. Haven't spent much time messing `luaconf.h`.

I played a full co-op game without crashes and issues. Lua 5.4.7 loads and runs scripts as expected.

Isaac0-dev:
honestly I'd like to update to Lua 5.4, I'm just skeptical about it i guess. I'll have a look at what may have changed in 5.4 that needs to be considered while updating.

Isaac0-dev (changes requested):
I've done some testing and it seems that Lua 5.4 works, but has a few strange bugs that make some mods incompatible.
This means we're better off not updating to 5.4 on all platforms, but if this is the only fix that can be found for freebsd, its fine for just this platform for now.
But, please don't add the unnecessary header files that coop doesn't use

djoslin0:
WIP & Stale. Will people upkeep FreeBSD? (likely not)

PR #669, opened 2025-02-15
[QOL] Scripts to download the latest release

voidxO7:
It downloads (and overwrites) the files inside the same directory as where the script is ran. Linux's ".sh" needs more testing as I only run Windows. Requires Python 3.6 or higher to run.

The ".bat" and ".sh" files check for and install the "requests" module then run "download.py".


voidxO7:
oh, I'm not sure without deleted the pull request if I can fix the spelling error in the ".sh" file name

voidxO7:
> A.) Make a launcher that has an auto update feature, this launcher could later be adapted to be a mod downloader/manager. Or B.) Just make a button in the djui that updates the game. If neither of those are an option, we could do it how dolphin emulator does it, have a separate executable that downloads and installs the update.

Yes, I'd like that too, but I'm not a person who really knows how to code. I just threw this together as a substitute for the time being. Think of it like a band aid, this is here to temporarily close an issue until the body (the people who know how to code) can work out a solution (if they decide to).

AgentXLP:
We'll figure something out

PR #674, opened 2025-02-18
Revert romhack camera changes

Paradox9955:
In my honest opinion, this should have never been changed. Doing this forces the x axis to invert itself upon entering a custom level. 

Think about it like this, imagine playing flood/Flood Expanded and you enter one of the custom levels, when trying to move your camera you realize it has suddenly been inverted and you end up failing because of your camera controls. 

We shouldn't be forced to go to the camera options and enable/disable the `invert x` option each time we enter/leave a level.


Isaac0-dev:
my question is why was the change made in the first place? what was it trying to fix?

Sunketchupm:
#551

PR #675, opened 2025-02-18
allow Lua to manipulate displaylists and vertices

Isaac0-dev:
thanks to @ManIsCat2 for the initial proof of concept, and for adding `gfx_set_combine_lerp`.

This allows modders to manipulate the displaylist of a model via Lua mid-game.
As an extension of previous commits that have enabled traversing the graph, this allows Lua to access `graphNode.displayList`.
`gfx_parse` allows for traversing the display list, taking a function for a callback, most importantly allowing access to vertices through `gfx_get_vtx`.
Displaylist words are immutable as a precaution for memory safety. 
`gfx_set_command` can execute graphics commands directly applied to the display list pointer provided. most of the commands that dynos supports are available.


AgentXLP:
WHAT THE SHIT

SonicTheHedgeHog64:
🤔

SonicTheHedgeHog64:
How would someone use this?

src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -106,3 +106,116 @@ void set_skybox_color(u8 index, u8 value) {
...
+    lua_State* L = gLuaState;
+    while (true) {
+        u32 op = cmd->words.w0 >> 24;
+        switch (op) {
+            case G_DL:
+                if (C0(16, 1) == 0) {
+                    gfx_parse((Gfx *) cmd->words.w1, func);
+                } else {
+                    cmd = (Gfx *) cmd->words.w1;
+                    --cmd;
```
PeachyPeachSM64, line 128: uh oh, out of bounds?
also, if I understand correctly, these are special cases that shouldn't call the callback, but call it anyway because the code just continues after the switch
shouldn't the callback be in a `default` branch of the switch instead?

src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -106,3 +106,116 @@ void set_skybox_color(u8 index, u8 value) {
...
+#ifdef F3DEX_GBI_2
+    u16 numVertices = C0(12, 8);
+    u16 destIndex = C0(1, 7) - C0(12, 8);
+#elif defined(F3DEX_GBI) || defined(F3DLP_GBI)
+    u16 numVertices = C0(10, 6);
+    u16 destIndex = C0(16, 8) / 2;
+#else
+    u16 numVertices = (C0(0, 16)) / sizeof(Vtx);
+    u16 destIndex = C0(16, 4);
+#endif
```
PeachyPeachSM64: are these really necessary? I don't think coop uses other microcodes
and even if it did, dynos stuff won't work anyway

src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -106,3 +106,116 @@ void set_skybox_color(u8 index, u8 value) {
...
+    if (offset >= numVertices) { return NULL; }
+    if (destIndex >= MAX_VERTICES) { return NULL; }
+    return &((Vtx *) cmd->words.w1)[offset];
+}
```
PeachyPeachSM64: ok, some points here:
- it seems you can retrieve only one vertex per function call, why can't you get the whole buffer?
- the `destIndex` and `MAX_VERTICES` stuff is tied to the rendering engine vertex buffer limitation, it's useless here
- no NULL check on `cmd->words.w1`

src/pc/lua/utils/smlua_gfx_utils.h
```diff
@@ -47,4 +48,27 @@ u8 get_skybox_color(u8 index);
...
+void gfx_set_prim_color(Gfx* gfx, u8 m, u8 l, u8 r, u8 g, u8 b, u8 a);
+/* |description|Sets the display list environment color.|descriptionEnd| */
+void gfx_set_env_color(Gfx* gfx, u8 r, u8 g, u8 b, u8 a);
+/* |description|Sets the display list fog color.|descriptionEnd| */
+void gfx_set_fog_color(Gfx* gfx, u8 r, u8 g, u8 b, u8 a);
+/* |description|Copies a player light parameters.|descriptionEnd| */
+void gfx_copy_lights_player_part(Gfx* gfx, u8 part);
+/* |description|Sets the display list combine mode.|descriptionEnd| */
+void gfx_set_combine_lerp(Gfx* gfx, u32 a0, u32 b0, u32 c0, u32 d0, u32 Aa0, u32 Ab0, u32 Ac0, u32 Ad0, u32 a1, u32 b1, u32 c1, u32 d1,	u32 Aa1, u32 Ab1, u32 Ac1, u32 Ad1);
+
```
PeachyPeachSM64, line 57: This looks cool and all, but...
- already 400+ symbols for a very low level api
- there is some utility functions but some of them still need you a great understanding of the gbi.h file just to fill in the arguments (gfx_set_combine_lerp)

I know this is still wip and there isn't much more options, but maybe it would be good to look into a `gfx_set_command(Gfx *gfx, const char *command)` function and make use of the dynos parser to reduce the global table overload and the complexity

another idea: the parse function could provide the vertex buffer or triangles or some other common things depending on the op it processes, like how the level_parse function provides information depending on the current level command

Cooliokid956:
got an idea a while ago:
gfx_parse_interp(gfx, callback(gfx, op, delta))

execute gfx_parse for that frame, then store the dl pointer and callback in an array of interps, iterating through these similarly to interp huds

src/pc/lua/utils/smlua_gfx_utils.c
avereyishere: where's rej?

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: well i don't know, I was just trying to make sure this function iterates the display list correctly by copying how its done in `gfx_pc.c`
but yeah it's probably a good idea to move the callback inside a default branch.

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: nah its not necessary, i was just doing it in case, i'll remove it

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: you can only get one vertex per call because we don't have any other way of sanity checking it every time a user tries to apply offset
its also faster performance wise than having a new smlua type that can act as an array of vertices, since functions can push data faster than the traditional smlua struct reads/writes

there's no null check on `cmd->words.w1` because its not a pointer? and if the op isn't `G_VTX` it won't run anyway... am i misunderstanding your comment?

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: also what do you mean by out of bounds? that code mimics what i found in `gfx_pc.c`, is it wrong?

PeachyPeachSM64: if `cmd` points to the beginning of a display list, `cmd--` is out of bounds
in `gfx_pc.c`, there is no code after the `cmd--`, so when `cmd` is incremented again, it correctly points to the next instruction to process

src/pc/lua/utils/smlua_gfx_utils.c
PeachyPeachSM64: `cmd->words.w1` holds the pointer to a Vtx array, which can be NULL (for example `gsSPVertex(NULL, 0, 0)`)

Isaac0-dev: oh right, thanks for explaining

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: oh yeah that line is there to prevent the incrementation of cmd, but i'll be passing an out of bounds pointer to the callback...
now that i have pushed the changes with the default branch it should be fine?

PeachyPeachSM64: should be good

src/pc/lua/utils/smlua_gfx_utils.h
Isaac0-dev: added `gfx_set_command`, which allows you to use any of the simple macros that dynos uses.

Isaac0-dev: I think this is a pretty advanced feature anyway, so I think it's still worth adding just for those who know what they're doing.
for functions like `gfx_set_combine_lerp`, its only there for the people who really know what it does, so yeah. very advanced feature.

Isaac0-dev:
@Cooliokid956 I'm not entirely convinced that it would be helpful, and at least in the way you described it I can't think of any applications for that.
I'm imagining that you'd want something like that so you can manually interpolate vertices as they are changed, but if i'm not mistaken the current interpolation code should do that already...
Have i misunderstood your comment perhaps, and could you explain what applications that could have?

Isaac0-dev:
Gonna merge this feature, contributors are still welcome to add to/change this feature via pull requests to the dev branch, and I'll still be changing it on the dev branch where necessary.

PR #676, opened 2025-02-20
Add various hooks

Sunketchupm:
After my previous PR for HOOK_HEALED_MARIO and HOOK_HURT_MARIO got closed, I took the time to add in a new hook that I'd also like to see added

This PR includes:
HOOK_HEALED_MARIO (Params: `MarioState` m, `MarioHealType` healType; Returns: None)
HOOK_HURT_MARIO (Params: `MarioState` m, `MarioHurtType` hurtType; Returns: None)
HOOK_AFTER_QUARTER_STEP (Params: `MarioState` m, `integer` stepType, `integer` stepResult, `integer` stepNumber; Returns: `integer` stepResult`)

I may add more hooks in the future but these are the main 3 I want in coop


Isaac0-dev:
can you please explain why these hooks are needed? and provide an example of something you may want to do that you can't already do with the current hooks?

Sunketchupm:
`HOOK_AFTER_PHYS_STEP` allows me to prevent ledgegrabs and bonks more cleanly and naturally without having to recreate the entire step system. My current use for this is to create an object that prevents bonking altogether in order to bounce Mario correctly.

`HOOK_HEALED_MARIO` and `HOOK_HURT_MARIO` can be replaced much easier and right now my main use for them is to make a custom health system. This can already be done with `HOOK_BEFORE_MARIO_UPDATE` and `HOOK_MARIO_UPDATE` I believe, however specific hooks for heals/hurt just makes sense for me for these uses.

This isn't related but I figured out why we don't have a hook that prevents objects from unloading.

Isaac0-dev:
just a heads up, you could already do `HOOK_AFTER_PHYS_STEP` like this for air steps (not tested, consider it psuedocode):
```lua
local sStep = false
hook_event(HOOK_BEFORE_PHYS_STEP, function(m, stepType, stepArgs)
    if sStep then return end
    if stepType ~= STEP_TYPE_AIR then return end
    sStep = true
    local stepResult = perform_air_step(m, stepArgs)
    sStep = false
    if stepResult == AIR_STEP_HIT_WALL then
        return AIR_STEP_NONE
    end
    return stepResult
end)
```
but it won't hurt to have it anyway. 
the hooks for damage and healing seems a bit unnecessary to me, since they're very use case specific..
not saying they need to be removed, i'm just skeptical.

src/game/mario_actions_automatic.c
```diff
@@ -365,26 +365,19 @@ s32 perform_hanging_step(struct MarioState *m, Vec3f nextPos) {
...
+    s32 stepResult = HANG_NONE;
+    if (floor == NULL) { stepResult = HANG_HIT_CEIL_OR_OOB; }
+    else if (ceil == NULL) { stepResult = HANG_LEFT_CEIL; }
+    else if (ceilHeight - floorHeight <= 160.0f) { stepResult = HANG_HIT_CEIL_OR_OOB; }
+    else if (ceil->type != SURFACE_HANGABLE) { stepResult = HANG_LEFT_CEIL; }
+    else {
+        ceilOffset = ceilHeight - (nextPos[1] + 160.0f);
+        if (ceilOffset < -30.0f) { stepResult = HANG_HIT_CEIL_OR_OOB; }
+        if (ceilOffset > 30.0f) { stepResult = HANG_LEFT_CEIL; }
+    }
```
Isaac0-dev, line 377: why change this code? it was much more readable before.

src/game/mario.c
```diff
@@ -1686,6 +1686,17 @@ void set_submerged_cam_preset_and_spawn_bubbles(struct MarioState *m) {
     }
 }
 
+void update_burning_health_common(struct MarioState* m) {
+    m->health -= 10;
+    smlua_call_event_hooks_mario_param_and_int_ret_bool(HOOK_HURT_MARIO, m, HURT_BURNING, NULL);
+    if (m->health < 0x100) {
+        if (m == &gMarioStates[0]) {
+            // never kill remote marios
+            set_mario_action(m, ACT_STANDING_DEATH, 0);
```
Isaac0-dev, line 1695: the previous way of killing mario is still needed, because they were specifically for airborne actions. it would set mario as dead, but the death action doesn't play until mario reaches the floor. setting mario's action while in the air will teleport mario to the ground even if he's not there yet.

Isaac0-dev (changes requested):
some changes, also please fix merge conflicts

src/game/mario_actions_automatic.c
Sunketchupm: The code was rewritten to be like this because all of the returns. although made the code easier to read, made it pretty incompatible with the hook.

Sunketchupm:
I don't think I'll go anywhere with this PR so I'll close it for now

PR #678, opened 2025-02-22
Lvl Macros

SonicTheHedgeHog64:
for dynos levels that use the normal macros

image

### DIFF
--- a/data/dynos_bin_pointer.cpp
+++ b/data/dynos_bin_pointer.cpp
@@ -136,6 +136,12 @@ static PointerData GetDataFromPointer(const void* aPtr, GfxData* aGfxData) {
         return { builtinActor, 0 };
     }
 
+// Built-in Level Macros
+    auto builtinLvlMacro = DynOS_Builtin_LvlMacro_GetFromData((const MacroObject*)aPtr);
+    if (builtinLvlMacro != NULL) {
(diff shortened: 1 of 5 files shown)


data/dynos_bin_pointer.cpp
```diff
@@ -136,6 +136,12 @@ static PointerData GetDataFromPointer(const void* aPtr, GfxData* aGfxData) {
         return { builtinActor, 0 };
     }
 
+// Built-in Level Macros
```
Isaac0-dev: please fix this indentation

data/dynos_mgr_builtin.cpp
```diff
@@ -441,9 +442,58 @@ s32 DynOS_Builtin_Actor_GetCount() {
     return sizeof(sDynosBuiltinActors) / (2 * sizeof(sDynosBuiltinActors[0]));
 }
 
-  ////////////////
- // Level Geos //
```
Isaac0-dev, line 445: it looks like you removed the comment for level geos?

avereyishere: 👍🏾

data/dynos_bin_pointer.cpp
avereyishere: 👍🏾

PR #635, opened 2025-01-18
[WIP] Added new logging system to create and write stuff into .log files

iZePlayzYT:

PR #679, opened 2025-02-24
New logging system with file output

kermeow:
This is a continuation of #635 that has been rebased with latest dev.
I also went through every file that already contained `LOG_*` macros to make sure they wouldn't spam the logs.

- `debuglog.h` has been replaced with `log.h`/`log.c`
- `LOG_*` macros have been made to replicate the previous behaviour
- `LOG_*_VERBOSE` macros have been made to replicate filename logging and debug print behaviour
- `LOG_DEBUG` and `LOG_DEBUG_VERBOSE` are occluded from release builds

Translation will be needed for the "Verbose Logs" misc option that has replaced the old debug log options.

cc: @iZePlayzYT


iZePlayzYT (dismissed):
Looks good!

kermeow:
This seems pretty ready to me now.
If there are any last minute changes needed then now is the perfect time to bring them up!
iZe suggested adding a crash report to the log but I feel like that should be a whole other thing instead of tying everything in with the log file.

A **_LOT_** of files were changed here and, while it builds and runs successfully, I don't trust myself to have not forgotten something. plz review @Isaac0-dev 🙏

iZePlayzYT:
@kermeow i didnt mean that we should create a seperate crash log file, but if the game crashes it would be helpful if some details why it crashed would be saved into the log file. especially helpful when you already closed the game and still want to report the "bug" or "problem" in discord for example.

AgentXLP:
Is anything gonna be done with this

kermeow:
This can probably be closed if there are no plans to review it any time soon.
In that case I'll consider coming up with a better way of doing this later.

AgentXLP:
Did you read what they said about coming up with a better way

iZePlayzYT:
> Did you read what they said about coming up with a better way

Yes but nobody said the current implementation is bad before. I guess @kermeow assumed this because it didnt get reviewed yet.

PR #682, opened 2025-02-25
Added a logging system with file output for "Makefile"

iZePlayzYT:
Similar to #679 i added a logging system for the Makefile Output


Isaac0-dev:
often people won't want a compulsory build log, when you can just do `make > output.log`

PR #689, opened 2025-03-04
Build docker image

cyberplant:


cyberplant:
Heh, sorry for the noise, I was doing local tests and just realized I opened the PR on upstream! 😂 🤦 

I'll come back here but after I have something working.

PR #690, opened 2025-03-07
Add HOOK_ALLOW_FALL_DAMAGE

ManIsCat2:


Blockyyy:
wait so why do we need this?

ManIsCat2:
> wait so why do we need this?

if you wanna use m.peakHeight while removing fall damage you can use this

EmilyEmmi:
I feel like a gLevelValues flag or something similar would be better. Don't really see the point of a hook?

Isaac0-dev:
you can already do this by having another variable to keep track of mario's peak height, I don't think this is necessary.
Is there something else this allows for that you can't already do?

ManIsCat2:
> you can already do this by having another variable to keep track of mario's peak height, I don't think this is necessary. Is there something else this allows for that you can't already do?

nope, that was basically it

PR #697, opened 2025-03-10
Allow Yoshi to do a flutter jump after doing a back flip or a side flip

Brob2nd:
By the way, please, if merging, do it after updating to Extra Characters version 8.1 on this branch


Brob2nd:
ok this can be merged now that extra characters got updated here

Isaac0-dev:
apparently this is not supposed to be part of the moveset by design, please discuss this with the other folks before re-opening if you think it should be in

PR #704, opened 2025-03-14
Properly fix lua interaction flags

Sunketchupm:
Reverts a change Isaac did a little while ago and fixes the source of the issue


AgentXLP:
Yeah idk what that was supposed to achieve

PR #709, opened 2025-03-16
Add hook_interval

kermeow:
`hook_interval()` allows Lua to run a callback at a rate other than 30hz. The period in seconds is given as `interval`.
- When `cumulative` is true, the callback is only run once to make up for skipped intervals, and the delta time is summed
- Delta time is always the interval when `cumulative` is false

### Parameters

| Field | Type |
| ----- | ---- |
| interval | `number` |
| cumulative | `boolean` |
| callback | `Lua Function` (`number` delta) |

### Lua Example

```lua
--- @param delta number
local function runs_at_100hz(delta)
    print("100hz! Delta:", delta) -- "100hz! Delta: 0.01"
end

hook_interval(1/100, false, runs_at_100hz)
```


kermeow:
side note: this should help with people having issues with audio looping

Isaac0-dev:
sorry, but this would be ok if access to smlua was blocked during these "intervals"

kermeow:
> if access to smlua was blocked during these "intervals"

err what does that mean
like block access to the apis? but why?

maybe would have been nice to discuss this before it was closed :(

AgentXLP:
The Mystery

src/pc/lua/smlua_hooks.c
```diff
@@ -2483,6 +2483,123 @@ void smlua_call_mod_menu_element_hook(struct LuaHookedModMenuElement* hooked, in
...
+void smlua_update_interval_hooks() {
+    lua_State* L = gLuaState;
+    if (L == NULL) { return; }
+    if (sHookedIntervalsCount == 0) { return; }
+
+    static f64 sLastIntervalUpdate;
+    f64 now = clock_elapsed_f64();
+    f64 deltaTime = now - sLastIntervalUpdate;
+    if (sLastIntervalUpdate == 0) { deltaTime = 0; }
+    sLastIntervalUpdate = now;
```
PeachyPeachSM64, line 2556: `sLastIntervalUpdate` shouldn't be a function static and should be updated outside of it
the ugly init hack works only on game boot, if you disconnect and host or join again with a mod having a hook_interval, you'll get a massive delay due to sLastIntervalUpdate not being updated

src/pc/lua/smlua_hooks.c
```diff
@@ -2483,6 +2483,123 @@ void smlua_call_mod_menu_element_hook(struct LuaHookedModMenuElement* hooked, in
...
+        lua_rawgeti(L, LUA_REGISTRYINDEX, hook->reference);
+
+        // push parameters
+        lua_pushnumber(L, hookDeltaTime);
+
+        // call the callback
+        if (0 != smlua_call_hook(L, 1, 0, 0, hook->mod)) {
+            LOG_LUA("Failed to call the interval callback: %s", hook->mod->name);
+            continue;
+        }
```
PeachyPeachSM64, line 2599: duplicated code!
the logic is simple, why is the code arranged in such a convoluted way?
if not? do while that leads to a continue? and the  "else" branch is just the do while body but with one line different?

src/pc/pc_main.c
```diff
@@ -226,6 +226,8 @@ void produce_interpolation_frames_and_delay(void) {
 
         sDrawnFrames++;
 
+        CTX_EXTENT(CTX_SMLUA, smlua_update_interval_hooks);
```
PeachyPeachSM64, line 229: I don't like the fact that it is framerate dependent. Such a feature would be better in a separate thread, but I'm not sure if it's possible at all.
This results in the `delta` parameter in the hooked function straight up lying to the user:
- if `cumulative` is true, `delta` gets the value of `hookDeltaTime`, which is NOT the delta between two successive calls of the hook (`interval`), but a value between `interval` and `interval + the current frame delta time`
- if `cumulative` is false, `delta` is always set to `interval`, which is also wrong (first `delta` is actually the elapsed time between the last frame and the current call, and next `delta`s are near-zero (or the time between two successive lua callbacks))

PeachyPeachSM64 (changes requested):

src/pc/pc_main.c
kermeow: To be completely honest it was that long ago I wrote this I cannot for the life of me remember why I did it that way

Isaac0-dev: do not put it in a separate thread, that'd be way worse because you could never be sure that stuff is going to be safe, Lua could just write anywhere while its being modified, and if you tried to mutex stuff it'd defeat the point of the feature in the first place
as i see it, doing it inside the interp loop is the closest you can get to "valid". but i agree, its not nice that it's done inside the interpolation loop

kermeow:
Putting this to draft while I decide what to do with it.

PR #711, opened 2025-03-17
Miscellaneous additions

Cooliokid956:
- Fixed script file path shortening for Windows in console errors (what I could at least, the one left unfixed is preprocessed and I can't mess with that)
- Adjusted audio sanity check messages to sound more natural
- 9 new functions added:

## `audio_stream_set_loop_points(audio, start, end)`
Sets an audio stream's loop points in samples, no longer requiring a hacky method to loop streams in mods

## `get_active_mod()`
Returns the mod currently being processed, can be used by mods with APIs to determine what they're being used by and react accordingly

## `get_area_update_counter()`
Returns `gAreaUpdateCounter`, can be used to determine various in-game statistics, as well as if the game is being updated

## Exposed several object pointers
`get_current_object()` to retrieve `gCurrentObject`
`get_dialog_object()` to retrieve `gContinueDialogFunctionObject`
`(get|set)_cutscene_focus()` to retrieve/assign `gCutsceneFocus`
`(get|set)_secondary_camera_focus()` to retrieve/assign `gSecondCameraFocus`


Sunketchupm:
`gObjectCutscebeDone` being exposed when

Cooliokid956:
One extra change I'd thought about a while ago but just remembered:

Allow ALL characters beyond the Unit Separator character for use with mod storage (with an exception for `[`, `]`, and `=` in keys)

Cooliokid956:
> `gObjectCutsceneDone` being exposed when

You might be hallucinating
image

Sunketchupm:
Might be a similar variable name, since I remember that it exists within `sparkle_star_spawn.inc.c` (or some similar file name)

src/pc/mods/mod_storage.cpp
```diff
@@ -43,11 +43,11 @@ static void strdelete(char* string, const char* substr) {
     string[i] = '\0';
 }
 
-bool char_valid(const char* buffer) {
+bool char_valid(const char* buffer, bool isKey) {
     if (buffer[0] == '\0') { return false; }
 
     while (*buffer != '\0') {
-        if ((*buffer >= 'a' && *buffer <= 'z') || (*buffer >= 'A' && *buffer <= 'Z') || (*buffer >= '0' && *buffer <= '9') || *buffer == '_' || *buffer == '.' || *buffer == '-') {
+        if (*buffer > '' && !(isKey && (*buffer == '[' || *buffer == ']' || *buffer == '='))) {
```
PeachyPeachSM64: how is this compiling? GCC doesn't even allow it
even if it does, that sounds like undefined behavior
image

src/pc/lua/utils/smlua_misc_utils.h
```diff
@@ -41,6 +41,8 @@ struct DateTime {
 
 /* |description|Gets the current area's networked timer|descriptionEnd| */
 u32 get_network_area_timer(void);
+/* |description|Gets the area update counter incremented when objects are updated|descriptionEnd| */
+u16 get_area_update_counter(void);
```
PeachyPeachSM64, line 45: what's the use for this? why this one and not gGlobalTimer?

src/pc/mods/mod_storage.cpp
```diff
@@ -148,7 +148,7 @@ C_FIELD bool mod_storage_load_bool(const char* key) {
 C_FIELD bool mod_storage_exists(const char* key) {
     if (gLuaActiveMod == NULL) { return false; }
     if (strlen(key) > MAX_KEY_VALUE_LENGTH) { return false; }
-    if (!char_valid((char *)key)) { return false; }
+    if (!char_valid((char *)key, true)) { return false; }
```
PeachyPeachSM64: pointless cast

src/pc/mods/mod_storage.cpp
```diff
@@ -164,7 +164,7 @@ C_FIELD bool mod_storage_exists(const char* key) {
 C_FIELD bool mod_storage_remove(const char* key) {
     if (gLuaActiveMod == NULL) { return false; }
     if (strlen(key) > MAX_KEY_VALUE_LENGTH) { return false; }
-    if (!char_valid((char *)key)) { return false; }
+    if (!char_valid((char *)key, true)) { return false; }
```
PeachyPeachSM64: same here

PeachyPeachSM64 (changes requested):

Cooliokid956:
I was asked to expose one more object:
`geo_get_current_object()` to retrieve `gCurGraphNodeObject` cast to a `struct Object*` which somehow works and is used throughout the game, useful in geo asm functions

src/pc/mods/mod_storage.cpp
Cooliokid956: there should be a `` there, but it seems that wasn't pushed through? i'll replace it with a space check instead

src/pc/lua/utils/smlua_misc_utils.h
Cooliokid956: `gAreaUpdateCounter` only updates in certain contexts while `gGlobalTimer` is always ticking

src/pc/mods/mod_storage.cpp
Cooliokid956: hey, I'm not the one to `Blame`!

src/pc/mods/mod_storage.cpp
Cooliokid956: :p

src/pc/mods/mod_storage.cpp
PeachyPeachSM64: > there should be a � there

no, never use special characters in code, what the hell
thanks git for preventing some war crimes

src/pc/lua/utils/smlua_misc_utils.h
PeachyPeachSM64: what are those "certain contexts"? how is it useful in lua mods? just curious

src/pc/mods/mod_storage.cpp
AgentXLP: The horrors

src/pc/lua/utils/smlua_misc_utils.h
AgentXLP: "To determine Mario's true blinking state"

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -349,88 +346,82 @@ void audio_stream_play(struct ModAudio* audio, bool restart, f32 volume) {
...
 void audio_stream_set_looping(struct ModAudio* audio, bool looping) {
-    if (!audio_sanity_check(audio, true, "set stream looping")) {
-        return;
-    }
+    if (!audio_sanity_check(audio, true, "set stream looping for")) return;
+    
     ma_sound_set_looping(&audio->sound, looping);
 }
 
+void audio_stream_set_loop_points(struct ModAudio* audio, u64 start, u64 end) {
```
Isaac0-dev: end is a Lua reserved word

also, Lua's integers in 5.3 are s64, not u64. if a modder provides a negative value, it would be treated as a very high positive value (i can see that you're just copying the data type used by `ma_data_source_set_loop_point_in_pcm_frames`, but there's no way a modder will need 64 bits to represent the number they're passing in anyway)

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -324,18 +324,15 @@ struct ModAudio* audio_stream_load(const char* filename) {
 }
```
Isaac0-dev, line 324: please don't remove the braces around the return statements in this file to keep consistent with the code style. 
its fine to keep it as a single line though

Isaac0-dev (changes requested):
looks mostly fine

src/pc/lua/utils/smlua_audio_utils.c
AgentXLP: Suggestion: loopStart and loopEnd

Cooliokid956: I did see an error in the autogen file on vsc, but shrugged it off since it loaded just fine ingame (probably not the actual script though)

will apply these changes

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: fixed

Cooliokid956:
That was something
here's the changelog from the commit description, plus changes not listed

- switched to loopStart/End wording
- switched param types to `s64`
- reverted return curly braces

Additionally, I:
- found two missing spaces after a for and if
- fixed `audio_stream_get_position`, now getting the true position no matter the loop point setup
- adjusted `audio_stream_set_loop_points`'s behavior to accept negative numbers, these will wrap around to the end of the stream (you may set `loopEnd` to `0` to set the loop end to the very end)
- provide a hint for rehabilitating the tempo functions (?)
- added missing `gCurGraphNodeObject` extern

Cooliokid956:
Today I worked on a few more changes for autogen

1. Make array types generate as actual arrays instead of using `Array_`:
image

2. Overhauled enum constant autogen to allow for enum autocomplete as such:
image

AgentXLP:
It's peak...

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -442,9 +438,8 @@ void audio_stream_set_volume(struct ModAudio* audio, f32 volume) {
...
 
 // void audio_stream_set_speed(struct ModAudio* audio, f32 initial_freq, f32 speed, bool pitch) {
-//     if (!audio_sanity_check(audio, true, "set stream speed")) {
+//     if (!audio_sanity_check(audio, true, "set stream speed for"))
 //         return;
-//     }
```
Isaac0-dev, line 447: braces are still removed here

Isaac0-dev (commented):
`Object[]` instead of `Array_Object`, but not `*Object` instead of `Pointer_Object`?
it doesn't look consistent, and if you went with both it wouldn't look great with the `*Object`.
imo just leave it as `Array_Object`.

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: ok ok

PR #712, opened 2025-03-18
Improvements to non-pvp player interactions

Sunketchupm:


Isaac0-dev:
please describe what this pr is supposed to do

Sunketchupm:
Not really sure what I can describe other than what each commit says.
The long jump and object jump bounces exist to make those bounces more fun and interesting to do.
The out of bounds player pushing being prevented was made pretty much because of star road mariohunt, but it really shouldn't have existed in the first place. Objects normally can't push Mario out of bounds.
Both of these are player interactions that, while can be used in pvp, aren't inherently pvp, which is why my pr name is what it is.

Isaac0-dev:
thanks i just wanted that to be clarified, it can be merged

PR #716, opened 2025-03-23
Basic Vertex Point Lighting Engine

AgentXLP:
This is a basic vertex point lighting engine. It works by having a geometry mode that's checked in the renderer and calculating the lighting for each lighting engine vertex by iterating through surrounding lights that are in range.

The main point of this lighting engine is for dynamic moving light sources. If you wish to use static vertex lighting in your level, you should consider baking lighting into the vertex colors of the mesh instead of this. I might make a tutorial or upload a script for this at some point.

To use this in your levels, you need to use [this Fast64 fork](https://github.com/coop-deluxe/fast64) and use a vertex colored texture preset, turn off `Show Simplified UI`, go to `Geo`, and enable `Lighting Engine`.

In your level, create a new object with the behavior `LE Ambient Light` (`bhvAmbientLight`.) The first 3 bytes of the behavior parameter are the R, G, and B values of the color.
To create a point light in your level, create a new object with the behavior `LE Point Light` (`bhvPointLight`.) The first 3 bytes of the behavior parameter are the R, G, and B values of color. The last byte is the radius of the light in SM64 units multiplied by 10 (e.g. 0x64 = 1000 units)

To attach lights to objects, you can use `hook_behavior` to spawn a light object on init.

```lua
--- @param o Object
--- I could probably make this a builtin function?
local function attach_point_light_to_obj(obj)
    spawn_non_sync_object(
        id_bhvPointLight,
        E_MODEL_NONE,
        obj.oPosX, obj.oPosY, obj.oPosZ,
        --- @param o Object
        function(o)
            o.parentObj = obj
        end
    )
end

id_bhvGoomba = hook_behavior(id_bhvGoomba, OBJ_LIST_PUSHABLE, false, attach_point_light_to_obj, nil, "bhvGoomba")
```
This script should work.


SonicTheHedgeHog64:
👏🏾

SonicTheHedgeHog64:
Would be nice too have this be per pixel, but for right now we'll just have to deal with it

SonicTheHedgeHog64:
image
fast64-coop-dx needs a rebase

docs/lua/examples/lighting-engine-demo/levels/le-ify.py
Isaac0-dev, line 1: a python file inside a coop mod? this should be somewhere else if its actually meant to be here

src/engine/lighting_engine.c
```diff
@@ -0,0 +1,160 @@
...
+
+        lightingDir[0] += dir[0] * dist * light->intensity;
+        lightingDir[1] += dir[1] * dist * light->intensity;
+        lightingDir[2] += dir[2] * dist * light->intensity;
+
+        count++;
+    }
+
+    count++; // additional increment
+    out[0] = lightingDir[0] / (f32)(count);
```
Isaac0-dev, line 84: what if there are no lights, but `sLights` is initialized? `count` and `lightingDir` should be defined

src/engine/lighting_engine.c
```diff
@@ -0,0 +1,160 @@
...
+    vec3f_set(light->pos, x, y, z);
+    color_set(light->color, r, g, b);
+    light->radius = radius;
+    light->intensity = intensity;
+    hmap_put(sLights, ++sLightID, light);
+    return sLightID;
+}
+
+void le_remove_light(s16 id) {
+    hmap_del(sLights, id);
```
Isaac0-dev, line 106: this memory needs to be freed, not just unlisted from the hmap

src/engine/lighting_engine.c
```diff
@@ -0,0 +1,160 @@
...
+    light->intensity = intensity;
+}
+
+void le_clear(void) {
+    if (sLights == NULL) { return; }
+
+    for (struct LELight* light = hmap_begin(sLights); light != NULL; light = hmap_next(sLights)) {
+        free(light);
+    }
+    hmap_clear(sLights);
```
Isaac0-dev, line 148: `sLights` should be set to NULL after it is cleared, otherwise it won't be recreated later

Isaac0-dev (commented):
all the files with only `extern`s or `#include`s are ignored by dynos

SonicTheHedgeHog64:
image

SonicTheHedgeHog64:
may be a level issue

docs/lua/examples/lighting-engine-demo/levels/le-ify.py
AgentXLP: Whoops, thats a leftover

AgentXLP:
> Would be nice too have this be per pixel, but for right now we'll just have to deal with it

That would need to be implemented into the shader code and if I did that I would just drop DirectX support tbh so I don't have to program it once and then do it all over again in some weird way I don't understand.

SonicTheHedgeHog64:
May just do it myself for no reason other then the funny

src/engine/lighting_engine.c
AgentXLP: `sLights` is set to `NULL` in `le_shutdown`. `le_clear` is called when transitioning to another area, all the lights are cleared but the hashmap is not destroyed or anything so we don't have to recreate it.

src/engine/lighting_engine.c
AgentXLP: Good catch

src/engine/lighting_engine.c
AgentXLP: Alright

AgentXLP:
> all the files with only `extern`s or `#include`s are ignored by dynos

I know, Fast64 exports those headers as boilerplate and they aren't removed in the custom level example either.

src/engine/lighting_engine.c
Isaac0-dev: :+1:  thats good, didn't see that

Isaac0-dev:
just a suggestion: if you want to give people a shot at trying this out for themselves, you could push a commit with [build] in the name, and change the branch name in the workflow to make it build for you

SonicTheHedgeHog64:
this should be f3dex3-ified, aka using the code from f3dex3 to port over packed normals to have LE work with shaded object's

SonicTheHedgeHog64:
> > Would be nice too have this be per pixel, but for right now we'll just have to deal with it
> 
> That would need to be implemented into the shader code and if I did that I would just drop DirectX support tbh so I don't have to program it once and then do it all over again in some weird way I don't understand.

you could ditch directx for vulkan --> https://github.com/theclashingfritz/sm64coopdx/tree/vulkan-renderer

AgentXLP:
@Isaac0-dev is this good to merge now

PR #718, opened 2025-03-23
Gfx set command: v2 + Gfx/Vtx dynamic alloc

PeachyPeachSM64:
`gfx_set_command` sets a display list command by passing in a string, formatted like commands in `model.inc.c` files, for example:
`gfx_set_command(gfx, "gsDPSetCombineLERP(0, 0, 0, SHADE, 0, 0, 0, ENVIRONMENT, 0, 0, 0, SHADE, 0, 0, 0, ENVIRONMENT)"`.

- No need to overload the global table with hundreds of constants
- Commands are only parsed once, then cached, so no more processing time after one iteration
- More user-friendly, as you can input commands the same way as in a `model.inc.c` file
- One function for all use cases
- Accepts parameter specifiers to pass integers, strings and pointers to the command (see the example in `docs/lua/examples/gfx-vtx-demo`)
- A usable Vtx type in Lua

Replaces the old `gfx_set_command` and its variations.

Also comes with a brand new feature: **Dynamic allocation of Gfx and Vtx buffers**!

Adds the following functions to easily manipulate display lists and vertex buffers:
- `gfx_get_display_list`: gets the display list pointer of a `gsSPDisplayList` command
- `gfx_get_vertex_buffer`: gets the vertex buffer of a `gsSPVertex` command
- `gfx_get_vertex_count`: gets the vertex count of a `gsSPVertex` command
- `gfx_get_from_name`: gets a display list and its length from its name (includes models, levels, vanilla and dynamically allocated)
- `gfx_get_length`: gets the max length of a display list
- `gfx_get_command`: gets the command of a display list at a given position
- `gfx_get_next_command`: gets the next command of a given display list pointer, intended to use in for loops
- `gfx_copy`: copy commands from a display list to another
- `gfx_new`: create a new named display list
- `gfx_realloc`: realloc a named display list allocated by `gfx_new`
- `gfx_delete`: delete a named display list allocated by `gfx_new`
- `vtx_get_from_name`: gets a vertex buffer and its vertex count from its name (includes models, levels and dynamically allocated)
- `vtx_get_count`: gets the max number of vertices of a vertex buffer
- `vtx_get_vertex`: gets the vertex of a vertex buffer at a given position
- `vtx_get_next_vertex`: gets the next vertex of a given vertex pointer, intended to use in for loops
- `vtx_copy`: copt vertices from a vertex buffer to another
- `vtx_new`: create a new named vertex buffer
- `vtx_realloc`: realloc a named vertex buffer allocated by `gfx_new`
- `vtx_delete`: delete a named vertex buffer allocated by `gfx_new`

**Demo**:
All shapes are generated by Lua with Gfx and Vtx manipulations.
Source files available in `docs/lua/examples/gfx-vtx-demo`.

https://github.com/user-attachments/assets/d9275089-c9db-4fc5-b344-eb5b4fa6b7f9


AgentXLP:
It's peak

Isaac0-dev (commented):
I assume this isn't meant to take variables as parameters? that was the reason I didn't go with this approach initially

SonicTheHedgeHog64:
Nistalgia Stamp Of approval

SonicTheHedgeHog64:
based commit imo

kermeow:
now this better suits my use case 🤤

AgentXLP:
Merge conflicts

data/dynos_mgr_moddata.hpp
```diff
@@ -0,0 +1,104 @@
...
+        return false;
+    }
+
+    aDeallocator(aPointer);
+    aModData[aModIndex].erase(itemName);
+    return true;
+}
+
+template <typename T, typename Deallocator>
+void DynOS_ModData_DeleteAll(ModData<T> &aModData, Deallocator aDeallocator) {
```
Isaac0-dev, line 97: For `DynOS_ModData_Realloc`, `DynOS_ModData_Delete` and `DynOS_ModData_DeleteAll`, why even have a deallocator parameter?
every use of it is just the default free function

src/pc/gfx/gfx_pc.c
```diff
@@ -770,6 +770,8 @@ static float gfx_adjust_x_for_aspect_ratio(float x) {
 }
 
 static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, const Vtx *vertices, bool luaVertexColor) {
+    if (!vertices) { return; }
```
Isaac0-dev, line 773: can mods actually create a reference to a null vertices? shouldn't that be disallowed?
It'd be a shame if we had some bug that we can't figure out because it's just being ignored here

data/dynos_bin_gfx.cpp
```diff
@@ -1144,4 +1187,248 @@ void DynOS_Gfx_Load(BinFile *aFile, GfxData *aGfxData) {
...
+    Gfx *gfxHead = gfxBuffer;
+    ParseGfxSymbol(aGfxData, &aNode, gfxHead, aTokenIndex);
+    if (aGfxData->mErrorCount > 0) {
+        return false;
+    }
+
+    // Cache parsed command
+    u32 commandLength = (u32) (gfxHead - gfxBuffer);
+    Gfx *cached = (Gfx *) calloc(commandLength, sizeof(Gfx));
+    memcpy(cached, gfxBuffer, commandLength * sizeof(Gfx));
```
Isaac0-dev: can use malloc here since you're immediately overwriting the memory
(calloc resets the memory back to 0, so its slower)

data/dynos_bin_gfx.cpp
```diff
@@ -1144,4 +1187,248 @@ void DynOS_Gfx_Load(BinFile *aFile, GfxData *aGfxData) {
...
+        return "";
+    }
+    return returnFunc(value);
+}
+
+static String ResolveParam(lua_State *L, GfxData *aGfxData, u32 paramIndex, char paramType) {
+    switch (paramType) {
+
+        // Integer
+        case 'i': return ConvertParam<s64>(L, aGfxData, paramIndex,
```
Isaac0-dev: here's a personal code style preferance:
add a return before the first parameter, to make it obvious that it's a function call with parameters on multiple lines. I find that it does help

like this:

```C
ConvertParam<s64>(
    L, aGfxData, paramIndex,
    "integer",
    [] (lua_State *L, u32 paramIndex) { return smlua_to_integer(L, paramIndex); },
    [] (s64 integer) { return String("%lld", integer); }
);
```
can also apply to the other instances of this kind of code below

src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -147,25 +190,41 @@ void gfx_parse(Gfx* cmd, LuaFunction func) {
                 if (lua_type(L, -1) == LUA_TBOOLEAN && smlua_to_boolean(L, -1)) {
                     return;
                 }
+                switch (op) {
```
Isaac0-dev: assuming this is moved so that mods can actually see those larger than normal commands:
previously, this prevented mods from overwriting a command that's larger than normal, because it would skip those commands.
say you have a command that's 128 bits in size, and you insert a command that's 64 bits. the second half of the original command is not a valid command, so it could cause all sorts of trouble.
My imagination here may be off, but at least you can think about it and decide if its a problem

Isaac0-dev (commented):
before i say anything else, this is awesome! thanks for expanding on this feature, I can see you've been very thoughtful about this!
I've decided to add comments that don't matter too, so feel free to ignore them if you don't care for them.
I'll probably be revisiting this again before its merged, even if no changes are made.

data/dynos_mgr_moddata.hpp
PeachyPeachSM64: for reusability
for example if a data type is allocated with `new`, you need to deallocate it with `delete`

src/pc/gfx/gfx_pc.c
PeachyPeachSM64: In th example mod, there is a `gsSPVertex(NULL, 0, 0)` command
If the lua code errors before setting the vertex pointer to something non-NULL, the display list will still try to execute gfx_sp_vertex with a NULL pointer
It will avoid crashing the game on errors like this, but the downside is, as you pointed out, we can miss possible bugs that would set vertex buffers to NULL for some reason

PeachyPeachSM64: same for `gsSPDisplayList(NULL)`

src/pc/lua/utils/smlua_gfx_utils.c
PeachyPeachSM64: a small rework of gfx_parse
after the Lua callback, the op of the command is retrieved again in case it changed during the callback, and the size of both ops are compared
if the new op size is smaller, NoOp commands are inserted so the data leftovers from the old command are not interpreted as commands

src/pc/gfx/gfx_pc.c
AgentXLP: Yeah

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: that's the best solution, good stuff!

src/pc/gfx/gfx_pc.c
Isaac0-dev: ah so its to allow placeholder command then that can be filled in by mods.
I guess we'll just have to deal with it then

data/dynos_mgr_moddata.hpp
Isaac0-dev: sounds like this could be made into it's own class

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: actually, while that solution is a good start, it needs to be applied probably inside `gfx_set_command`. because otherwise you can still make invalid commands by using `gfx_get_next_command`, or perhaps the display list may start with a texture command.
@PeachyPeachSM64

data/dynos_mgr_moddata.hpp
PeachyPeachSM64: what do you mean?

Isaac0-dev: I was just thinking that the mod data code could be made into a class that starts up the map for you and stuff.. but it probably doesn't matter

Isaac0-dev (approved):

PR #720, opened 2025-03-25
duplicate every instance of a display list

Isaac0-dev:
This is to fix a crash caused by writing to read-only memory, if a mod tries to modify a display list from vanilla sm64.
it also makes sure that modifications to display lists are unique to the display list it was applied to
This is attempt 2, now as a pull request.
It now recursively duplicates display lists and vertices referenced by other display lists.
Its now done inside dynos to make sure things are only freed when they need to be.


SonicTheHedgeHog64:
HELL YEAHHHHHHHHHHHHHHHHHHHHHHHHH

PR #722, opened 2025-03-28
Update revamped flying damage

Sunketchupm:
I figured that this would make more sense. 2 damage with 60 speed, +1 for every extra 20 speed.


Isaac0-dev:
sure

PR #723, opened 2025-03-29
Miscellaneous Additions: Addendum

Cooliokid956:
## Master Volume rework
Master Volume has been simplified into a simple `gMasterVolume` that only updates when necessary (on startup/on volume change)
This variable replaces what were previously instances of the same master volume calculation ((f32)v1/127 * (f32)v2/127), which should improve performance by a miniscule amount?
Master Volume has also been made to affect the samples directly instead of the sequence players' volumes. This helps the sound quality by avoiding cut notes at low volumes

also skipped some processing when `shouldMute`, now audio simply isn't played

## Autogen
`Pointer_`s have been converted to aliases of their real counterparts
"Total constants" now reads the actual amount of constants (save for the manual constants)
Disabled sorting for constants and functions so that they match their source files more closely

### Mouse
Mouse buttons are now accessible to the API through the following functions:
- `djui_hud_get_mouse_down()`
- `djui_hud_get_mouse_click()`
- `djui_hud_get_mouse_released()`

These can be checked against using the following flags:
- MOUSE_1 (L_MOUSE)
- MOUSE_2 (M_MOUSE)
- MOUSE_3 (R_MOUSE)
- MOUSE_4
- MOUSE_5

### Sequence Player
The following functions will help provide a more faithful Streamed Music experience:
- `sequence_player_get_volume(player)`
- `sequence_player_get_fade_volume(player)`
- `sequence_player_get_mute_volume_scale(player)`


PeachyPeachSM64:
> Disabled sorting for constants and functions so that they match their source files more closely

I don't see the point
The documentation is intended for end users (Lua modders), not coopdx developers
Alphanumeric sorting is better than no sorting at all, the end user doesn't care about how the functions are organized in the source code

src/game/game_init.c
```diff
@@ -508,6 +509,15 @@ void read_controller_inputs(void) {
...
     gPlayer3Controller->buttonDown = gPlayer1Controller->buttonDown;*/
+
+    // Mouse Input
+    controller_mouse_read_window();
+    extern u8 gMouseDown;
+    extern u8 gMouseClick;
+    extern u8 gMouseReleased;
+    gMouseClick = ~gMouseDown & mouse_window_buttons;
+    gMouseReleased = ~mouse_window_buttons & gMouseDown;
+    gMouseDown = mouse_window_buttons;
```
PeachyPeachSM64:
```c
u32 prevMouseWindowButtonsDown = mouse_window_buttons;
controller_mouse_read_window();
mouse_window_buttons_pressed = ~prevMouseWindowButtonsDown & mouse_window_buttons;
mouse_window_buttons_released = ~mouse_window_buttons & prevMouseWindowButtonsDown;
```

More general remarks about the mouse stuff:
- mouse buttons are "pressed", not "clicked": for consistency, rename the 3 functions: `djui_hud_get_mouse_buttons_down`, `djui_hud_get_mouse_buttons_pressed` and `djui_hud_get_mouse_buttons_released`
- Also for consistency, instead of the gVariables, declare `mouse_window_buttons_pressed` and `mouse_window_buttons_released` as `u32` in `controller_mouse.h` and define them in `controller_mouse.c` under the already existing `mouse_window_buttons`, which is the "down" state
- rename the `MOUSE` constants `MOUSE_BUTTONS`

src/pc/pc_main.c
```diff
@@ -255,19 +255,28 @@ void produce_interpolation_frames_and_delay(void) {
 // It also may help static analysis and bug catching.
 static s16 sAudioBuffer[SAMPLES_HIGH * 2 * 2] = { 0 };
 
+f32 gMasterVolume;
```
PeachyPeachSM64: Inappropriate place to define a global variable

When you create a new global var, please declare it in the appropriate header file and define it in the corresponding source file, and avoid using `extern` everywhere you need it

src/game/game_init.c
Cooliokid956:
```c
ok
```

- ok
- ok
- not a fan of the `S` so i'll just drop that but otherwise ok

src/pc/pc_main.c
Cooliokid956: Thanks for the advice

Cooliokid956:
# Forced 4:3 Mode is back!!

https://github.com/user-attachments/assets/fc69dcc1-c125-47c4-9b86-24c834bc26a3

Cooliokid956:
# 💥NEW: Scrolling
- Scrolling added to chat box (hold ctrl to scroll fewer lines, shift to scroll faster)
- Scrolling functions added to smlua


https://github.com/user-attachments/assets/0a2ddfd3-56f2-484b-ad29-0ce5bf7e5548

src/game/game_init.c
PeachyPeachSM64: > not a fan of the S so i'll just drop that but otherwise ok

yeah the extra S at the end was not intended, I indeed meant `MOUSE_BUTTON_x`

@Cooliokid956, you forgot to change the return type of the `djui_hud_get_mouse_buttons` functions to `u32`

everything else is good

src/pc/djui/djui_inputbox.c
```diff
@@ -13,9 +13,9 @@
 #define DJUI_INPUTBOX_MID_BLINK (DJUI_INPUTBOX_MAX_BLINK / 2)
 #define DJUI_INPUTBOX_CURSOR_WIDTH (2.0f / 32.0f)
 
-static u8 sHeldShift   = 0;
-static u8 sHeldControl = 0;
-static u8 sHeldAlt     = 0;
+u8 heldShift   = 0;
```
Isaac0-dev: this naming isn't great, they're globals so at least `gHeld_`, although `gDjuiInputHeld_` is more appropriate, but its just a long name.
also declare them in a header file i think

Isaac0-dev (commented):
I think something like `djui_get_window_dimensions` would be good, since it seems like you're adding translation for 4:3 to every call of `get_dimensions`.
also I'm not a fan of removing the sorted constants either.

Cooliokid956:
> also I'm not a fan of removing the sorted constants either. 

I did this because there were many cases where sorting would jumble up related constants, leading to

and

there were existing examples of sections already outta wack that I also noticed
the solution may be to convert these to enums, but you also have to consider the intentional placement by the author; these constants are already sorted.

AgentXLP:
I don't know if this is going to be ready by tomorrow or April 1st. Initially I figured it would be merge-able but a lot has been added that feels too experimental to put in with so little time left in the update's development cycle.

src/pc/djui/djui_inputbox.c
Cooliokid956: image

Baconator2558:
> I don't know if this is going to be ready by tomorrow or April 1st. Initially I figured it would be merge-able but a lot has been added that feels too experimental to put in with so little time left in the update's development cycle.

Personally, I'd say to hold off until after 1.3 releases. There's a lot of changes here that people are unsure on, and rushing the decision might not work out well.

src/pc/djui/djui_inputbox.c
Cooliokid956: i have settled for the `gDjuiInputHeld` prefix
ok

Cooliokid956:
That's as much as i'll be adding in this PR, feel free to review

Isaac0-dev:
there's now merge conflicts. also is this still wip because the title has not been changed

Cooliokid956:
Rebased, reran autogen, should be good now

Isaac0-dev (approved):
i didn't realise `gfx_get_dimensions` existed

it looks fine to me :+1:

Cooliokid956:
> i didn't realise `gfx_get_dimensions` existed
> 
> it looks fine to me 👍

should i just replace all `djui_gfx_get_dimensions` with it?

Isaac0-dev:
honestly i don't mind, it would be better but it doesn't matter too much to me

Cooliokid956:
Done and done

AgentXLP:
I think I'm going to merge this now, but for future reference please split features like these into their own separate PRs.

PR #725, opened 2025-03-31
Updated coopnet libraries for macOS

EmeraldLoc:
Uses https://github.com/Isaac0-dev/coopnet/pull/1, does connect fine using 2 clients (one arm, the other intel) with the host using a mod.


AgentXLP:
Rip my perfect 120 commits count for the update :pensive:
IT'S FOR THE BETTER OF COOP, PULL IT TOGETHER AGENT

PR #727, opened 2025-04-02
Further fix camera

Sunketchupm:
This should now not conflict with mods that set the course specific settings on their own. My own small test seems to confirm this, but this probably needs to be tested a little further.


Isaac0-dev (changes requested):
`camera_set_use_course_specific_settings` is already called in `network_shutdown`, better remove that

PR #728, opened 2025-04-02
Fix water pvp

Sunketchupm:
Often the simplest solutions are the best ones. I tried to calculate the direction in 3D space, but instead it's just better to check both pitch and yaw.


Sunketchupm:
I just realized there is a pretty big bug here which is why I tried to calculate the direction in the first place. I'll attempt to fix it again and will undraft if I do. Otherwise, will just leave this as the fix (which will be good enough for most cases).

AgentXLP:
@Sunketchupm test that asap or its not getting in the update

Sunketchupm:
Finally tried again and it seems to be working with no blind spots

AgentXLP:
Awesome!

PR #729, opened 2025-04-04
reset permanent model duplicates on session reset

Isaac0-dev:
fixes this
image

it feels like I implemented this in a hacky way so please do let me know if you can see a cleaner way.

The issue is that a mod can edit the display list of a permanent model, and it never gets reset when the session ends.
To fix this, we store a pointer to all the duplicates of a permanent model, and just copy the memory from the original display list, and swap the pointers to child display lists or vertices.


PeachyPeachSM64:
tbh I don't think we should allow mods to modify vanilla display lists
most mods come with their custom models anyway

Isaac0-dev:
I'm not against it, but why do you say that? Is it because of the way I've implemented this?

PR #733, opened 2025-04-06
Gfx Vtx vanilla only duplicates

PeachyPeachSM64:
Some changes for gfx vtx duplication:
- Only vanilla (read-only) stuff is duplicated
- Stuff is duplicated only once, and uses a map rom->ram for next iterations
- Stuff is restored to original values on network shutdown

also fixes incorrect types in some extern declarations


src/pc/network/network.c
```diff
@@ -666,6 +666,7 @@ void network_shutdown(bool sendLeaving, bool exiting, bool popup, bool reconnect
     }
 
     dynos_model_clear_pool(MODEL_POOL_SESSION);
+    dynos_model_restore_vanilla_display_lists();
```
Isaac0-dev: don't need to reset this before shutting down the game

data/dynos_mgr_models.cpp
```diff
@@ -241,54 +229,91 @@ void DynOS_Model_OverwriteSlot(u32 srcSlot, u32 dstSlot) {
...
+    // Check if it's vanilla
+    if (!shouldDuplicate) {
+        shouldDuplicate = (DynOS_Builtin_Gfx_GetFromData(aGfx) != NULL);
+    }
+
+    // Duplicate display list
+    Gfx *gfxDuplicate = aGfx;
+    u32 gfxSize = gfx_get_size(aGfx);
+    if (shouldDuplicate) {
+        gfxDuplicate = (Gfx *) malloc(gfxSize * sizeof(Gfx));
```
Isaac0-dev: the compiler would cover this, but `gfxSize * sizeof(Gfx)` is calculated 3 times
imo that should be in a variable

data/dynos_mgr_models.cpp
```diff
@@ -241,54 +229,91 @@ void DynOS_Model_OverwriteSlot(u32 srcSlot, u32 dstSlot) {
...
+
+    // Return duplicate if it already exists
+    auto it = sRomToRamGfxVtxMap.find((void *) aVtx);
+    if (it != sRomToRamGfxVtxMap.end()) {
+        return (Vtx *) it->second.first;
+    }
+
+    // Duplicate vertex buffer and return the copy
+    if (shouldDuplicate) {
+        Vtx *vtxDuplicate = (Vtx *) malloc(vtxCount * sizeof(Vtx));
```
Isaac0-dev: `vtxCount * sizeof(Vtx)` is also calculated 3 times

Isaac0-dev (changes requested):

Isaac0-dev (approved):

PR #742, opened 2025-04-12
Another revamped update

Sunketchupm:
This PR really only exists to show that this update exists. It should not be merged for v1.3.1. All changes are revamped only unless otherwise specified.
Changelog:
* Dives no longer gain increased knockback based on speed. They instead have an increased horizontal distance knockback.
* Kicks have even more horizontal distance and even less vertical distance knockback.
* Fast punches (>24 speed) will no longer be able to combo, and instead deal 3 damage.
* Crouching decreases knockback taken
* * ! See commit for details on this
* Metal Mario can now have less than ×1 knockback.
* Ground pounds deal more damage the higher the fall. Every 700 units increases damage by 1, which is roughly triple jump height.


Sunketchupm:
Closing this for now, since there's a random display list conflict which I don't know what to do with and I'm moving revamped to its own dedicated branch

PR #746, opened 2025-04-15
make HOOK_ON_WARP trigger on instant warps

Blockyyy:
also made a new WARP_TYPE_INSTANT_WARP constant and removed remnant code


Beckowl:
I don't think `HOOK_ON_WARP` is appropriate for this. The displacement parameters are missing, and adding them to `HOOK_ON_WARP` would make it too bloated. This should be its own hook.

AgentXLP:
Agreed, if you add a HOOK_ON_INSTANT_WARP I will consider it.

PR #747, opened 2025-04-15
HOOK_ON_INSTANT_WARP

Blockyyy:
as suggested from #746 this adds a new hook instead of using HOOK_ON_WARP, this also makes it possible to get the displacement


AgentXLP:
Missing entry in hooks.md

src/game/level_update.c
```diff
@@ -56,7 +56,6 @@
 
 struct SavedWarpValues gReceiveWarp = { 0 };
 extern s8 sReceivedLoadedActNum;
-u8 gRejectInstantWarp = 0;
```
AgentXLP, line 59: Why was this removed

src/pc/lua/smlua_hooks.c
```diff
@@ -656,6 +656,33 @@ void smlua_call_event_hooks_warp_params(enum LuaHookedEventType hookType, u8 typ
...
+    for (int i = 0; i < hook->count; i++) {
+        s32 prevTop = lua_gettop(L);
+
+        // push the callback onto the stack
+        lua_rawgeti(L, LUA_REGISTRYINDEX, hook->reference[i]);
+
+        // push params
+        lua_pushinteger(L, area);
+        lua_pushinteger(L, warpId);
+        lua_pushinteger(L, displacementX);
```
AgentXLP: Don't you think this would be better as a Vec3s table?

src/pc/lua/smlua_hooks.h
```diff
@@ -126,6 +127,7 @@ static const char* LuaHookedEventTypeName[] = {
     "HOOK_ON_INTERACTIONS",
     "HOOK_ALLOW_FORCE_WATER_ACTION",
     "HOOK_BEFORE_WARP",
+    "HOOK_ON_INSTANT_WARP"
```
AgentXLP, line 130: No hooks.md entry, once again.

Blockyyy:
I do test my code, it was working fine and nobody really showed any concerns about what i was doing up until now, this is still fairly new to me and i am learning this alone as i go, i'll add it to hooks.md and i'll make it a vec3s.

I knew the HOOK_ON_WARP params sucked but i didn't want to make a new hook *just* for instant warps.

Also, for the room override thing, it wasn't consistent, and i had forgotten about the issue until it happened to me again, don't act like this never happened to you either when the same thing happened when you made the skybox lighting function.

src/game/level_update.c
Blockyyy: remnant, it wasn't needed anymore it wasn't doing anything (i was with sunk and told me it was good to be removed as well)

AgentXLP:
> don't act like this never happened to you either when the same thing happened when you made the skybox lighting function.
Touché

src/pc/lua/smlua_hooks.c
```diff
@@ -656,6 +656,50 @@ void smlua_call_event_hooks_warp_params(enum LuaHookedEventType hookType, u8 typ
...
+        lua_pushinteger(L, buf[0]);
+        lua_settable(L, tbl);
+
+        lua_pushstring(L, "y");
+        lua_pushinteger(L, buf[1]);
+        lua_settable(L, tbl);
+
+        lua_pushstring(L, "z");
+        lua_pushinteger(L, buf[2]);
+        lua_settable(L, tbl);
```
PeachyPeachSM64: `buf` is not needed here, just do `lua_pushinteger(L, displacement[...]);`

PeachyPeachSM64 (changes requested):

AgentXLP:
This should be good now I think?

PR #753, opened 2025-04-18
fix HOOK_BEFORE_WARP

Blockyyy:
Thank you emily for letting me know about the bug, this also adds `type` to the return table and params

### DIFF
--- a/docs/lua/guides/hooks.md
+++ b/docs/lua/guides/hooks.md
@@ -146,7 +146,7 @@ The lua functions sent to `hook_event()` will be automatically called by SM64 wh
 | HOOK_MARIO_OVERRIDE_GEOMETRY_INPUTS | Called before running Mario's geometry input logic, return `false` to not run it. | [MarioState](../structs.md) m | 
 | HOOK_ON_INTERACTIONS | Called when the Mario interactions are processed | [MarioState](../structs.md#MarioState) mario |
 | HOOK_ALLOW_FORCE_WATER_ACTION | Called when executing a non-water action while under the water's surface, or vice versa. Return `false` to prevent the player from being forced out of the action at the water's surface | [MarioState](../structs.md#MarioState) mario, `boolean` isInWaterAction |
-| HOOK_BEFORE_WARP | Called before the local player warps. Return a table with `destLevel`, `destArea`, `destAreaWarpNode`, and `arg` to override the warp | `integer` destLevel, `integer` destArea, `integer` destAreaWarpNode, `integer` arg |
+| HOOK_BEFORE_WARP | Called before the local player warps. Return a table with `type`, `destLevel`, `destArea`, `destAreaWarpNode`, and `arg` to override the warp | `integer` type, `integer` destLevel, `integer` destArea, `integer` destAreaWarpNode, `integer` arg |
 
 ### Parameters
 
--- a/src/game/level_update.c
+++ b/src/game/level_update.c
@@ -755,9 +755,7 @@ s16 music_changed_through_warp(s16 arg) {
 /**
  * Set the current warp type and destination level/area/node.
  */
-void initiate_warp(s16 destLevel, s16 destArea, s16 destWarpNode, s32 arg) {//
-
-    smlua_call_event_hooks_before_warp(HOOK_BEFORE_WARP, &destLevel, &destArea, &destWarpNode, &arg);
+void initiate_warp(s16 destLevel, s16 destArea, s16 destWarpNode, s32 arg) {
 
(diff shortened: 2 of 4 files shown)


Blockyyy:
https://github.com/user-attachments/assets/6cdb4972-cd72-410a-8cd6-1a8ca9e3eacc

src/game/level_update.c
```diff
@@ -769,6 +767,8 @@ void initiate_warp(s16 destLevel, s16 destArea, s16 destWarpNode, s32 arg) {//
         sWarpDest.type = WARP_TYPE_SAME_AREA;
     }
 
+    smlua_call_event_hooks_before_warp(HOOK_BEFORE_WARP, &sWarpDest.type, &destLevel, &destArea, &destWarpNode, &arg);
```
PeachyPeachSM64: how is moving the call a few lines down fixing the crashes?
actually it's even worse, because the warp type is no longer in sync with the destination if the user modifies one of the warp parameters
the user shouldn't have to handle the warp type themselves, it should be computed by the game after calling the hook
the problem of the previous PR was not the call to the hook itself, it was the lack of checking whether or not the warp is valid after being modified by the hook

you can still crash the game with the following script, if you "Exit to Castle" after entering BOB:
```lua
hook_event(HOOK_BEFORE_WARP, function (type, destLevel, destArea, destWarpNode, arg)
    return {
        type = gNetworkPlayers[0].currLevelNum ~= LEVEL_BOB and WARP_TYPE_CHANGE_LEVEL or WARP_TYPE_SAME_AREA,
        destLevel = LEVEL_BOB,
        destArea = 1,
        destWarpNode = 0x0A,
        arg = 0,
    }
end)
```

docs/lua/guides/hooks.md
```diff
@@ -146,7 +146,7 @@ The lua functions sent to `hook_event()` will be automatically called by SM64 wh
 | HOOK_MARIO_OVERRIDE_GEOMETRY_INPUTS | Called before running Mario's geometry input logic, return `false` to not run it. | [MarioState](../structs.md) m | 
 | HOOK_ON_INTERACTIONS | Called when the Mario interactions are processed | [MarioState](../structs.md#MarioState) mario |
 | HOOK_ALLOW_FORCE_WATER_ACTION | Called when executing a non-water action while under the water's surface, or vice versa. Return `false` to prevent the player from being forced out of the action at the water's surface | [MarioState](../structs.md#MarioState) mario, `boolean` isInWaterAction |
-| HOOK_BEFORE_WARP | Called before the local player warps. Return a table with `destLevel`, `destArea`, `destAreaWarpNode`, and `arg` to override the warp | `integer` destLevel, `integer` destArea, `integer` destAreaWarpNode, `integer` arg |
+| HOOK_BEFORE_WARP | Called before the local player warps. Return a table with `type`, `destLevel`, `destArea`, `destAreaWarpNode`, and `arg` to override the warp | `integer` type, `integer` destLevel, `integer` destArea, `integer` destAreaWarpNode, `integer` arg |
```
PeachyPeachSM64: it should be `destWarpNode`, not `destAreaWarpNode`

PeachyPeachSM64 (changes requested):

src/game/level_update.c
```diff
@@ -755,12 +755,14 @@ s16 music_changed_through_warp(s16 arg) {
...
  */
-void initiate_warp(s16 destLevel, s16 destArea, s16 destWarpNode, s32 arg) {//
+void initiate_warp(s16 destLevel, s16 destArea, s16 destWarpNode, s32 arg) {
 
     smlua_call_event_hooks_before_warp(HOOK_BEFORE_WARP, &destLevel, &destArea, &destWarpNode, &arg);
 
     if (destWarpNode >= WARP_NODE_CREDITS_MIN) {
         sWarpDest.type = WARP_TYPE_CHANGE_LEVEL;
+    } else if (arg == WARP_ARG_EXIT_COURSE) {
+        sWarpDest.type = WARP_TYPE_CHANGE_LEVEL;
```
PeachyPeachSM64, line 757: that seems to be a working solution
however, there is one specific case where this special value can cause a problem: if the destination warp happens to be a warp door, Mario could spawn on the wrong side of the door (since sWarpDest.arg is used to determine where to spawn for warp doors)
I would suggest to change the constant value to -1 instead of 1, and reset arg to 0 immediately after setting sWarpDest.type

PeachyPeachSM64 (changes requested):

src/game/level_update.c
Blockyyy: ah, nice catch

PR #756, opened 2025-04-19
Gfx/Vtx improvements

PeachyPeachSM64:
- Renamed the `new` and `realloc` functions to `create` and `resize`
- Added `delete_all`
- Made Mod Data a class:
  - Allocation is now limited to prevent out-of-memory crashes: 1024 display lists of max size 2048 and 1024 vertex buffers of max size 4096 per mod
  - Added error codes to identify the cause of a failure (name not found, pointer not found, max size exceeded, item pool is full, ...)


src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -134,16 +134,22 @@ void set_skybox_color(u8 index, u8 value) {
 static const Gfx SENTINEL_GFX[1] = {{{ _SHIFTL(G_ENDDL, 24, 8) | _SHIFTL(UINT32_MAX, 0, 24), UINT32_MAX }}};
 static const u8  SENTINEL_VTX[sizeof(Vtx)] = {[0 ... sizeof(Vtx) - 1] = UINT8_MAX};
 
-Gfx *gfx_allocate_internal(u32 length) {
-    if (!length) { return NULL; }
-    Gfx *gfx = calloc(length + 1, sizeof(Gfx));
+Gfx *gfx_allocate_internal(Gfx *gfx, u32 length) {
+    if (!gfx) {
+        gfx = calloc(length + 1, sizeof(Gfx));
```
Isaac0-dev: I should have thought of this earlier but maybe it would be good to have a constant like SENTINEL_SIZE just so people can understand what the +1 is for.

src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -134,16 +134,22 @@ void set_skybox_color(u8 index, u8 value) {
...
+    } else {
+        memset(gfx, 0, length * sizeof(Gfx));
```
Isaac0-dev, line 141: if gfx already exists:
here, you're resetting the content of the display list, not the sentinel.
But then, you overwrite the sentinel, even though it shouldn't have changed?
same with `vtx_allocate_internal`

src/pc/lua/utils/smlua_gfx_utils.c
```diff
@@ -272,45 +278,66 @@ void gfx_copy(Gfx *dest, Gfx *src, u32 length) {
...
+    Gfx *gfx = dynos_gfx_create(name, length);
     if (!gfx) {
-        LOG_LUA_LINE("gfx_new: Display list `%s` already exists", name);
+        switch (dynos_mod_data_get_last_error()) {
+            case DYNOS_MOD_DATA_ERROR_SIZE_IS_ABOVE_MAX:
+                LOG_LUA_LINE("gfx_create: Cannot allocate display list of length %u, max length exceeded", length); break;
+            case DYNOS_MOD_DATA_ERROR_ALREADY_EXISTS:
+                LOG_LUA_LINE("gfx_create: Display list `%s` already exists", name); break;
+            case DYNOS_MOD_DATA_ERROR_POOL_IS_FULL:
+                LOG_LUA_LINE("gfx_create: Cannot allocate more display lists"); break;
```
Isaac0-dev: "Cannot allocate more display lists" seems a little vague considering the default one is very similar.
Something like "Cannot allocate more display lists, limit reached" is more specific because it says to the user that they allocated too many.
Same with the vtx one.

data/dynos_mgr_gfx.cpp
```diff
@@ -205,32 +265,31 @@ Vtx *DynOS_Vtx_Get(const char *aName, u32 *outCount) {
     return NULL;
 }
 
-Vtx *DynOS_Vtx_New(const char *aName, u32 aCount) {
-    if (!aName || !aCount) { return NULL; }
+Vtx *DynOS_Vtx_Create(const char *aName, u32 aCount) {
     s32 modIndex = (gLuaActiveMod ? gLuaActiveMod->index : -1);
```
Isaac0-dev, line 269: just a comment: `gLuaActiveMod` is not reliable. 
not related to this pr really, but an improved way of getting the active mod should be made (separate to this pr)

data/dynos_mgr_moddata.hpp
```diff
@@ -1,104 +1,278 @@
...
-        return NULL;
+    inline ~ModData() {
+        Clear();
     }
 
-    if (outCount) {
-        *outCount = itItem->second.second;
+public:
+    ModDataResult<ModDataItem<T> *> Get(const char *aName) {
+        if (!aName) {
```
Isaac0-dev, line 34: check the string length and add an error for that maybe?
also could use an std::string for good practice, but that hardly matters.
goes for all the functions below that do the same

data/dynos_mgr_moddata.hpp
```diff
@@ -1,104 +1,278 @@
...
+    }
+
+    bool Resize(s32 aModIndex, T *aPointer, u32 aNewSize) {
+        ModDataT *modData = GetModData(aModIndex);
+        auto resizeResult = modData->Resize(aPointer, aNewSize);
+        if (!resizeResult.first) {
+            gDynosModDataLastError = resizeResult.second;
+            return false;
+        }
+        return resizeResult.first;
```
Isaac0-dev: this will always be true, since you checked it above. 
could just be useful to say that instead to be readable
same goes for the Delete function below.

docs/lua/examples/gfx-vtx-demo/update.lua
```diff
@@ -24,25 +29,27 @@ local id_bhvShape = hook_behavior(nil, OBJ_LIST_DEFAULT, true, bhv_shape_init, b
 local function mario_update(m)
     if m.playerIndex == 0 then
         if m.controller.buttonPressed & X_BUTTON ~= 0 then
-            local obj = obj_get_first_with_behavior_id(id_bhvShape)
-            if obj == nil then
-                obj = spawn_non_sync_object(id_bhvShape, E_MODEL_SHAPE, 0, 0, 0, nil)
+            shape_toggle = not shape_toggle
+            if shape_toggle then
+                spawn_non_sync_object(id_bhvShape, E_MODEL_SHAPE, 0, 0, 0, nil).oAction = 1
```
Isaac0-dev: just a note: this will error if `spawn_non_sync_object` returns nil (which it can)
the ugly solution is to have a callback that sets the action.
or just
```lua
local obj1 = spawn_non_sync_object(id_bhvShape, E_MODEL_SHAPE, 0, 0, 0, nil)
if obj1 then obj.oAction = 1 end
```
either way it doesn't matter, `spawn_non_sync_object` is less likely to fail than `spawn_sync_object`

src/pc/lua/utils/smlua_gfx_utils.c
PeachyPeachSM64: the function does the same if gfx is NULL or not: zeroing memory and adding the sentinel at offset length
the only difference is that if gfx is not allocated beforehand, the function does the allocation itself
the point is in the mod data pool, allocated buffers should never change their address once they're allocated

data/dynos_mgr_gfx.cpp
PeachyPeachSM64: what do you mean not reliable? is there some specific cases where gLuaActiveMod is NULL?
in that case, what would be a good alternative to retrieve the current mod index?

src/pc/lua/utils/smlua_gfx_utils.c
Isaac0-dev: Yes I can see that, but I was meaning that I think you can move the memcpy to only happen after the calloc, because if Gfx is not null, the sentinel should already be there

data/dynos_mgr_gfx.cpp
Isaac0-dev: No I don't think there will be any situation where it's null, but there can be situations where it points to a mod which actually did not call the Gfx/vtx function. This can happen with APIs, and mods can also intentionally spoof it too. You just need another mod to call a function belonging to your mod.

Isaac0-dev: Oh and there isn't any good alternative right now, but I was prototyping a function that can find what mod is active using Lua's debug tools..  don't need to worry about it right now, I was just making a note of it

src/pc/lua/utils/smlua_gfx_utils.c
PeachyPeachSM64: your length can be different

Isaac0-dev: Oh I see that you're right sorry about that

data/dynos_mgr_gfx.cpp
PeachyPeachSM64: oh I haven't thought about APIs, you're right

src/pc/lua/utils/smlua_gfx_utils.c
PeachyPeachSM64: not a constant, but I can add a comment

data/dynos_mgr_gfx.cpp
Isaac0-dev: I just pushed to the dev branch a better way to get the last active mod, `smlua_get_last_active_mod`
its not a very fast way to know, but its good when its critical that you have accuracy

Isaac0-dev (approved):

PR #757, opened 2025-04-19
Fix obj_get_model_id_extended returning wrong model ids

PeachyPeachSM64:
`obj_get_model_id_extended` can sometimes return a wrong extended model id, depending on what was loaded before.
For example, if you set the main menu to LLL, then enter CCM, all penguin objects will incorrectly return `E_MODEL_BULLY_BOSS` (instead of `E_MODEL_PENGUIN`)

The cause is not clear, but it seems that if two loaded vanilla models have the same vanilla id (`MODEL_PENGUIN` and `MODEL_BULLY_BOSS` have the same id of 0x57), the lowest `E_MODEL_` value is returned, regardless of the actual model (`E_MODEL_BULLY_BOSS` is 81, `E_MODEL_PENGUIN` is 103).

Not a real fix. If anyone has some knowledge about how the model loaded ids work, please take a look at this issue.

### DIFF
--- a/src/pc/lua/utils/smlua_model_utils.c
+++ b/src/pc/lua/utils/smlua_model_utils.c
@@ -504,12 +504,22 @@ u16 smlua_model_util_load(enum ModelExtendedId extId) {
     return (u16)id;
 }
 
+static void smlua_model_util_unregister_model_id(u32 id, struct ModelUtilsInfo *models, u32 count) {
+    for (u32 i = 0; i < count; i++) {
+        struct ModelUtilsInfo *m = &models[i];
+        if (m->loadedId == id) {
+            m->loadedId = UNLOADED_ID;
+        }
+    }
+}
+
 // Links the regular model id created by DynOS to our models list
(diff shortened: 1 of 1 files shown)


theclashingfritz:
This is behaving as intended when it comes to MODEL_PENGUIN and MODEL_BULLY_BOSS having the same id. This is because each model is reused in group sets. Which was done for some id reuse in the original game. The real fix is too either redo the id system or separate E_MODEL_ from the original model ids.

PeachyPeachSM64:
I know it's a limitation of the original game, but couldn't it be possible now to have both penguin and bully boss in the same level?

Isaac0-dev:
ah.
image

imo the best fix for this issue is to move everything we can to the extended model id system, and then have a system for translating regular model ids just for the situations we can't fully swap over (dynos levels)

AgentXLP:
I will merge this FOR NOW. And next update when Isaac's model id system rework is finished we can revert this if needed and merge his.

PR #765, opened 2025-04-22
[WIP] reworked model id system

Isaac0-dev:
Since there have been some issues with model ids recently, this will rework dynos' model management system to use the extended model ids. This will allow every model to have a unique id. Due to object macros, we can't fully replace the original model id system entirely. Instead, this will just replace as much as possible with the extended model id system, and will add a tool to translate a vanilla id into an extended model id.

some notes that I've made while planning this:
- Vanilla models always should have a static, consistent model id; but custom models are dynamic. Since extended model ids were first introduced for syncing models over the network, we need to keep them consistent over the network. Therefore, DynOS packs need to be separate to mod custom models, so dynos packs don't scramble other dynamic models.
- We can't fully remove the original model id system because:
Each "SpecialPreset" (a form of object macro) has a preset model, which uses the original model id system. 
If we have a special preset that uses a model which has multiple constants for the same object id (which is common), the extended model id system would have a new id for each constant. Special presets are used within the collision data of a level, which is loaded at runtime from the rom now. This means we can't add a new special preset. We have to have a translation layer that can determine which constant to use. 
- The original model id system was so limiting because it tried to fit every model id within a single byte. Our extended model id system already goes beyond the maximum of a single byte, so I've increased the size of (hopefully) each variable that stores a model id. In order to do this, I had to modify some level command macros.

The majority of the changes I've made are not here yet since my prototype has some bugs, and I need to more carefully plan how to implement it. This pr is just open as a placeholder, but discussion is more than welcome.


PeachyPeachSM64:
does that mean all already generated .lvl need to be generated again?

Isaac0-dev:
> does that mean all already generated .lvl need to be generated again?

no, the plan is to avoid that, else that would be very counter productive

PeachyPeachSM64:
but aren't the changes in data/dynos_level.cpp going to read invalid values on existing .lvl files?

Isaac0-dev:
oh you're right thanks for pointing that out. there will have to be a compatibility layer too.
this whole project may become more effort than its worth, but I'll still try

PR #767, opened 2025-04-23
Add some sanity checks to prevent bugs and exploits.

theclashingfritz:
Fix shell riding without a ridable object.
Fix everyone in a area being warped back to start if skip credits is trigged, Only the skipper warps now.
Fix being to able to trigger another blown cap off of Mario if you already have a blown off cap.


AgentXLP:
:trollface:

sher-b:
> Fix shell riding without a ridable object.

i was wondering in regards to this, does it patch out the shell mario glitch
i'd test it myself but i don't know how to

theclashingfritz:
> > Fix shell riding without a ridable object.
> 
> i was wondering in regards to this, does it patch out the shell mario glitch i'd test it myself but i don't know how to

It actually might, Which is a shame. But there's also a koopa shell duplication bug unique to Co-Op I believe it'd also patch out.

Blockyyy:
some kaizo hacks and mods actually require the shell mario glitch, it'd be nice imo to add this back if its patched

theclashingfritz:
> Contributor

That's true. I suppose it would be prudent to define what shell glitch the Shell Mario Glitch is. It's riding the shell in the non-ride state so you get the forward momentum? Or is it the glitch where you can ride an invisible shell? Only riding invisible shells is patched with this.

src/game/mario_actions_moving.c
```diff
@@ -1344,7 +1344,11 @@ s32 act_hold_decelerating(struct MarioState *m) {
 
 s32 act_riding_shell_ground(struct MarioState *m) {
     if (!m) { return FALSE; }
-    s16 startYaw = m->faceAngle[1];
+    
+    // If we don't have a object we're riding. Then the state is invalid or from a bug.
+    if (m->riddenObj == NULL || m->riddenObj->behavior != segmented_to_virtual(smlua_override_behavior(bhvKoopaShell))) {
```
PeachyPeachSM64: nitpick, but that prevents custom objects that have the INTERACT_KOOPA_SHELL interaction type from being ridden

src/game/interaction.c
```diff
@@ -375,37 +375,59 @@ u32 does_mario_have_normal_cap_on_head(struct MarioState *m) {
...
+
+    while (listHead != node) {
+        struct Object *obj = (struct Object *)node;
+        if (obj->behavior == behaviorAddr && obj->oBehParams == m->playerIndex + 1) {
+            return TRUE;
+        }
+
+        node = node->next;
+    }
+
```
PeachyPeachSM64: `obj_get_first_with_behavior_id_and_field_s32(id_bhvNormalCap, 0x40, m->playerIndex + 1) != NULL` does exactly what you coded here (0x40 is oBehParams)
also please change the return type to `bool` when applicable, Lua doesn't handle integers and booleans the same way

PeachyPeachSM64 (changes requested):

sher-b:
> > some kaizo hacks and mods actually require the shell mario glitch, it'd be nice imo to add this back if its patched
> 
> That's true. I suppose it would be prudent to define what shell glitch the Shell Mario Glitch is. Is it riding the shell in the non-ride state so you get the forward momentum? Or is it the glitch where you can ride an invisible shell? Only riding invisible shells is patched with this.

https://ukikipedia.net/wiki/Shell_Mario
in this case it would be the latter, although it'd be nice to test and see if the shell mario glitch is still there just to make sure it really hasn't been patched out

Sunketchupm:
A pretty consistent way to tell if Shell Mario works is to have Star Road enabled as well as a mod that lets you spawn a shell. Warp to course 1, spawn a shell, then just jump into out of bounds.

src/game/mario_actions_moving.c
theclashingfritz: That's a good catch. I'll adjust for that.

src/game/interaction.c
theclashingfritz: Fixed.

PeachyPeachSM64 (approved):

include/types.h
```diff
@@ -355,90 +355,113 @@ struct MarioAnimation
...
-    /*0xCA*/ struct Object* heldByObj;
-    /*????*/ u8 isSnoring;
-    /*????*/ struct Object* bubbleObj;
-    /*????*/ u8 freeze;
-
+    // Please try to keep this 64 bit aligned.
+    // I went through alot of trouble to group and sort
+    // all of these fields for size and optimization best I could.
+    // I reduced the size of this structure by 32 bytes for gcc.
+    // - Prince Frizzy
```
Isaac0-dev, line 367: please no personalised comments. and also please explain why it needs to be 64 bit aligned.

include/display_lists.inl
Isaac0-dev, line 1: are you able to remove the changes to `display_lists.inl` from this diff?

Isaac0-dev (dismissed):

Sunketchupm:
I will take a look at this later but I think this PR will patch out ride cloning. This isn't used anywhere as far as I can tell and is practically useless, but it is something to note.

theclashingfritz:
> I will take a look at this later but I think this PR will patch out ride cloning. This isn't used anywhere as far as I can tell and is practically useless, but it is something to note.

I believe it might-But the check can be disabled by mods if it's needed for whatever reason.

PR #768, opened 2025-04-24
Add G_PACKED_NORMALS Geo Mode from F3DEX3

ManIsCat2:
adds a new geometry mode called G_PACKED_NORMALS_EXT (originally called G_PACKED_NORMALS in f3dex3), this lets you use vertex colors and have lighting at the same time.

what this basically does is that it packs the normals that would be used for lighting into 1 integer and puts them in vn->flag, it unpacks that integer in the code and uses them as normals. vertex colors stay the same (they are still at v->cn)

you can export this via this fast64 plugin:

https://github.com/ManIsCat2/Fast64-Gart

This PR also fixes a bug in DynOS_Vtx_Parse where it should parse u16 but instead it was parsing a u8.

### DIFF
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -37,6 +37,8 @@
 #include "engine/lighting_engine.h"
 #include "pc/debug_context.h"
 
+#include "game/object_helpers.h"
+
 #define SUPPORT_CHECK(x) assert(x)
 
 // this is used for multi-textures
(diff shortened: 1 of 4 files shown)


src/pc/gfx/gfx_pc.c
```diff
@@ -37,6 +37,8 @@
 #include "engine/lighting_engine.h"
 #include "pc/debug_context.h"
 
+#include "game/object_helpers.h"
```
AgentXLP, line 40: Why is this included?

ManIsCat2: It's to get absi() for a SUPPORT_CHECK

AgentXLP: Oh I see

AgentXLP (approved):
Looks good

PR #778, opened 2025-04-30
Fix star doors

Sunketchupm:
The issue this fixes is for romhacks that have star doors with a requirement higher than 127. Since `oBehParams` is an `s32`, this means that any star door with a higher star requirement than 127 will require negative stars, which just means it'll always open.


PeachyPeachSM64:
ah yes good ol' invisible unsigned cast

PeachyPeachSM64:
actually no, `0xFF000000` is not guaranteed to be treated as an unsigned int, it's implementation-dependent
an explicit cast of oBehParams is required

AgentXLP:
Yeah, I'm gonna do peachy's method

PR #779, opened 2025-04-30
Fix warping while in act select

Sunketchupm:
Fixes this bug*:
https://github.com/user-attachments/assets/1d2d8961-c46b-4eb2-be29-77f0297ae5f1

Will be drafted until after v1.3.1, since this seems like the type of thing that will be very unstable when we least expect it.

\* This bug will still happen if a level happens to have an area at `MAX_AREAS - 1`

### DIFF
--- a/levels/menu/script.c
+++ b/levels/menu/script.c
@@ -65,13 +65,13 @@ const LevelScript level_main_menu_entry_2[] = {
     LOAD_MIO0(/*seg*/ 0x07, _menu_segment_7SegmentRomStart, _menu_segment_7SegmentRomEnd),
     ALLOC_LEVEL_POOL(),
 
-    AREA(/*index*/ 2, geo_menu_act_selector_strings),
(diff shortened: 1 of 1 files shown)


Isaac0-dev (commented):
i don't trust this fix, it seems like a hack. it must be due to some other bug. there shouldn't be any situation where another client's warp depends on what area you warp to

Sunketchupm:
Some mods, specifically forced warp mods, force all players into the same level and sometimes area that the host is in. Since the host is in the real level but in the menu level's area 2 while in act select, all players warp to the real level's area 2 while in act select.

Isaac0-dev:
sounds like its not a coop bug, but an issue with that mod. it can be fixed by modifying coop, but since there's no situation this can actually cause an issue normally, it doesn't really need to be fixed within coop

Sunketchupm:
Then I'll close this for now since it's probably not too hard to fix on the mod's end.

PR #781, opened 2025-04-30
Change misleading GraphNodeSwitchCase field label "numCases"

kermeow:
tl;dr the `numCases` field is in fact not actually the number of cases nor is it supposed to be

> This is an absolutely game destroying bug. Super Mario 64 Coop Deluxe is unplayable as long as this remains in the source code on GitHub.com. A field of the GraphNodeSwitchCase structure has **_incorrectly_** been named as `numCases`. Personally, this has thrown off my spiritual energy completely and I have lost all will to continue working on Super Mario 64 Coop Deluxe Super Mario Lua Scripting mods. Until this is fixed, I cannot continue to pursue this career.

Should be fine to merge because this field is both used incorrectly by Fast64 and labelled completely wrong in all versions of decomp so we can safely assume zero mods use it right now.

Pretty sure there is only one instance in the game where this field was actually used for "number of cases" and from looking at the function it was more likely supposed to be "number of anim states" (the relevant comment has been adjusted). 

- gave it a more ambiguous (but fitting) name because it can be used for basically anything
  - i actually took the name from another graph node with (probably) similar functionality
- made it mutable for modders

Will consider a follow up PR to Fast64 (they think its "default case" because of [yet another misleading comment about the same field](https://github.com/kermeow/sm64coopdx/blob/7b3d51d32b140ff6a71ec12a708893c577f7e488/src/engine/geo_layout.c#L345)) if I'm not lazy or nobody else wants to...

**This PR does not introduce an actual numCases functionality, it only amends a mistake in naming passed down from decomp.**

_looking at how it's used, this probably should've been spotted by decomp_

### DIFF
--- a/actors/number/geo.inc.c
+++ b/actors/number/geo.inc.c
@@ -33,7 +33,7 @@ Gfx *geo_num3_switch(s32 callContext, struct GraphNode *node, UNUSED void *conte
         s32 number = ((struct Object *) gCurGraphNodeObject)->oBehParams2ndByte;
         struct GraphNodeSwitchCase *switchCase = (struct GraphNodeSwitchCase *) node;
         s32 numMin, numMax, numDiv;
-        switch (switchCase->numCases) {
+        switch (switchCase->parameter) {
             case 0: numMin =   0; numMax =   9; numDiv =   1; break;
             case 1: numMin =  10; numMax =  99; numDiv =  10; break;
             case 2: numMin =  10; numMax =  99; numDiv =   1; break;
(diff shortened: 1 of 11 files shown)


Cooliokid956:
i fear that this is going to break my mod

Beckowl:
I think the best approach here would be something similar to what [Harmony](https://harmony.pardeike.net/articles/intro.html) does. It's a library that allows you to inject and run external code in C# applications.

Harmony provides two main types of patches: `prefix` and `postfix`. All prefix patches execute before the target function, and they can modify arguments or prevent the original function from running. If the function is allowed to proceed, it runs after all prefixes. After the original function runs (if it was not skipped), all postfix patches are executed, and they can modify the return value.

With just these features, you can create pretty cool mods, but implementing this in a safe way would be difficult (at least in my understanding)

PeachyPeachSM64:
> i fear that this is going to break my mod

no problem, we can just `union { s16 numCases; s16 parameter; }`

PR #789, opened 2025-05-03
Expose useful global graphnodes

Beckowl:
This makes it easier to access useful graph nodes without having to dig through the graph manually or rely on weird stuff like ``marioObj.header.gfx.node.parent.parent.parent`` to get to the camera node, for example.

The function names are a little too verbose for my liking, request changes if you have better ideas


AgentXLP:
These can be merged next update

Beckowl:
https://github.com/user-attachments/assets/894be1a2-66d2-4425-b8da-a65a11382a12

made the viewport dimensions/position mutable

```lua
--- name: geofunc
--- description: test

local id_bhvGeoFunc = hook_behavior(nil, OBJ_LIST_DEFAULT, true, function() end, function(o) o.header.gfx.skipInViewCheck = true end)
local E_MODEL_GEOFUNC = smlua_model_util_get_id("geofunc_geo")

local base_width = 120
local base_height = 90

function geo_function()
    local viewport = geo_get_current_root()

    local t = get_global_timer() * 0.1

    viewport.width = base_width + math.sin(t) * base_width / 4
    viewport.height = base_height + math.sin(t + math.pi / 2) * base_height / 4
end

function mario_update(m)
    if m.playerIndex ~= 0 then
        return
    end

    if (m.controller.buttonPressed & D_JPAD) ~= 0 then
        spawn_non_sync_object(id_bhvGeoFunc, E_MODEL_GEOFUNC, 0, 0, 0, nil)
    end
end

``

Isaac0-dev (changes requested):
`unk04` should be given a different name, preferably something like `graphNodeRoot`, or just `root`.

`unk15` should not be exposed as it's unused, its basically just padding

Isaac0-dev (approved):
i actually meant to make all cases of `unk04` have the new name, but this is fine

PR #792, opened 2025-05-05
More Optimizations

theclashingfritz:
A significant speedup is present but there is a weird bug where Goombas never die. And Water Bubbles kill you in one hit. I'm not sure as to why.

Please do testing as more bugs are likely present.


include/types.h
```diff
@@ -120,63 +126,84 @@ struct AnimationTable {
...
     struct GraphNode *next;
-    u8 _guard2;
     struct GraphNode *parent;
     struct GraphNode *children;
     const void *georef;
+    
+    s16 type; // structure type
+    s16 flags; // hi = drawing layer, lo = rendering modes
+    
+    u8 _guard1;
```
Isaac0-dev: By moving the guards next to each other, it mostly defeats the point of it. The idea was to catch memory corruption, and this will be less effective.

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
+/*
+.I files are for inlined functions (sometimes forced). These functions are seperated from the .c/.cpp files so they 
+can be recompiled by the compiler if needed for further optimizations.
```
PeachyPeachSM64: Not really. `.I` or `.i` files are for intermediate files, i.e. files that have been fully preprocessed but not compiled yet.
`.inl` is the common extension for "inline" files.

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+INLINE OPTIMIZE_O3 f32 *vec3f_sum(Vec3f dest, Vec3f a, Vec3f b) {
+    dest[0] = a[0] + b[0];
+    dest[1] = a[1] + b[1];
+    dest[2] = a[2] + b[2];
+    return dest;
+}
+
+/// Subtracts vector a from 'dest'.
+INLINE OPTIMIZE_O3 f32 *vec3f_sub(Vec3f dest, Vec3f a) {
+    if (!dest || !a) { return NULL; }
```
PeachyPeachSM64: why the NULL check here? `Vec3f` is not a pointer type

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+*/
+
+#ifndef MATH_UTIL_I
+#define MATH_UTIL_I
+
+/*
+Vec3f Functions
+*/
+
+/// Copy vector 'src' to 'dest'
```
PeachyPeachSM64: Use `|description| ... |descriptionEnd|` for autogen to work properly. If autogen doesn't reach this file, it would be good to update it as well.

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+// Normalizes each component of the 3D floating-point vector 'v'.
+INLINE OPTIMIZE_O3 f32 *vec3f_normalize2(Vec3f v) {
+    float s = vec3f_length(v);
+    vec3f_div(v, s);
+    return v;
+}
+
+/// Get length of vector 'a'
+INLINE OPTIMIZE_O3 f32 vec3f_length(Vec3f a) {
+	return sqrtf(a[0] * a[0] + a[1] * a[1] + a[2] * a[2]);
```
PeachyPeachSM64: tab instead of spaces

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/// Get dot product of vectors 'a' and 'b'
+INLINE OPTIMIZE_O3 f32 vec3f_dot(Vec3f a, Vec3f b) {
+	return a[0] * b[0] + a[1] * b[1] + a[2] * b[2];
```
PeachyPeachSM64: same here

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/// Takes respective scales of vecA and vecB, and sums them
+INLINE OPTIMIZE_O3 f32 *vec3f_combine(Vec3f dest, Vec3f vecA, Vec3f vecB, f32 sclA, f32 sclB) {
+    for (int i = 0; i < 3; ++i) {
```
PeachyPeachSM64: please use `s32` instead of `int`

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+*/
+
+/// Copy matrix 'src' to 'dest'
+INLINE OPTIMIZE_O3 void mtxf_copy(Mat4 dest, Mat4 src) {
+    u32 *d = (u32 *) dest;
+    u32 *s = (u32 *) src;
+
+    for (s32 i = 0; i < 16; i++) {
+        *d++ = *s++;
+    }
```
PeachyPeachSM64: no way this is beating `memcpy(dest, src, sizeof(Mat4))` in performance
also why no return value for the mtxf functions?

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+    s32 i;
+    f32 *dest;
+    
+    // These loops must be one line to match on -O2
+
+    // Initialize everything except the first and last cells to 0
+    for (dest = (f32 *) mtx + 1, i = 0; i < 14; dest++, i++) *dest = 0;
+
+    // Initialize the diagonal cells to 1
+    for (dest = (f32 *) mtx, i = 0; i < 4; dest += 5, i++) *dest = 1;
```
PeachyPeachSM64: yeah I know this is old decomp code, but these functions should really be optimized
this is almost 7 times faster on modern CPUs:
```c
    static const Mat4 sMat4Identity = {
        { 1, 0, 0, 0 },
        { 0, 1, 0, 0 },
        { 0, 0, 1, 0 },
        { 0, 0, 0, 1 },
    };
    memcpy(mtx, sMat4Identity, sizeof(Mat4));
```

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/**
+ * Set dest to a translation matrix of vector b
+ */
+INLINE OPTIMIZE_O3 void mtxf_translate(Mat4 dest, Vec3f b) {
+    mtxf_identity(dest);
+    dest[3][0] = b[0];
+    dest[3][1] = b[1];
+    dest[3][2] = b[2];
```
PeachyPeachSM64: could be `vec3f_copy(dest[3], b);`

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Set matrix 'dest' to 'mtx' scaled by vector s
+ */
+INLINE OPTIMIZE_O3 void mtxf_scale_vec3f(Mat4 dest, Mat4 mtx, Vec3f s) {
+    for (s32 i = 0; i < 4; i++) {
+        dest[0][i] = mtx[0][i] * s[0];
+        dest[1][i] = mtx[1][i] * s[1];
+        dest[2][i] = mtx[2][i] * s[2];
+        dest[3][i] = mtx[3][i];
+    }
```
PeachyPeachSM64: same here, avoid duplicating code
```c
    mtxf_copy(dest, mtx);
    vec3f_mul(dest[0], s[0]);
    vec3f_mul(dest[1], s[1]);
    vec3f_mul(dest[2], s[2]);
```
(also appears to be slightly faster)

src/engine/math_util.c
```diff
@@ -132,58 +32,39 @@ void *find_vector_perpendicular_to_plane(Vec3f dest, Vec3f a, Vec3f b, Vec3f c)
...
 
-/// Make vector 'dest' the cross product of vectors a and b.
-void *vec3f_cross(Vec3f dest, Vec3f a, Vec3f b) {
-    dest[0] = a[1] * b[2] - b[1] * a[2];
-    dest[1] = a[2] * b[0] - b[2] * a[0];
-    dest[2] = a[0] * b[1] - b[0] * a[1];
-    return dest;
+#pragma GCC diagnostic pop
+
+void vec3f_project(Vec3f vec, Vec3f onto, Vec3f out) {
```
PeachyPeachSM64: to keep consistency with other functions, order and name of arguments should be `Vec3f dest, Vec3f v, Vec3f onto`

src/engine/math_util.c
```diff
@@ -132,58 +32,39 @@ void *find_vector_perpendicular_to_plane(Vec3f dest, Vec3f a, Vec3f b, Vec3f c)
...
 /// Scale vector 'dest' so it has length 1
-void *vec3f_normalize(Vec3f dest) {
-    f32 div = sqrtf(dest[0] * dest[0] + dest[1] * dest[1] + dest[2] * dest[2]);
+OPTIMIZE_O3 f32 *vec3f_normalize(Vec3f dest) {
+    f32 div = vec3f_length(dest);
     if (div == 0) {
-        dest[0] = 0;
-        dest[1] = 0;
-        dest[2] = 0;
+        vec3f_set(dest, 0, 0, 0);
```
PeachyPeachSM64, line 67: it would be good to have `vec3s_zero`, `vec3f_zero` and `mtxf_zero` functions

src/engine/math_util.c
```diff
@@ -203,46 +84,32 @@ void *vec3f_rotate_zxy(Vec3f dest, Vec3s rotate) {
...
-    register s32 i;
-    register f32 *dest;
-    // These loops must be one line to match on -O2
-
-    // initialize everything except the first and last cells to 0
-    for (dest = (f32 *) mtx + 1, i = 0; i < 14; dest++, i++) *dest = 0;
+void vec3f_get_dist_and_angle(Vec3f from, Vec3f to, f32 *dist, s16 *pitch, s16 *yaw) {
+    register f32 x = to[0] - from[0];
+    register f32 y = to[1] - from[1];
+    register f32 z = to[2] - from[2];
```
PeachyPeachSM64: `register` keyword should not be used, compilers are better than us to tell what to put into registers

src/game/mario_actions_cutscene.c
```diff
@@ -2247,7 +2247,7 @@ static s32 jumbo_star_cutscene_taking_off(struct MarioState *m) {
 
     if (m->actionState == 0) {
         set_character_animation(m, CHAR_ANIM_FINAL_BOWSER_RAISE_HAND_SPIN);
-        marioObj->rawData.asF32[0x22] = 0.0f;
+        marioObj->oMarioUnkPosY = 0.0f;
```
PeachyPeachSM64: not Y, but Z
a better name would be `oMarioJumboStarCutscenePosZ`

src/game/spawn_object.c
```diff
@@ -292,8 +292,8 @@ struct Object *allocate_object(struct ObjectNode *objList) {
     obj->numCollidedObjs = 0;
 
     for (s32 i = 0; i < OBJECT_NUM_FIELDS; i++) {
-        obj->rawData.asS32[i] = 0;
-        obj->ptrData.asVoidPtr[i] = NULL;
+        obj->OBJECT_FIELD_U32(i) = 0;
+        obj->OBJECT_FIELD_VPTR(i) = NULL;
     }
```
PeachyPeachSM64: `memset(obj->rawData, 0, sizeof(obj->rawData));`?
`memset(obj->ptrData, 0, sizeof(obj->ptrData));`?

src/engine/math_util.I
```diff
@@ -0,0 +1,236 @@
...
+    dest[2] = a[0] * b[1] - b[0] * a[1];
+    return dest;
+}
+
+// Normalizes each component of the 3D floating-point vector 'v'.
+INLINE OPTIMIZE_O3 f32 *vec3f_normalize2(Vec3f v) {
+    float s = vec3f_length(v);
+    vec3f_div(v, s);
+    return v;
+}
```
PeachyPeachSM64: uh, isn't that just `vec3f_normalize`? why does it need a second definition?

src/pc/gfx/gfx_pc.c
```diff
@@ -703,24 +697,24 @@ static void calculate_normal_dir(const Light_t *light, float coeffs[3], bool app
...
+        for (int32_t j = 0; j < 4; j++) {
             tmp[i][j] = a[i][0] * b[0][j] +
                         a[i][1] * b[1][j] +
                         a[i][2] * b[2][j] +
                         a[i][3] * b[3][j];
         }
     }
-    memcpy(res, tmp, sizeof(tmp));
+    mtxf_copy(res, tmp);
 }
```
PeachyPeachSM64: that's `mtxf_mul`, no need for a duplicate
speaking of `mtxf_` functions, some should be updated to use SSE like `gfx_sp_vertex`

src/pc/gfx/gfx_pc.c
```diff
@@ -737,17 +731,20 @@ static void gfx_sp_matrix(uint8_t parameters, const int32_t *addr) {
 
     if (parameters & G_MTX_PROJECTION) {
         if (parameters & G_MTX_LOAD) {
-            memcpy(rsp.P_matrix, matrix, sizeof(matrix));
+            //memcpy(rsp.P_matrix, matrix, sizeof(matrix));
```
PeachyPeachSM64: why the comment? just remove the line

src/pc/gfx/gfx_pc.c
```diff
@@ -737,17 +731,20 @@ static void gfx_sp_matrix(uint8_t parameters, const int32_t *addr) {
...
+            mtxf_copy(rsp.P_matrix, matrix);
         } else {
             gfx_matrix_mul(rsp.P_matrix, matrix, rsp.P_matrix);
         }
     } else { // G_MTX_MODELVIEW
         if ((parameters & G_MTX_PUSH) && rsp.modelview_matrix_stack_size < 11) {
             ++rsp.modelview_matrix_stack_size;
-            memcpy(rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 1], rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 2], sizeof(matrix));
+            //memcpy(rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 1], rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 2], sizeof(matrix));
```
PeachyPeachSM64: same here

src/pc/gfx/gfx_pc.c
```diff
@@ -737,17 +731,20 @@ static void gfx_sp_matrix(uint8_t parameters, const int32_t *addr) {
...
+            mtxf_copy(rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 1], rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 2]);
         }
         if (parameters & G_MTX_LOAD) {
-            memcpy(rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 1], matrix, sizeof(matrix));
+            //memcpy(rsp.modelview_matrix_stack[rsp.modelview_matrix_stack_size - 1], matrix, sizeof(matrix));
```
PeachyPeachSM64: same here

PeachyPeachSM64 (changes requested):

Blockyyy:
Couldn't test out everything today, i'll resume when i can (excuse the background noise, forgot to not record discord)

weird bug where only blue coins increment the coin counter/heal you:

https://github.com/user-attachments/assets/9f9bd802-f6b2-4813-9311-df456b4b6e05
https://github.com/user-attachments/assets/7287650f-2d80-41dc-b52d-5fe471299531

Skeeters, Amps, Mr Blizzards & their projectiles, Unagi, Pokeys All one shot you:

https://github.com/user-attachments/assets/591fd740-ba59-4633-bcaa-34558c52f1f1
https://github.com/user-attachments/assets/3a46a701-c305-418f-bb32-143e3e916f0f
https://github.com/user-attachments/assets/b780566a-6609-4981-bcf2-c6cf70e548b5
https://github.com/user-attachments/assets/b8eea50a-9211-466f-b062-07cef002d693

Skeeters, Pokeys also can't be killed

https://github.com/user-attachments/assets/ca559f90-de27-4856-8e03-39a9b04531a5
https://github.com/user-attachments/assets/3c90efa1-7da4-47ed-9ef8-58a5e5109967

Fly guys don't die either (and also deal 0 damage):
https://github.com/user-attachments/assets/61a236b4-de6e-4ec4-9e58-2c15c76f4589
https://github.com/user-attachments/assets/7a20101e-3bed-4193-9770-cbe517f10717

Clams, Spindrifts, Moneybags don't do damage:

https://github.com/user-attachments/assets/ef62ebee-2bfb-4bea-91b6-ea95c0c2b1e8
https://github.com/user-attachments/assets/64596612-d34a-4564-9a6d-a39930bda4a6

Mr I doesn't do damage, but the projectile does:

https://github.com/user-attachments/assets/9a625a99-8045-4d99-9b85-8422252393ad

whatever happened with eyerock:

https://github.com/user-attachments/assets/bc4b423b-8b4c-43e8-bac0-c42d9a29eb2b

Crazy box drops one coin:

https://github.com/user-attachments/assets/9b55f08a-d4fc-413a-940b-86d0942c812a

PeachyPeachSM64:
> weird bug where only blue coins increment the coin counter/heal you
> Skeeters, Amps, Mr Blizzards & their projectiles, Unagi, Pokeys All one shot you
> Skeeters, Pokeys also can't be killed
> Fly guys don't die either (and also deal 0 damage)
> Clams, Spindrifts, Moneybags don't do damage
> Mr I doesn't do damage, but the projectile does
> whatever happened with eyerock
> Crazy box drops one coin

That's because of the `struct ObjectHitbox` change: all hitboxes declared in object behaviors files assume the order of the fields is the one before the change, meaning now some fields are filled with wrong values: the `downOffset` value becomes `damageOrCoinValue`, the `damageOrCoinValue` becomes `health` and so on...
A lot of structures in the game are declared like this, and that's just how C works.

src/engine/math_util.I
theclashingfritz: Well I don't know everything. I'll change it. Thank you

src/engine/math_util.I
theclashingfritz: C treats these types similar to pointers, Which is why they are able to be updated without using a pointer. 

I'm not entirely sure if C++ acts the same but I haven't had issues there so far.

src/engine/math_util.I
theclashingfritz: I did at one point make autogen reach this file but since it already reaches the .h file with pre-declarations. It wasn't needed.

src/engine/math_util.I
theclashingfritz: It was taken from elsewhere in the code. I wasn't sure if updating it to the other one would cause issues so I left it as a separate function.

include/types.h
theclashingfritz: Noted. I'll move them. Thank you

src/engine/math_util.I
theclashingfritz: Fixed. Thank you

src/engine/math_util.I
theclashingfritz: Mxtf functions don't have a return value because you can't just return a 'f32 *'. I actually did try and got compile errors. Maybe a 'f32 **' would work instead?

src/engine/math_util.I
theclashingfritz: Yeah. I actually planned to do the same for this one.

src/engine/math_util.I
theclashingfritz: Thank you.

src/engine/math_util.c
theclashingfritz: This is how it was before. I do agree but I was worried on breaking mods with it.

src/pc/gfx/gfx_pc.c
theclashingfritz: I have been looking into the SSE for this.

src/engine/math_util.I
theclashingfritz: As for the difference in memcpy-I don't actually know in this case. But that's probably true for modern CPUs

src/engine/math_util.I
theclashingfritz: Thank you.

src/engine/math_util.I
theclashingfritz: As for why it's there. It was a carryover from the Camera files. Which I didn't want to crash if they really needed this check for some reason.

src/engine/math_util.I
theclashingfritz: Thank you.

src/engine/math_util.I
theclashingfritz: Thank you.

theclashingfritz:
> > weird bug where only blue coins increment the coin counter/heal you
> > Skeeters, Amps, Mr Blizzards & their projectiles, Unagi, Pokeys All one shot you
> > Skeeters, Pokeys also can't be killed
> > Fly guys don't die either (and also deal 0 damage)
> > Clams, Spindrifts, Moneybags don't do damage
> > Mr I doesn't do damage, but the projectile does
> > whatever happened with eyerock
> > Crazy box drops one coin
> 
> That's because of the `struct ObjectHitbox` change: all hitboxes declared in object behaviors files assume the order of the fields is the one before the change, meaning now some fields are filled with wrong values: the `downOffset` value becomes `damageOrCoinValue`, the `damageOrCoinValue` becomes `health` and so on... A lot of structures in the game are declared like this, and that's just how C works.
> 
> Some examples:
> 
> * Amp: They deal 40 damage and have 1 health
> 
> ```c
> static struct ObjectHitbox sAmpHitbox = {
>     /* interactType:      */ INTERACT_SHOCK,
>     /* damageOrCoinValue: */ 40,
>     /* health:            */ 1,
>     /* numLootCoins:      */ 0,
>     /* downOffset:        */ 0,
>     /* radius:            */ 40,
>     /* height:            */ 50,
>     /* hurtboxRadius:     */ 50,
>     /* hurtboxHeight:     */ 60,
> };
> ```
> 
> * Eyerok: They have no health, meaning they break in one hit, but gives 4 coins
> 
> ```c
> struct ObjectHitbox sEyerokHitbox = {
>     /* interactType:      */ INTERACT_BOUNCE_TOP,
>     /* damageOrCoinValue: */ 0,
>     /* health:            */ 0,
>     /* numLootCoins:      */ 4,
>     /* downOffset:        */ 0,
>     /* radius:            */ 150,
>     /* height:            */ 100,
>     /* hurtboxRadius:     */ 1,
>     /* hurtboxHeight:     */ 1,
> };
> ```
> 
> * Mr. Blizzard, the funny one: gives a whopping 99 coins on death!
> 
> ```c
> struct ObjectHitbox sMrBlizzardHitbox = {
>     /* interactType:      */ INTERACT_MR_BLIZZARD,
>     /* damageOrCoinValue: */ 24,
>     /* health:            */ 2,
>     /* numLootCoins:      */ 99,
>     /* downOffset:        */ 3,
>     /* radius:            */ 65,
>     /* height:            */ 170,
>     /* hurtboxRadius:     */ 65,
>     /* hurtboxHeight:     */ 170,
> };
> ```
> 
> image

Yikes I didn't think of that. Thanks for pointing this out.

src/game/spawn_object.c
theclashingfritz: You're right. That'd be better

src/engine/math_util.I
PeachyPeachSM64: oh I didn't see the description was already in the .h file
you can ignore this point then

src/engine/math_util.I
PeachyPeachSM64: ah yeah mtxf would be f32**
let's keep void in that case

Isaac0-dev:
> another thing I just noticed most of `vec3f` functions have a C version and a Lua version in autogen, for instance: `function vec3f_project(vec, onto)` in `src/pc/lua/smlua_constants_autogen.c` and `smlua_bind_function(L, "vec3f_project", smlua_func_vec3f_project)` in `src/pc/lua/smlua_functions_autogen.c` and the signatures are not the same
> 
> maybe it would be good to remove the Lua definitions in favor of the C ones?

for performance, yes i think that'd be faster

theclashingfritz:
> > > weird bug where only blue coins increment the coin counter/heal you
> > > Skeeters, Amps, Mr Blizzards & their projectiles, Unagi, Pokeys All one shot you
> > > Skeeters, Pokeys also can't be killed
> > > Fly guys don't die either (and also deal 0 damage)
> > > Clams, Spindrifts, Moneybags don't do damage
> > > Mr I doesn't do damage, but the projectile does
> > > whatever happened with eyerock
> > > Crazy box drops one coin
> > 
> > 
> > That's because of the `struct ObjectHitbox` change: all hitboxes declared in object behaviors files assume the order of the fields is the one before the change, meaning now some fields are filled with wrong values: the `downOffset` value becomes `damageOrCoinValue`, the `damageOrCoinValue` becomes `health` and so on... A lot of structures in the game are declared like this, and that's just how C works.
> > Some examples:
> > 
> > * Amp: They deal 40 damage and have 1 health
> > 
> > ```c
> > static struct ObjectHitbox sAmpHitbox = {
> >     /* interactType:      */ INTERACT_SHOCK,
> >     /* damageOrCoinValue: */ 40,
> >     /* health:            */ 1,
> >     /* numLootCoins:      */ 0,
> >     /* downOffset:        */ 0,
> >     /* radius:            */ 40,
> >     /* height:            */ 50,
> >     /* hurtboxRadius:     */ 50,
> >     /* hurtboxHeight:     */ 60,
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Eyerok: They have no health, meaning they break in one hit, but gives 4 coins
> > 
> > ```c
> > struct ObjectHitbox sEyerokHitbox = {
> >     /* interactType:      */ INTERACT_BOUNCE_TOP,
> >     /* damageOrCoinValue: */ 0,
> >     /* health:            */ 0,
> >     /* numLootCoins:      */ 4,
> >     /* downOffset:        */ 0,
> >     /* radius:            */ 150,
> >     /* height:            */ 100,
> >     /* hurtboxRadius:     */ 1,
> >     /* hurtboxHeight:     */ 1,
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Mr. Blizzard, the funny one: gives a whopping 99 coins on death!
> > 
> > ```c
> > struct ObjectHitbox sMrBlizzardHitbox = {
> >     /* interactType:      */ INTERACT_MR_BLIZZARD,
> >     /* damageOrCoinValue: */ 24,
> >     /* health:            */ 2,
> >     /* numLootCoins:      */ 99,
> >     /* downOffset:        */ 3,
> >     /* radius:            */ 65,
> >     /* height:            */ 170,
> >     /* hurtboxRadius:     */ 65,
> >     /* hurtboxHeight:     */ 170,
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> 
> Yikes I didn't think of that. Thanks for pointing this out.

This should hopefully now be fixed. Please retest and verify.

include/object_fields.h
```diff
@@ -7,12 +7,16 @@
  */
 
 #ifdef OBJECT_FIELDS_INDEX_DIRECTLY
+#define OBJECT_FIELD_U16(index, subIndex) index + subIndex
+#define OBJECT_FIELD_S16(index, subIndex) index + subIndex
```
PeachyPeachSM64, line 11: these fields shouldn't appear here
it's meant to be used with the get/set/add/or_int/float functions and behavior commands, but they always expect 32-bit values

include/types.h
```diff
@@ -95,88 +99,116 @@ struct VblankHandler
...
-    /*0x04*/ s16 startFrame;
-    /*0x06*/ s16 loopStart;
-    /*0x08*/ s16 loopEnd;
-    /*0x0A*/ s16 unusedBoneCount;
-    /*0x0C*/ u16 *values;
-    /*0x10*/ u16 *index;
-    /*0x14*/ u32 length; // only used with Mario animations to determine how much to load. 0 otherwise.
-    /*????*/ u32 valuesLength;
-    /*????*/ u32 indexLength;
+    // TODO: Optimize this later if possible.
```
PeachyPeachSM64, line 105: TODO
same as ObjectHitbox, but this struct is largely more used

src/game/mario_step.c
```diff
@@ -16,9 +16,22 @@
...
+    { 0, 0, 0 }, // Vertex 2
+    { 0, 0, 0 }, // Vertex 3
+    { 0, 0, 0 }, // Prev Vertex 1
+    { 0, 0, 0 }, // Prev Vertex 2
+    { 0, 0, 0 }, // Prev Vertex 3
+    { 0.0f, 1.0f, 0.0f }, // Normal
+    0.0f, // Origin Offset
+    0, // Modified Timestamp
+    NULL // Object
 };
```
PeachyPeachSM64: might be good to do
```c
.type = ...,
.flags = ...,
.room = ...,
...
```
like the hitboxes in case the struct layout is changed in the future

src/pc/platform.h
```diff
@@ -4,6 +4,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
+#include "platform_info.h"
```
PeachyPeachSM64: why is it needed?

PeachyPeachSM64 (changes requested):

src/game/spawn_object.c
```diff
@@ -291,10 +291,8 @@ struct Object *allocate_object(struct ObjectNode *objList) {
     obj->collidedObjInteractTypes = 0;
     obj->numCollidedObjs = 0;
 
-    for (s32 i = 0; i < OBJECT_NUM_FIELDS; i++) {
-        obj->rawData.asS32[i] = 0;
-        obj->ptrData.asVoidPtr[i] = NULL;
-    }
+    memset(&obj->rawData, 0, sizeof(obj->rawData));
+    memset(&obj->ptrData, 0, sizeof(obj->ptrData));
```
Isaac0-dev, line 295: massive hack for performance:
since `rawData` and `ptrData` are next to each other in memory, so you could just
`memset(&obj->rawData, 0, sizeof(obj->rawData) + sizeof(obj->ptrData));`
and add a static assert to be sure they're next to each other
not really necessary since this function is an event, not a regular call

but then again, why not just `memset` the whole Object struct, or at least parts of it that are needed, and apply the flags that aren't normally 0 after?

src/pc/platform.h
Isaac0-dev: is it needed? :thinking:

src/engine/math_util.inl
```diff
@@ -0,0 +1,243 @@
+/*
+.inl files are for inlined functions and function templates.
+
+It's best to put them in this file so they can be recompiled if needed.
+
+Credit to PeachyPeach, Issac, Blockyyy, and others for suggestions
```
Isaac0-dev: my name is Isaac

src/game/spawn_object.c
theclashingfritz: You can't memset the Object structure because of the header. The header is a linked list which is setup beforehand

Isaac0-dev: That doesn't stop you from skipping the header

theclashingfritz: Yeah your right it doesn't, 
But I'm not going to do something like that because it can cause issues if things are moved around or if sizes change for whatever reason. 
This is a function for object reinitialization-That's why the linked list is already setup to begin with, Zeroing out everything like that isn't necessary.

include/types.h
theclashingfritz: It is pretty similar. But this one is used in mods for actors more then likely. So I'm not sure about optimizing it till later on. If it turns out it'd invalidate old mods. It'd be best to do it with a bigger update.

src/game/spawn_object.c
theclashingfritz: If we redesign the Object Linked list however. It wouldn't be completely against the idea. But doing memset on offsets in structures is dangerous as it is. It's a recipe for possible overflows into unrelated memory which I don't want to risk unless we absolutely have too.

Isaac0-dev: that's why i mentioned the static assert, you can check at compile time if the fields are next to each other in memory and throw an error if they aren't. 
either way its definitely a hack so i'm not expecting that change to be made, i was just mentioning what i noticed

src/game/mario_step.c
theclashingfritz: Actually. That's a good idea. Thank you.

include/object_fields.h
theclashingfritz: Looking into it. You're correct. I'll go ahead and remove these.

src/pc/platform.h
theclashingfritz: Thanks for pointing this out. This is a remnant of a scrapped code change. It's not needed and I'll remove it.

src/engine/math_util.inl
theclashingfritz: Oops sorry. I'll correct it.

include/object_fields.h
theclashingfritz: Tried doing that. Results in a compile error because of the defines in object_fields.h itself in the files that need to use the index based defines. So they will remain. 
image

theclashingfritz:
As of now. I'd say this PR is ready for proper review and merging. I wouldn't want to add any more to this PR and instead put any more optimizations into another separate PR. As I don't want this PR to get too big and then have a big issue show up.

theclashingfritz:
> > another thing I just noticed most of `vec3f` functions have a C version and a Lua version in autogen, for instance: `function vec3f_project(vec, onto)` in `src/pc/lua/smlua_constants_autogen.c` and `smlua_bind_function(L, "vec3f_project", smlua_func_vec3f_project)` in `src/pc/lua/smlua_functions_autogen.c` and the signatures are not the same
> > maybe it would be good to remove the Lua definitions in favor of the C ones?
> 
> for performance, yes i think that'd be faster

I went ahead and did this. And moved the old LUA functions to be deprecated. Some were not yet implemented in C so I went ahead and implemented those also.

autogen/lua_constants/deprecated.lua
```diff
@@ -1,2 +1,173 @@
...
+
+--- @param v1 Vec3s
+--- @param v2 Vec3s
+--- @return number
+function vec3s_dist(v1, v2)
+    dx = v1.x - v2.x
+    dy = v1.y - v2.y
+    dz = v1.z - v2.z
+    return math.sqrt(dx * dx + dy * dy + dz * dz)
+end
```
PeachyPeachSM64: why do these definitions appear here now? they should be removed completely from lua

src/engine/math_util.inl
```diff
@@ -141,52 +167,109 @@ INLINE OPTIMIZE_O3 s16 *vec3f_to_vec3s(Vec3s dest, Vec3f a) {
...
+ * Get length of vector 'a'.
+ */
+INLINE OPTIMIZE_O3 s32 vec3s_length(Vec3s a) {
+    return sqrt(a[0] * a[0] + a[1] * a[1] + a[2] * a[2]);
+}
+
+/**
+ * Calculates the distance between point 'v1' and 'v2'.
+ */
+INLINE OPTIMIZE_O3 s32 vec3s_dist(Vec3s v1, Vec3s v2) {
```
PeachyPeachSM64: imo `length` and `dist` should return `f32`, even for integer vectors

src/engine/math_util.inl
```diff
@@ -198,14 +281,27 @@ INLINE OPTIMIZE_O3 f32 *vec3s_to_vec3f(Vec3f dest, Vec3s a) {
...
+   _mm_store_ps(dest[2], r3);
+   _mm_store_ps(dest[3], r4);
+#else
     u32 *d = (u32 *) dest;
     u32 *s = (u32 *) src;
 
     for (s32 i = 0; i < 16; i++) {
         *d++ = *s++;
     }
+#endif
```
PeachyPeachSM64: unnecessary, just memcpy

src/pc/gfx/gfx_pc.c
```diff
@@ -93,11 +96,9 @@ struct LoadedVertex {
...
     struct TextureHashmapNode *next;
-
     const uint8_t *texture_addr;
-    uint8_t fmt, siz;
-
     uint32_t texture_id;
+    uint8_t fmt, siz;
     uint8_t cms, cmt;
     bool linear_filter;
 };
```
PeachyPeachSM64: DynOS also uses this struct in some way, so `struct THN` have to follow the exact same order
please update it as well

PeachyPeachSM64 (changes requested):

src/engine/math_util.inl
theclashingfritz: You're the one who suggested using SSE for the mxtf functions, I didn't even think of it till you did tbh. We don't explicitly enable SSE with the GCC compiler currently. So it won't be faster for all systems.

autogen/lua_constants/deprecated.lua
theclashingfritz: If you say so.

src/pc/gfx/gfx_pc.c
theclashingfritz: Didn't know it did-If so. The structure should be reused anyways and not separated. I'll go ahead and do that.

src/engine/math_util.inl
Isaac0-dev: generally memcpy is the right approach, imo the whole function can just be a memcpy.
the compiler can decide the best option to choose, you're already using -O3. maybe its worth profiling it?

src/pc/lua/smlua_cobject_autogen.c
```diff
@@ -1047,14 +1047,15 @@ static struct LuaObjectField sGlobalTexturesFields[LUA_GLOBAL_TEXTURES_FIELD_COU
...
-#define LUA_GRAPH_NODE_FIELD_COUNT 8
+#define LUA_GRAPH_NODE_FIELD_COUNT 9
 static struct LuaObjectField sGraphNodeFields[LUA_GRAPH_NODE_FIELD_COUNT] = {
     { "children",    LVT_COBJECT_P, offsetof(struct GraphNode, children),    true,  LOT_GRAPHNODE, 1, sizeof(struct GraphNode*) },
     { "extraFlags",  LVT_U8,        offsetof(struct GraphNode, extraFlags),  false, LOT_NONE,      1, sizeof(u8)                },
     { "flags",       LVT_S16,       offsetof(struct GraphNode, flags),       false, LOT_NONE,      1, sizeof(s16)               },
 //  { "georef",      LVT_???,       offsetof(struct GraphNode, georef),      true,  LOT_???,       1, sizeof(const void*)       }, <--- UNIMPLEMENTED
     { "hookProcess", LVT_U8,        offsetof(struct GraphNode, hookProcess), false, LOT_NONE,      1, sizeof(u8)                },
     { "next",        LVT_COBJECT_P, offsetof(struct GraphNode, next),        true,  LOT_GRAPHNODE, 1, sizeof(struct GraphNode*) },
+    { "padding",     LVT_S16,       offsetof(struct GraphNode, padding),     false, LOT_NONE,      1, sizeof(s16)               },
```
Isaac0-dev: no point in exposing this, perhaps black list it?

include/types.h
Isaac0-dev: i doubt it'd invalidate old mods, unless you change the way dynos reads bins, or change the format of one of the smlua_anim_utils functions
Lua doesn't rely on the data structures being a consistent format

src/engine/math_util.inl
theclashingfritz: I looked it up and you can see which optimizations are enabled by default for your GCC using 'gcc -Q --help=target'. They are for me so I don't see any perf difference between the two above. Because GCC auto-optimizes it anyways

src/pc/gfx/gfx_pc.c
PeachyPeachSM64: yes, that would be better actually

src/engine/math_util.inl
Isaac0-dev: for my compiler, its enabled. so memcpy is best.

PeachyPeachSM64: > You're the one who suggested using SSE for the mxtf functions

Yes, but... not for something as simple as a copy

> the compiler can decide the best option to choose, you're already using -O3

The Makefile explicitly sets it to `-O2` if no `DEBUG` or `OPT_FLAGS` are passed

> maybe its worth profile it?

The asm code is the following:
```asm
mtxf_copy_sse(float (*) [4], float (*) [4]):
        movaps  xmm3, XMMWORD PTR [rsi]
        movaps  xmm2, XMMWORD PTR [rsi+16]
        movaps  xmm1, XMMWORD PTR [rsi+32]
        movaps  xmm0, XMMWORD PTR [rsi+48]
        movaps  XMMWORD PTR [rdi], xmm3
        movaps  XMMWORD PTR [rdi+16], xmm2
        movaps  XMMWORD PTR [rdi+32], xmm1
        movaps  XMMWORD PTR [rdi+48], xmm0
        ret
mtxf_copy_memcpy(float (*) [4], float (*) [4]):
        movdqu  xmm0, XMMWORD PTR [rsi]
        movups  XMMWORD PTR [rdi], xmm0
        movdqu  xmm0, XMMWORD PTR [rsi+16]
        movups  XMMWORD PTR [rdi+16], xmm0
        movdqu  xmm0, XMMWORD PTR [rsi+32]
        movups  XMMWORD PTR [rdi+32], xmm0
        movdqu  xmm0, XMMWORD PTR [rsi+48]
        movups  XMMWORD PTR [rdi+48], xmm0
        ret
mtxf_copy_vanilla(float (*) [4], float (*) [4]):
        xor     eax, eax
.L5:
        mov     edx, DWORD PTR [rsi+rax]
        mov     DWORD PTR [rdi+rax], edx
        add     rax, 4
        cmp     rax, 64
        jne     .L5
        ret
```
Under the hood, `memcpy` does basically the same as the SSE implementation (in modern architectures `movaps`, `movups`, `movdqa`, `movdqu` instructions do all the same in practice), but it looks nicer, and the compiler knows what to do with it.
And vanilla implementation is just bad, it shouldn't be kept.

Isaac0-dev: thanks for the asm code, technically `mtxf_copy_sse` has the potential to be faster, but in C its way more confusing and memcpy is actually using inline SSE. it'd hardly make a difference performance wise.

theclashingfritz: Okay. Seeing this-Replacing it with a simple memcpy is better. I'll go ahead and just do that.

That being said. How'd you get the ASM for all three of those so fast? I'd have to disassemble the entire .exe just to grab it.

Isaac0-dev: > The Makefile explicitly sets it to `-O2` if no `DEBUG` or `OPT_FLAGS` are passed

doesn't the `OPTIMIZE_O3` flag in the function definition override that though?

PeachyPeachSM64: [Compiler explorer](https://godbolt.org/)

The code I used here:
```c
typedef float Mat4[4][4];
typedef unsigned int u32;
typedef int s32;

#include <xmmintrin.h>
#include <string.h>

void mtxf_copy_sse(Mat4 dest, Mat4 src) {
   __m128 r1 = _mm_load_ps(src[0]); // Load floats into 128-bits
   __m128 r2 = _mm_load_ps(src[1]);
   __m128 r3 = _mm_load_ps(src[2]);
   __m128 r4 = _mm_load_ps(src[3]);
   _mm_store_ps(dest[0], r1); // Stores 128 bits into a float[4].
   _mm_store_ps(dest[1], r2);
   _mm_store_ps(dest[2], r3);
   _mm_store_ps(dest[3], r4);
}

void mtxf_copy_memcpy(Mat4 dest, Mat4 src) {
    memcpy(dest, src, sizeof(Mat4));
}

void mtxf_copy_vanilla(Mat4 dest, Mat4 src) {
    u32 *d = (u32 *) dest;
    u32 *s = (u32 *) src;
    for (s32 i = 0; i < 16; i++) {
        *d++ = *s++;
    }
}
```
compiler: x86-64 GCC 14.1
flags: -O2
output: Intel asm

theclashingfritz: Thank you. The default implementation gave me this for O3. It's still horrible though. But it does explain why I didn't see to much of a performance difference.

```asm
mtxf_copy_vanilla(float (*) [4], float (*) [4]):
        mov     rdx, rsi
        mov     rcx, rdi
        lea     rsi, [rsi+4]
        sub     rcx, rsi
        cmp     rcx, 8
        jbe     .L5
        movdqu  xmm0, XMMWORD PTR [rdx]
        movups  XMMWORD PTR [rdi], xmm0
        movdqu  xmm0, XMMWORD PTR [rdx+16]
        movups  XMMWORD PTR [rdi+16], xmm0
        movdqu  xmm0, XMMWORD PTR [rdx+32]
        movups  XMMWORD PTR [rdi+32], xmm0
        movdqu  xmm0, XMMWORD PTR [rdx+48]
        movups  XMMWORD PTR [rdi+48], xmm0
        ret
.L5:
        mov     ecx, DWORD PTR [rdx]
        mov     DWORD PTR [rdi], ecx
        mov     ecx, DWORD PTR [rdx+4]
        mov     DWORD PTR [rdi+4], ecx
        mov     ecx, DWORD PTR [rdx+8]
        mov     DWORD PTR [rdi+8], ecx
        mov     ecx, DWORD PTR [rdx+12]
        mov     DWORD PTR [rdi+12], ecx
        mov     ecx, DWORD PTR [rdx+16]
        mov     DWORD PTR [rdi+16], ecx
        mov     ecx, DWORD PTR [rdx+20]
        mov     DWORD PTR [rdi+20], ecx
        mov     ecx, DWORD PTR [rdx+24]
        mov     DWORD PTR [rdi+24], ecx
        mov     ecx, DWORD PTR [rdx+28]
        mov     DWORD PTR [rdi+28], ecx
        mov     ecx, DWORD PTR [rdx+32]
        mov     DWORD PTR [rdi+32], ecx
        mov     ecx, DWORD PTR [rdx+36]
        mov     DWORD PTR [rdi+36], ecx
        mov     ecx, DWORD PTR [rdx+40]
        mov     DWORD PTR [rdi+40], ecx
        mov     ecx, DWORD PTR [rdx+44]
        mov     DWORD PTR [rdi+44], ecx
        mov     ecx, DWORD PTR [rdx+48]
        mov     DWORD PTR [rdi+48], ecx
        mov     ecx, DWORD PTR [rdx+52]
        mov     DWORD PTR [rdi+52], ecx
        mov     ecx, DWORD PTR [rdx+56]
        mov     DWORD PTR [rdi+56], ecx
        mov     edx, DWORD PTR [rdx+60]
        mov     DWORD PTR [rdi+60], edx
        ret
```

PeachyPeachSM64: > doesn't the OPTIMIZE_O3 flag in the function definition override that though?

yes it does

src/engine/math_util.inl
theclashingfritz: Actually, You're right. I'll change it.

src/pc/lua/smlua_cobject_autogen.c
theclashingfritz: Done.

PR #793, opened 2025-05-05
Add HOOK_ON_STAR_DANCE_CUTSCENE

Sunketchupm:
This allows Lua mods to affect the camera movement after a star dance, not the star dance itself. In the clip below, the camera stays in place, as where normally the camera zooms out after collecting the star.
https://github.com/user-attachments/assets/8ad238c9-087a-433c-ae74-eb254f4bad15


JairoThePlumber:
I can't tell the difference with this compare of how it is, but i'm guessing you can make custom star dance functions?

Sunketchupm:
The hook is for the camera movement that happens after star dance. The name does suck since you'd think that it would affect the star dance itself, but the code itself says star dance cutscene.

JairoThePlumber:
I see

Blockyyy:
is there really no way of doing this without the hook? this seems oddly specific

Sunketchupm:
There does seem to be a way to do this without the hook that doesn't involve recreating the entire camera so this hook isn't really needed. I'll close this for now, however I may look into exposing more cutscene stuff later down the line.

AgentXLP:
This hook is too specific

PR #795, opened 2025-05-06
Fix interactions again

PeachyPeachSM64:
- if an interaction is set for a custom action, return that interaction
- no more INT_LUA needed
- helper funcs for gp lands


Isaac0-dev (approved):
code looks solid and i tested it with omm rebirth and sonic character rebooted

PR #797, opened 2025-05-07
Fix packed normals

ManIsCat2:
Fixes a bug where G_TEXTURE_GEN didn't look correct when using G_PACKED_NORMALS_EXT

### DIFF
--- a/src/game/rendering_graph_node.c
+++ b/src/game/rendering_graph_node.c
@@ -70,7 +70,7 @@ Mtx sPrevCamTranf, sCurrCamTranf = {
 };
 
 static Gfx obj_sanitize_gfx[] = {
-    gsSPClearGeometryMode(G_TEXTURE_GEN),
+    gsSPClearGeometryMode(G_TEXTURE_GEN | G_PACKED_NORMALS_EXT),
     gsSPSetGeometryMode(G_LIGHTING),
     gsDPSetCombineMode(G_CC_SHADE, G_CC_SHADE),
     gsSPTexture(0xFFFF, 0xFFFF, 0, 0, G_OFF),
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -842,9 +842,9 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
             float g = rsp.current_lights[rsp.current_num_lights - 1].col[1] * globalLightCached[1][1];
             float b = rsp.current_lights[rsp.current_num_lights - 1].col[2] * globalLightCached[1][2];
 
-            int nx = vn->n[0];
-            int ny = vn->n[1];
-            int nz = vn->n[2];
+            signed char nx = vn->n[0];
+            signed char ny = vn->n[1];
+            signed char nz = vn->n[2];
 
             if (rsp.geometry_mode & G_PACKED_NORMALS_EXT) {
                 unsigned short packedNormal = vn->flag;
@@ -853,10 +853,10 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
 
                 nx = xo & 0x7F;
                 ny = yo & 0x7F;
-                nz = nx + ny;
+                nz = (nx + ny) ^ 0x7F;
+
                 int x2 = nx ^ 0x7F;
                 int y2 = ny ^ 0x7F;
-                nz = nz ^ 0x7F;
 
                 if (nz & 0x80) {
                     nx = x2;
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -855,17 +855,14 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
                 ny = yo & 0x7F;
                 nz = (nx + ny) ^ 0x7F;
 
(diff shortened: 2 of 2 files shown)


AgentXLP:
this is actually a better way of going about it, nice

AgentXLP:
:worried: I don't want to "ninja merge" but I literally do not see anything wrong with this code at all

src/pc/gfx/gfx_pc.c
```diff
@@ -842,38 +842,39 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
             float g = rsp.current_lights[rsp.current_num_lights - 1].col[1] * globalLightCached[1][1];
             float b = rsp.current_lights[rsp.current_num_lights - 1].col[2] * globalLightCached[1][2];
 
+            int nx = vn->n[0];
+            int ny = vn->n[1];
+            int nz = vn->n[2];
```
Isaac0-dev: make these be `s8`, since that's what `vn->n` is

src/pc/gfx/gfx_pc.c
```diff
@@ -842,38 +842,39 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
...
+                unsigned short packedNormal = vn->flag;
+                int xo = packedNormal >> 8;
+                int yo = packedNormal & 0xFF;
+
+                nx = xo & 0x7F;
+                ny = yo & 0x7F;
+                nz = nx + ny;
+                int x2 = nx ^ 0x7F;
+                int y2 = ny ^ 0x7F;
+                nz = nz ^ 0x7F;
```
Isaac0-dev: combine this into the calculation above, so
`nz = (nx + ny) ^ 0x7F;`

src/game/rendering_graph_node.c
```diff
@@ -70,7 +70,7 @@ Mtx sPrevCamTranf, sCurrCamTranf = {
 };
 
 static Gfx obj_sanitize_gfx[] = {
-    gsSPClearGeometryMode(G_TEXTURE_GEN),
+    gsSPClearGeometryMode(G_TEXTURE_GEN | G_PACKED_NORMALS_EXT),
```
Isaac0-dev, line 73: its good to do this, but what happens without it?

ManIsCat2: if a model uses it and doesnt clear it, all the displaylists that are being drawn will use it. and that doesnt look good

src/pc/gfx/gfx_pc.c
ManIsCat2: done

src/game/rendering_graph_node.c
Isaac0-dev: Ah, but can't the model manually disable the mode?
Either way it doesn't matter and this should absolutely stay in the pr I'm just interested.

ManIsCat2: oh alright im keeping it then

src/pc/gfx/gfx_pc.c
```diff
@@ -842,39 +842,40 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
...
+                int y2 = ny ^ 0x7F;
+
+                if (nz & 0x80) {
+                    nx = x2;
+                    ny = y2;
                 }
 
+                nx = (xo & 0x80) ? -nx : nx;
+                ny = (yo & 0x80) ? -ny : ny;
+                nz = (nz & 0x80) ? (nz - 0x100) : nz;
```
PeachyPeachSM64: this is no longer needed since nz is already s8
edit: I'm talking about line 868 only, not the two above

src/pc/gfx/gfx_pc.c
```diff
@@ -842,39 +842,40 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
...
+                ny = yo & 0x7F;
+                nz = (nx + ny) ^ 0x7F;
+
+                int x2 = nx ^ 0x7F;
+                int y2 = ny ^ 0x7F;
+
+                if (nz & 0x80) {
+                    nx = x2;
+                    ny = y2;
                 }
```
PeachyPeachSM64: could be:
```c
if (nz & 0x80) {
    nx ^= 0x7F;
    ny ^= 0x7F;
}
```

PeachyPeachSM64 (changes requested):

PeachyPeachSM64:
something strange I noticed with packed normals is that they describe an octahedron instead of a sphere, meaning they always have a magnitude below 127
will not that result in an intensity lower than expected? is it by design or due to some kind of limitation?

src/pc/gfx/gfx_pc.c
ManIsCat2: thankyou

ManIsCat2:
> something strange I noticed with packed normals is that they describe an octahedron instead of a sphere, meaning they always have a magnitude below 127 will not that result in an intensity lower than expected? is it by design or due to some kind of limitation?

not sure 🙏 , this is just the original f3dex3 implementation translated into C so people can use it

PeachyPeachSM64 (approved):

thegreatmario64:
> Fixes a bug where G_TEXTURE_GEN didn't look correct when using G_PACKED_NORMALS_EXT

O: That explains it

PR #812, opened 2025-05-17
Add HOOK_MARIO_OVERRIDE_FLOOR_CLASS

NitroDisPro:
Adds a `HOOK_MARIO_OVERRIDE_FLOOR_CLASS` hook event to allow changing the surface class for Mario.

Couple basic implementations:
```lua
local function slip_every_floor(m)
    return SURFACE_CLASS_VERY_SLIPPERY --All surfaces are now slippery
end

hook_event(HOOK_MARIO_OVERRIDE_FLOOR_CLASS, override_surf_class)
```

```lua
local function no_slip_floor(m, surfaceClass)
	if m.action == ACT_CRAWLING and surfaceClass == SURFACE_CLASS_VERY_SLIPPERY then
	return SURFACE_CLASS_NOT_SLIPPERY --Change very slippery surfaces to non-slippery only when crawling
	end
end

hook_event(HOOK_MARIO_OVERRIDE_FLOOR_CLASS, no_slip_floor)
```


AgentXLP:
.... did you manually edit the autogen files using the online github code editor?
That makes me feel like you didn't test this either, idk if I'm going to merge this

PeachyPeachSM64:
no, there are some problems with this:
`s32 returnValue;` is never initialized, meaning if the callback doesn't return a value it will be garbage data and return an invalid surface class
also `SURFACE_CLASS_DEFAULT` has the value 0, so you won't never be able to override the surface class type with this value

NitroDisPro:
Fixed it I think, I did test it before but very little. This should hopefully work now.

PeachyPeachSM64 (approved):

PR #813, opened 2025-05-17
Improving math functions

PeachyPeachSM64:
- Autogen vector functions for different vector types by using a template
- Remove `smlua_math_utils` with poor performance, add `clamp`, `sqr` and `hypot` to built-in Lua math lib
- Add `set_magnitude`, `hdist`, `is_zero` to vector functions
- Add `vec3f_rotate_around_n` and `vec3f_transform` to `Vec3f`
- `not_zero` -> `replace_value_if_not_zero`
- Universal `clamp` macro


Isaac0-dev (commented):
Not a full review but use the local math functions to make your new ones perform faster

autogen/convert_structs.py
```diff
@@ -70,9 +70,6 @@
 #endif
 """
 
-override_field_names = {
```
Isaac0-dev, line 73: good move :+1:

autogen/lua_constants/built-in.lua
```diff
@@ -135,6 +135,37 @@ function network_player_get_override_palette_color(np, part)
...
+    return x * x
+end
+
+--- @param x number
+--- @param a number
+--- @param b number
+--- @return number
+--- Clamps the number `x` between bounds `a` (minimum) and `b` (maximum)
+function math.clamp(x, a, b)
+    return math.min(math.max(x, a), b)
```
Isaac0-dev: as i mentioned before, use a local version of max and min here

autogen/lua_constants/built-in.lua
```diff
@@ -135,6 +135,37 @@ function network_player_get_override_palette_color(np, part)
...
+end
+
+--- @param a number
+--- @param b number
+--- @return number
+--- Computes the hypotenuse of a right-angled triangle given sides `a` and `b` using the Pythagorean theorem
+function math.hypot(a, b)
+    return math.sqrt(a * a + b * b)
```
Isaac0-dev: same here for `math.sqrt`

src/game/rendering_graph_node.c
```diff
@@ -566,7 +566,7 @@ static void geo_process_camera(struct GraphNodeCamera *node) {
     if (node->fnNode.func != NULL) {
         node->fnNode.func(GEO_CONTEXT_RENDER, &node->fnNode.node, gMatStack[gMatStackIndex]);
     }
-    mtxf_rotate_xy(rollMtx, node->rollScreen);
+    mtxf_rotate_xy(rollMtx->m, node->rollScreen);
```
Isaac0-dev, line 569: what difference does this make? curious

autogen/gen_math.py
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+    }
+}
+
+
+def main():
+    verbose = len(sys.argv) > 1 and (sys.argv[1] == "-v" or sys.argv[1] == "--verbose")
+    for template_file, x in TEMPLATES.items():
+        tmpl = open(template_file).read()
```
Isaac0-dev:
```py
with open(template_file) as f:
    tmpl = f.read()
```

autogen/gen_math.py
```diff
@@ -0,0 +1,95 @@
...
+            "args": {
+                "size": 3
+            }
+        }
+    }
+}
+
+
+def main():
+    verbose = len(sys.argv) > 1 and (sys.argv[1] == "-v" or sys.argv[1] == "--verbose")
```
Isaac0-dev, line 71: `verbose` doesn't seem to be used

Isaac0-dev (changes requested):
thanks git for the messy diff

i like the template idea, i wish there was an easy way to just have automatic macros like our min/max/sqr/absx ones we have currently

Isaac0-dev:
also another thought. `autogen.sh` is starting to accumulate scripts that aren't related to smlua. I wonder if we should have a separate script, or an optional cli flag to run scripts that might not change as often as smlua does. These templates aren't likely to change too often, same with the display list one. 
Admittedly it doesn't waste much time when running the script, at least on linux.

autogen/lua_constants/built-in.lua
PeachyPeachSM64: I'll take a look

src/game/rendering_graph_node.c
PeachyPeachSM64: function signature has changed: Mtx -> Mat4
I'm planning to replace all instances of Mtx in favor of Mat4 in a future PR, since it's almost the same type but Mtx is not usable in Lua

autogen/gen_math.py
PeachyPeachSM64: why? that makes no difference

autogen/gen_math.py
PeachyPeachSM64: yeah, I was planning to print some stuff, but there is nothing really useful to print

autogen/gen_math.py
Isaac0-dev: because without it, you're not closing the file and just leaving it to python to clean up after you.
its just good programmer practice. if you want to ignore it and have python clean up, thats fine i guess.

PeachyPeachSM64: you're right, I'll change it

theclashingfritz: I agree with Isaac here. I hate when programs open a file and leave it locked because they refuse to release the file even after they don't need it. Isaacs example will have it automatically release when we're done using it.

theclashingfritz:
> also another thought. `autogen.sh` is starting to accumulate scripts that aren't related to smlua. I wonder if we should have a separate script, or an optional cli flag to run scripts that might not change as often as smlua does. These templates aren't likely to change too often, same with the display list one. Admittedly it doesn't waste much time when running the script, at least on linux.

We should probably separate autogen into parts. Lua included. That way we don't always need a 'full' autogen for everything.

Isaac0-dev (approved):
i definitely have wrestled with the diff in `math_util.c` to understand what has changed and it's likely that i missed something.
but I approve of what i understood :+1:

PR #815, opened 2025-05-18
Add gLevelValues.wallMaxRadius

Sunketchupm:
Created to be able to do this much easier
https://github.com/user-attachments/assets/ef38ec93-a283-4b4f-b1a2-326c149181ac

### DIFF
--- a/src/game/hardcoded.h
+++ b/src/game/hardcoded.h
@@ -262,6 +262,7 @@ struct BehaviorValues {
     u8  MultipleCapCollection;
     u8  InfiniteRenderDistance;
     s16 CourtyardBoosRequirement;
+    f32 WallMaxRadius;
     struct StarsNeededForDialog starsNeededForDialog;
     struct BehaviorDialogs dialogs;
     struct BehaviorTrajectories trajectories;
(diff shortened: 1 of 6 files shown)


PeachyPeachSM64:
imo it belongs more to gLevelValues than gBehaviorValues

AgentXLP:
I agree

src/game/hardcoded.h
```diff
@@ -94,6 +94,7 @@ struct LevelValues {
     u8 useGlobalStarIds;
     u8 zoomOutCameraOnPause;
     u8 jrbDarkenSkybox;
+    f32 WallMaxRadius;
```
Isaac0-dev: this is inconsistent field naming, do `wallMaxRadius`

Isaac0-dev (changes requested):

src/game/hardcoded.h
Sunketchupm: Was a leftover from gBehaviorValues, oops. Although, it does make you wonder why gBehaviorValues mostly has fields named this way.

AgentXLP: yeah I don't know why tbh

PR #816, opened 2025-05-18
Fix hand/foot pos and get any Mario part position

PeachyPeachSM64:
Title

Since these values are updated during Mario's graph node processing, there is a 1-frame delay if they are retrieved before rendering (for example during an object update).
But there is a simple solution to that problem: retrieve these values right after Mario's graph node processing, thanks to `HOOK_ON_GEO_PROCESS`:
```lua
--- @param m MarioState
local function hook_mario_graph_node_processing(m)
    m.marioObj.header.gfx.sharedChild.hookProcess = 1
end

hook_event(HOOK_MARIO_UPDATE, hook_mario_graph_node_processing)

local function on_geo_process(node, matStackIndex)
    local buttPos = { x = 0, y = 0, z = 0 }
    if get_mario_anim_part_pos(geo_get_mario_state(), MARIO_ANIM_PART_BUTT, buttPos) then

        ... do whatever you want with this ...

    end
end

hook_event(HOOK_ON_GEO_PROCESS, on_geo_process)
```
Another solution using `HOOK_ON_OBJECT_RENDER` and hooking the object directly (thanks Arena hammer code):
```lua
-- Update function of a custom behavior
local function some_object_update(o)
    o.hookRender = 1
end

local function on_object_render(o)
    local buttPos = { x = 0, y = 0, z = 0 }
    if get_mario_anim_part_pos(geo_get_mario_state(), MARIO_ANIM_PART_BUTT, buttPos) then
        o.header.gfx.pos.x = buttPos.x
        o.header.gfx.pos.y = buttPos.y
        o.header.gfx.pos.z = buttPos.z
    end
end

hook_event(HOOK_ON_OBJECT_RENDER, on_object_render)
```


Baconator2558:
Wouldn't it make a bit more sense to keep parts 0-3 as the hands and feet?

Isaac0-dev (approved):

src/game/rendering_graph_node.c
```diff
@@ -941,6 +942,18 @@ static void geo_process_animated_part(struct GraphNodeAnimatedPart *node) {
     // Increment the matrix stack, If we fail to do so. Just return.
     if (!increment_mat_stack()) { return; }
 
+    // Mario anim part pos
+    if (gCurMarioBodyState && !gCurGraphNodeHeldObject) {
+        gCurMarioBodyState->currAnimPart++;
```
Isaac0-dev: previously i assumed you were incrementing `currAnimPart` at the start of the function to avoid having issues if one of the above conditions returns from the function. If that actually happens it'd be less than ideal, but I thought it was a good idea.

PeachyPeachSM64: how did I miss that

PeachyPeachSM64: ah yes, it makes checking `gCurMarioBodyState && !gCurGraphNodeHeldObject` twice, but that hardly matters

Isaac0-dev:
so does this mean that head/torso pos and angle vectors are probably a frame off?

PeachyPeachSM64:
> so does this mean that head/torso pos and angle vectors are probably a frame off?

yes, since it's computed during graph node processing, which happens after object update
only ways to have it synced with the frame rendering is getting the values during HOOK_ON_OBJECT_RENDER and HOOK_ON_GEO_PROCESS

Baconator2558:
Just to clarify, the hand/foot values were also offset in the same way with the previous method, right? Since they were also done in a graph node function.

PeachyPeachSM64:
> Just to clarify, the hand/foot values were also offset in the same way with the previous method, right? Since they were also done in a graph node function.

yes, but you couldn't retrieve the left foot position with the old method (and during some actions you couldn't retrieve anything)

PR #817, opened 2025-05-18
Improve fixed collision

Sunketchupm:
All fixed collision effects have been separated.
The fixed collision structuring has been improved and a compatibility layer was added to support the now-old structure.
Added `fixWallOnSlope`, which prevents walls from protruding out onto slopes. This patch was made by FramePerfection, who gave me permission to use it. It's a direct copy from his patch, so optimizations are possible.


src/engine/surface_collision.c
```diff
@@ -123,6 +123,14 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
     register f32 w1, w2, w3;
     register f32 y1, y2, y3;
+    register f32 v0x, v0y, v0z;
+	register f32 v1x, v1y, v1z;
+	register f32 v2x, v2y, v2z;
+	register f32 d00, d01, d11, d20, d21;
+	register f32 invDenom;
+	register f32 v, w;
+	register f32 margin_radius = radius - 1.0f;
+    const f32 corner_threshold = -0.9f;
```
PeachyPeachSM64: indenting is wrong
also please remove `register` keywords, it's not N64 anymore

src/engine/surface_collision.c
```diff
@@ -283,22 +293,138 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+            //Face
+            d00 = v0x * v0x + v0y * v0y + v0z * v0z;
+            d01 = v0x * v1x + v0y * v1y + v0z * v1z;
+            d11 = v1x * v1x + v1y * v1y + v1z * v1z;
+            d20 = v2x * v0x + v2y * v0y + v2z * v0z;
+            d21 = v2x * v1x + v2y * v1y + v2z * v1z;
+            invDenom = 1.0f / (d00 * d11 - d01 * d01);
+            v = (d11 * d20 - d01 * d21) * invDenom;
+            if (v < 0.0f || v > 1.0f)
+                goto edge_1_2;
```
PeachyPeachSM64: please no goto programming
SM64 code is already a mess, no need to make it harder to read for no reason
if you can't understand the code you submit, just don't PR it, thanks

src/engine/surface_collision.c
```diff
@@ -309,6 +435,11 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
         numCols++;
     }
 
+    if (gLevelValues.fixCollision.fixWallOnSlope) {
+        data->x = x;
+	    data->z = z;
```
PeachyPeachSM64: indenting is also wrong here

src/engine/surface_collision.c
```diff
@@ -283,22 +293,138 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
             data->z = cPos[2] + cNorm[2] * radius;
             x = data->x;
             z = data->z;
             data->normalAddition[0] += cNorm[0];
             data->normalAddition[2] += cNorm[2];
             data->normalCount++;
-        } else {
-            data->x += surf->normal.x * (radius - offset);
-            data->z += surf->normal.z * (radius - offset);
+            } else {
```
PeachyPeachSM64: same here

src/game/hardcoded.h
```diff
@@ -41,12 +41,21 @@ struct StarPositions {
...
+    u8 disableFalseLedgeGrab;
+    u8 disableGroundPoundBonks;
+    u8 pickBestWall;
+    u8 fixExposedCeilings;
+    u8 fixFloorOvershadowing;
+    u8 fixOnlyFirstWall;
+    u8 fixObjectOobPush;
+    u8 fixWallOnSlope;
+    u8 throwaway; // Used for legacy fixCollisionBugs support
+};
```
PeachyPeachSM64: booleans exist for a reason
I know this is too late to change that without losing consistency, but having to do `if roundedCorners == 1` in Lua because that stupid language doesn't treat 0 as false is slightly annoying and this alone created so much unnecessary bugs

src/pc/djui/djui_panel_misc.c
```diff
@@ -18,7 +18,7 @@ void djui_panel_options_debug_create(struct DjuiBase* caller) {
     {
         djui_checkbox_create(body, DLANG(MISC, LUA_PROFILER), &configLuaProfiler, NULL);
 #ifdef DEVELOPMENT
-        djui_checkbox_create(body, DLANG(MISC, FIXED_COLLISIONS), (bool*)&gLevelValues.fixCollisionBugs, NULL);
+        //djui_checkbox_create(body, DLANG(MISC, FIXED_COLLISIONS), (bool*)&gLevelValues.fixCollisionBugs, NULL);
```
PeachyPeachSM64: please add a replacement for this, it is an important feature for development

src/pc/lua/smlua_cobject.c
```diff
@@ -528,11 +528,22 @@ static int smlua__get_field(lua_State* L) {
         return 1;
     }
 
-    const char *key = lua_tostring(L, 2);
-    if (!key) {
+    // This warns due to not being a const
+    char *tempKey = lua_tostring(L, 2);
```
PeachyPeachSM64: then define it `const`? also why does this need to be a separate variable?

src/pc/lua/smlua_cobject.c
```diff
@@ -528,11 +528,22 @@ static int smlua__get_field(lua_State* L) {
...
+    // Legacy fixCollisionBugs support
+    if (lot == (enum LuaObjectType)LOT_LEVELVALUES && strncmp(tempKey, "fixCollisionBugs", 16) == 0) {
+        lot = (enum LuaObjectType)LOT_FIXCOLLISIONBUGS;
+        if (strcmp(tempKey, "fixCollisionBugs") == 0) { tempKey = "throwaway"; }
+        else if (strcmp(tempKey, "fixCollisionBugsRoundedCorners") == 0) { tempKey = "roundedCorners"; }
+        else if (strcmp(tempKey, "fixCollisionBugsFalseLedgeGrab") == 0) { tempKey = "disableFalseLedgeGrab"; }
+        else if (strcmp(tempKey, "fixCollisionBugsGroundPoundBonks") == 0) { tempKey = "disableGroundPoundBonks"; }
+        else if (strcmp(tempKey, "fixCollisionBugsPickBestWall") == 0) { tempKey = "pickBestWall"; }
+    }
+    const char *key = tempKey;
```
PeachyPeachSM64: that's really a hacky fix
try to see if this can be resolved with an union
autogen can map two different names to the same address
see `Vtx_L` for example

src/pc/lua/smlua_cobject.c
```diff
@@ -614,11 +625,29 @@ static int smlua__set_field(lua_State* L) {
...
+                // Not ideal since several mods do disable some of the options
+                fix_collision_bugs_set_all(isEnabled);
+            }
+        }
+        else if (strcmp(tempKey, "fixCollisionBugsRoundedCorners") == 0) { tempKey = "roundedCorners"; }
+        else if (strcmp(tempKey, "fixCollisionBugsFalseLedgeGrab") == 0) { tempKey = "disableFalseLedgeGrab"; }
+        else if (strcmp(tempKey, "fixCollisionBugsGroundPoundBonks") == 0) { tempKey = "disableGroundPoundBonks"; }
+        else if (strcmp(tempKey, "fixCollisionBugsPickBestWall") == 0) { tempKey = "pickBestWall"; }
+    }
+    const char *key = tempKey;
```
PeachyPeachSM64: same here as above

PeachyPeachSM64 (changes requested):

src/engine/surface_collision.c
theclashingfritz: This looks really readable to me despite it being goto code. I suppose you could separate it out into functions though. As for the messy code. If you ask me. Audio code is a good example of how to do inlines and gotos wrong. It's a bloody mess.

src/game/hardcoded.h
theclashingfritz: I think fixing LUA at this point is a better decision. I agree about bools tbh and we should through the codebase and correct it. But that being said. The real issue stems from LUA and it's horrible syntax.

src/pc/djui/djui_panel_misc.c
theclashingfritz: Agreed. I added it for that purpose originally.

src/pc/lua/smlua_cobject.c
theclashingfritz: The variable is reused later which is why it's separate-But agreed, Define it as const or at the very least cast it.

src/pc/lua/smlua_cobject.c
theclashingfritz: I don't know if a union is possible? I mean most of these toggles can be compressed into bitflags anyways actually now that I think about it.

src/engine/surface_collision.c
Sunketchupm: I'm also not able to replicate the behavior correctly without the use of gotos. I don't know if there is something I don't understand with gotos or if my implementations are wrong.

PeachyPeachSM64: goto programming in C is a sin
I agree, these checks should be in separate functions
also, a bit of explanation of what it does would be useful
they look like the formulas used to check if a point lies within a triangle, and if a point can be projected on an edge, but I'm not sure

src/pc/lua/smlua_cobject.c
PeachyPeachSM64: something like this could work:
```c
struct LevelValues {
    union {
        struct FixCollisionBugs fixCollision;
        struct {
            u8 fixCollisionBugsRoundedCorners;
            u8 fixCollisionBugsFalseLedgeGrab;
            u8 fixCollisionBugsGroundPoundBonks;
            u8 fixCollisionBugsPickBestWall;
        };
    };
    u8 fixVanishFloors;
    ...
```
but that wouldn't work for `fixCollisionBugs`
now that I think about it, maybe define a metatable for `gLevelValues` in Lua instead

Sunketchupm:
The only thing I need to figure out now is making the compatibility layer with fixCollisionBugs less hacky

src/pc/lua/smlua_cobject.c
Isaac0-dev: i'm not a fan of using a metatable for that, especially if you define it in Lua (you should never do that).
because of `fixCollisionBugs`, the union method may not be that great
i agree that the system that's already here is hacky, but rather than translating prompt keys, i think this may be a viable approach, perhaps put it inside a new function dedicated to deprecated code that has to be handled this way:

```C
    if (lot == LOT_LEVELVALUES && !strncmp(key, "fixCollisionBugs", 16)) {
        if (!strcmp(key, "fixCollisionBugs")) {
            lua_pushinteger(L, gLevelValues.fixCollision.roundedCorners ||
                               gLevelValues.fixCollision.disableFalseLedgeGrab ||
                               gLevelValues.fixCollision.disableGroundPoundBonks ||
                               gLevelValues.fixCollision.pickBestWall);
        }
        if (!strcmp(key, "fixCollisionBugsRoundedCorners"))          { lua_pushinteger(L, gLevelValues.fixCollision.roundedCorners); }
        if (!strcmp(key, "fixCollisionBugsDisableFalseLedgeGrab"))   { lua_pushinteger(L, gLevelValues.fixCollision.disableFalseLedgeGrab); }
        if (!strcmp(key, "fixCollisionBugsDisableGroundPoundBonks")) { lua_pushinteger(L, gLevelValues.fixCollision.disableGroundPoundBonks); }
        if (!strcmp(key, "fixCollisionBugsPickBestWall"))            { lua_pushinteger(L, gLevelValues.fixCollision.pickBestWall); }
        return 1;
    }



    if (lot == LOT_LEVELVALUES && !strncmp(key, "fixCollisionBugs", 16)) {
        int value = smlua_to_integer(L, 3);
        if (gSmLuaConvertSuccess) {
            if (!strcmp(key, "fixCollisionBugs")) { fix_collision_bugs_set_all(value); }
            if (!strcmp(key, "fixCollisionBugsRoundedCorners"))          { gLevelValues.fixCollision.roundedCorners = value; }
            if (!strcmp(key, "fixCollisionBugsDisableFalseLedgeGrab"))   { gLevelValues.fixCollision.disableFalseLedgeGrab = value; }
            if (!strcmp(key, "fixCollisionBugsDisableGroundPoundBonks")) { gLevelValues.fixCollision.disableGroundPoundBonks = value; }
            if (!strcmp(key, "fixCollisionBugsPickBestWall"))            { gLevelValues.fixCollision.pickBestWall = value; }
            return 1;
        }
        LOG_LUA_LINE("_set_field failed to retrieve value type '%d', key '%s'", data->valueType, key);
        return 0;
    }
```

not tested or anything, its psuedocode

Sunketchupm: I'm also able to just add the old `fixCollisionBugs` fields back in, though I don't think that's desirable

Isaac0-dev: but they wouldn't directly effect the actual fields that are checked

Sunketchupm: The undesirable part would be checking the old fields alongside the new fields. I could revert the new fields and just use the old system, but that was a pretty badly made system.

Isaac0-dev: if you moved `throwaway` to the start of the  `FixCollisionBugs` struct, you could use the union method peachy suggested ~~without any drawbacks that i can see~~
but you'd still need a way to have the `throwaway` field actually affect the other fields

Sunketchupm: Is there a way to make a field invisible to lua but not excluded entirely from autogen? `override_field_invisible` seems to do both of them which causes an error.

Isaac0-dev: there's no existing list for that, you'd need to implement it
assuming you mean making fields that are accepted by the engine, but not in the docs

Sunketchupm:
I used Isaac's suggestion for this refactor and made it somewhat friendly for potential future deprecated fields.

src/engine/surface_collision.c
theclashingfritz: Goto programming isn't a sin. It's just old practice but it is hard to maintain for newer coders because you need to be able to understand and read code flows. But it isn't recommended if it can be helped.

src/engine/surface_collision.c
```diff
@@ -102,6 +102,36 @@ void closest_point_to_triangle(struct Surface* surf, Vec3f src, Vec3f out) {
     out[2] = v1[2] + s * edge0[2] + t * edge1[2];
 }
 
+u8 is_point_past_facing_edge(struct Surface* surf, Vec3f posRelToEdge, Vec3f edgePos, f32* returnX, f32* returnZ, f32* margin_radius) {
```
Isaac0-dev: this function can be static, and make it return a boolean, not an integer.
also, add a check if `edgePos[1]` is 0. you already do that with all calls for the function, but move them into the function itself (return true if `edgePos == 0`)

the function name is misleading because it does more than just return whether the point is past an edge. it also updates `returnX` and `returnZ`.
for example, `adjust_point_at_surface_edge` could work

src/engine/surface_collision.c
```diff
@@ -281,20 +326,104 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+            posRelToEdge[0] = x - (f32)surf->vertex1[0];
+            posRelToEdge[1] = y - (f32)surf->vertex1[1];
+            posRelToEdge[2] = z - (f32)surf->vertex1[2];
+
+            // Face (Initial check)
+            dotEdge12 = edge12[0] * edge12[0] + edge12[1] * edge12[1] + edge12[2] * edge12[2];
+            dotEdge12_13 = edge12[0] * edge13[0] + edge12[1] * edge13[1] + edge12[2] * edge13[2];
+            dotEdge13 = edge13[0] * edge13[0] + edge13[1] * edge13[1] + edge13[2] * edge13[2];
+            dotPosWithEdge12 = posRelToEdge[0] * edge12[0] + posRelToEdge[1] * edge12[1] + posRelToEdge[2] * edge12[2];
+            dotPosWithEdge13 = posRelToEdge[0] * edge13[0] + posRelToEdge[1] * edge13[1] + posRelToEdge[2] * edge13[2];
```
Isaac0-dev: this is calculating the same thing more than once, please dedicate a variable for that
there's 3 times this happens in this block that i can see.

src/engine/surface_collision.c
```diff
@@ -281,20 +326,104 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+            invDenom = 1.0f / (dotEdge12 * dotEdge13 - dotEdge12_13 * dotEdge12_13);
+            vertex2Weight = (dotEdge13 * dotPosWithEdge12 - dotEdge12_13 * dotPosWithEdge13) * invDenom;
+            // Is the position outside the 12 edge?
+            if (vertex2Weight < 0.0f || vertex2Weight > 1.0f) {
+                // do nothing
+            } else {
+                vertex3Weight = (dotEdge12 * dotPosWithEdge13 - dotEdge12_13 * dotPosWithEdge12) * invDenom;
+                // Is the position outside the 13 edge?
+                if (vertex3Weight < 0.0f || vertex3Weight > 1.0f || vertex2Weight + vertex3Weight > 1.0f) {
+                    // do nothing
```
Isaac0-dev: remove these empty if conditions and at least do
```C
if (!(vertex3Weight < 0.0f || vertex3Weight > 1.0f || vertex2Weight + vertex3Weight > 1.0f)) {
    x += surf->normal.x * (radius - offset);
    z += surf->normal.z * (radius - offset);
    hasCollision = TRUE;
}
```

src/engine/surface_collision.c
```diff
@@ -281,20 +326,104 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+                edge23[1] = (f32)(surf->vertex3[1] - surf->vertex2[1]);
+                edge23[2] = (f32)(surf->vertex3[2] - surf->vertex2[2]);
+
+                posRelToEdge[0] = x - (f32)surf->vertex2[0];
+                posRelToEdge[1] = y - (f32)surf->vertex2[1];
+                posRelToEdge[2] = z - (f32)surf->vertex2[2];
+
+                if (edge23[1] != 0.0f) {
+                    if (!is_point_past_facing_edge(surf, posRelToEdge, edge23, &x, &z, &margin_radius)) {
+                        // do nothing (Collision found)
```
Isaac0-dev: same here

src/engine/surface_collision.c
```diff
@@ -281,20 +326,104 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+                    }
+                } else {
+                    continue;
+                }
+            }
+        }
+
         //! (Wall Overlaps) Because this doesn't update the x and z local variables,
         //  multiple walls can push mario more than is required.
         //  <Fixed when gLevelValues.fixCollisionBugs != 0>
```
Isaac0-dev: these comments need updating now because you've given each feature it's own field
- 4 in surface_collision.c
- 1 in surface_load.c
- 1 in interaction.c

src/engine/surface_collision.c
```diff
@@ -131,6 +161,17 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
 
+    // Used for fixWallOnSlope
+    Vec3f edge12;
+    Vec3f edge13;
+    Vec3f edge23;
+    Vec3f posRelToEdge;
+    f32 dotEdge12, dotEdge12_13, dotEdge13, dotPosWithEdge12, dotPosWithEdge13;
+    f32 invDenom;
+    f32 vertex2Weight, vertex3Weight;
+    f32 margin_radius = radius - 1.0f;
```
Isaac0-dev: make this `marginRadius`

src/pc/lua/smlua_cobject.c
```diff
@@ -470,6 +470,56 @@ static bool smlua_set_field(lua_State* L, u8* p, struct LuaObjectField *data) {
...
+
+
+static int smlua_set_field_deprecated(lua_State* L, enum LuaObjectType lot, const char* key, UNUSED u64 pointer, struct LuaObjectField* data) {
+    // Legacy LevelValues support
+    if (lot == (enum LuaObjectType)LOT_LEVELVALUES) {
+        // Legacy fixCollisionBugs support
+        if (!strncmp(key, "fixCollisionBugs", 16)) {
+            int value = smlua_to_integer(L, 3);
+            if (gSmLuaConvertSuccess) {
+                if (!strcmp(key, "fixCollisionBugs")) { fix_collision_bugs_set_all(value); }
```
Isaac0-dev, line 510: what if a mod does
```lua
gLevelValues.fixCollisionBugsRoundedCorners = 0
gLevelValues.fixCollisionBugs = 1
```
suddenly, that logic is broken. original behavior would have enabled everything about fixed collisions except for rounded corners.
but now it will enable everything with fix collision bugs.

src/pc/lua/smlua_cobject.c
```diff
@@ -470,6 +470,56 @@ static bool smlua_set_field(lua_State* L, u8* p, struct LuaObjectField *data) {
...
+        }
+    }
+    return -1;
+}
+
+
+static int smlua_set_field_deprecated(lua_State* L, enum LuaObjectType lot, const char* key, UNUSED u64 pointer, struct LuaObjectField* data) {
+    // Legacy LevelValues support
+    if (lot == (enum LuaObjectType)LOT_LEVELVALUES) {
+        // Legacy fixCollisionBugs support
```
Isaac0-dev, line 504: some extreme nitpicks about this code style:
- put a space after a cast, so `(enum LuaObjectType) LOT_LEVELVALUES`
- remove the  double return between `smlua_get_field_deprecated` and `smlua_set_field_deprecated`
- those comments look messy. you can just remove them and put a comment before both functions "Legacy field support", or something like it.

src/engine/surface_collision.c
```diff
@@ -141,7 +182,10 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
             continue;
         }
 
-        if (gLevelValues.fixCollisionBugs && gLevelValues.fixCollisionBugsRoundedCorners && !gFindWallDirectionAirborne) {
+        // Used with fixWallOnSlope so it will be calculated regardless of roundedCorners
+        offset = surf->normal.x * x + surf->normal.y * y + surf->normal.z * z + surf->originOffset;
```
Isaac0-dev: check it then!
you can literally just do `if (!gLevelValues.fixCollision.roundedCorners || gFindWallDirectionAirborne || gLevelValues.fixCollision.fixWallOnSlope)`
and it's worth it because collision checks can be expensive

src/engine/surface_collision.c
```diff
@@ -184,66 +228,67 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
 
         } else {
 
-            offset = surf->normal.x * x + surf->normal.y * y + surf->normal.z * z + surf->originOffset;
-
-            if (offset < -radius || offset > radius) {
+            // Used for walls near the edge of the slope, which roundedCorner handles (though less well)
+            if (offset < (gLevelValues.fixCollision.fixWallOnSlope ? 0 : -radius) || offset > radius) {
```
Isaac0-dev: offset is a float, make that 0 be 0.f

src/engine/surface_collision.c
```diff
@@ -281,20 +326,104 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+            posRelToEdge[2] = z - (f32)surf->vertex1[2];
+
+            // Face (Initial check)
+            dotEdge12 = edge12[0] * edge12[0] + edge12[1] * edge12[1] + edge12[2] * edge12[2];
+            dotEdge12_13 = edge12[0] * edge13[0] + edge12[1] * edge13[1] + edge12[2] * edge13[2];
+            dotEdge13 = edge13[0] * edge13[0] + edge13[1] * edge13[1] + edge13[2] * edge13[2];
+            dotPosWithEdge12 = posRelToEdge[0] * edge12[0] + posRelToEdge[1] * edge12[1] + posRelToEdge[2] * edge12[2];
+            dotPosWithEdge13 = posRelToEdge[0] * edge13[0] + posRelToEdge[1] * edge13[1] + posRelToEdge[2] * edge13[2];
+
+            invDenom = 1.0f / (dotEdge12 * dotEdge13 - dotEdge12_13 * dotEdge12_13);
```
Isaac0-dev: a division by 0 might be possible here, so add a check in case

src/engine/surface_collision.c
```diff
@@ -281,20 +326,104 @@ static s32 find_wall_collisions_from_list(struct SurfaceNode *surfaceNode,
...
+                } else {
+                    x += surf->normal.x * (radius - offset);
+                    z += surf->normal.z * (radius - offset);
+                    hasCollision = TRUE;
+                }
+            }
+
+            //Edge 1-2
+            if (!hasCollision) {
+                if (offset < 0) {
```
Isaac0-dev: offset is a float

Isaac0-dev (changes requested):

src/pc/lua/smlua_cobject.c
Sunketchupm: I personally prefer there being no space after a cast but I'll change that

Sunketchupm:
The issue I couldn't resolve is if a mod sets the `fixCollisionBugs` settings before setting `fixCollisionBugs` itself. So far I haven't thought of a solution for this.

Issue #811, opened 2025-05-16
[Lua] No way to unbubble the player


bluisblu:
I can check mario_can_bubble and call mario_set_bubbled but I can't seem to be able pop the player's bubble, either mario_set_bubbled could be updated to take an optional boolean (true by default) or there could be a mario_pop_bubble function or something.

Sunketchupm:
Currently for popping the player in Lua, we just copy over the bubble pop code in `act_bubbled` and run that when the bubble should be popped. I don't think we need a separate function for this, as there is nothing in the game that explicitly checks for if a bubble has popped.

PR #819, opened 2025-05-20
Expose and create some useful functions

Sunketchupm:
Bubble popping function suggestion from #811
I made this originally to expose `closest_point_to_triangle` but figured I'd listen to the suggestion which I wrote off a little while ago.

### DIFF
--- a/src/game/mario_actions_automatic.c
+++ b/src/game/mario_actions_automatic.c
@@ -1024,6 +1024,29 @@ static struct MarioState* nearest_antibubble_mario_state_to_object(struct Object
     return nearest;
 }
 
+/* |description|
+Makes Mario act like he was popped from a bubble. Useful for custom bubble popping behaviors.
+|descriptionEnd| */
+void mario_pop_bubble(struct MarioState* m) {
+    if (!m) { return; }
+    m->marioObj->activeFlags &= ~ACTIVE_FLAG_MOVE_THROUGH_GRATE;
+    m->hurtCounter = 0;
+    m->healCounter = 31;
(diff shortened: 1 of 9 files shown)


src/game/mario_actions_automatic.c
```diff
@@ -1024,6 +1024,29 @@ static struct MarioState* nearest_antibubble_mario_state_to_object(struct Object
...
+    m->vel[1] = 0.0f;
+    m->marioObj->header.gfx.node.flags &= ~GRAPH_RENDER_INVISIBLE;
+    m->invincTimer = 30 * 3;
+    if (m->playerIndex == 0) {
+        if (m->statusForCamera) { m->statusForCamera->action = m->action; }
+        soft_reset_camera(m->area->camera);
+    }
+    u8 underWater = (m->pos[1] < ((f32)m->waterLevel));
+    set_mario_action(m, underWater ? ACT_WATER_IDLE : ACT_FREEFALL, 0);
+}
```
PeachyPeachSM64, line 1048: Is it intended for this function to have an effect even if Mario's action is not ACT_BUBBLED?
If not, it would be good to have a `if (m->action != ACT_BUBBLED) { return false; }` at the top and change the return type to `bool` (and return true if Mario's bubble has been popped, false otherwise)
Otherwise just ignore that comment

Sunketchupm: It is intended to have an effect even if Mario is not in a bubble

PeachyPeachSM64: ok thanks for the confirmation

PR #820, opened 2025-05-22
[WIP] More math improvements

PeachyPeachSM64:
- Exposing vector constants `gVec3fZero`, `gVec3fOne`, etc... and making them immutable
- `gGlobalSoundSource` is also immutable now
- Implements `table.copy` and `table.deepcopy`
- Changing vector functions return type to vector pointer types and allowing chaining function calls
- Adding `vec3X_mult` and `vec3X_prod`

Because typing `local v = { x = 0, y = 0, z = 0 }` is annoying, you can now initialize `v` with:
```lua
local v = gVec3fZero()
```
(don't forget the `()`, because otherwise it will just reference the constant, and since the constants are immutable, `v` will be immutable too!)

Chaining function calls allows stuff like this:
```lua
-- res = (v1 * v2) + (v3 - v4)
local res = vec3f_add(vec3f_prod(gVec3fZero(), v1, v2), vec3f_dif(gVec3fZero(), v3, v4))

-- it's the same as this, but you need only one variable!
local prod = gVec3fZero()
local dif = gVec3fZero()
local res = gVec3fZero()
vec3f_prod(prod, v1, v2)
vec3f_dif(dif, v3, v4)
vec3f_sum(res, prod, dif)
```


EmilyEmmi:
> gGlobalSoundSource is also immutable now

Does this cause any mods that define gGlobalSoundSource themselves to produce a script error? I typically use gGlobalSoundSource = {x = 0, y = 0, z = 0}

Isaac0-dev:
> > gGlobalSoundSource is also immutable now
> 
> Does this cause any mods that define gGlobalSoundSource themselves to produce a script error? I typically use gGlobalSoundSource = {x = 0, y = 0, z = 0}

mods that do that will just override `gGlobalSoundSource` for the mod

Beckowl:
if you could make this for matrices as well it'd help a lot

src/pc/lua/smlua_functions.c
```diff
@@ -40,6 +40,108 @@ bool smlua_functions_valid_param_range(lua_State* L, int min, int max) {
     return true;
 }
 
+  ///////////
+ // table //
+///////////
+
+int smlua_func_table_copy(lua_State *L) {
+    if (!smlua_functions_valid_param_count(L, 1)) { return 0; }
+
```
Isaac0-dev, line 51: can you put `LUA_STACK_CHECK_`  in these functions that directly use the Lua api?
the stack management seems sound but it's the type of function that could easily have issues if someone modifies them

src/pc/lua/smlua_functions.c
```diff
@@ -40,6 +40,108 @@ bool smlua_functions_valid_param_range(lua_State* L, int min, int max) {
     return true;
 }
 
+  ///////////
+ // table //
+///////////
+
+int smlua_func_table_copy(lua_State *L) {
```
Isaac0-dev, line 47: these functions (`smlua_func_table_copy` and `smlua_func_table_deepcopy`) look like they can be handled by autogen still. similar to how `smlua_collision_util_find_surface_types` works, you can still manipulate Lua like this if you define your C function without any parameters.

autogen/lua_constants/built-in.lua
```diff
@@ -21,12 +21,45 @@ _ReadOnlyTable = {
...
+--- @param t table
+--- @return table
+function table.deepcopy(t)
+    return table_deepcopy(t)
+end
+
+--- Utility function to create a read-only table
+--- @param data table
+--- @return table
+function create_read_only_table(data)
```
Isaac0-dev, line 45: this is vulnerable to `rawset`.
the best solution would be to use the Lua Api from C, where you can actually make the memory read only

```lua
function create_read_only_table(data)
    local t = {}
    local mt = {
        __index   = data,
        __newindex= function() end,
        __call    = function() return table_copy(data) end,
        __metatable = false
    }
    setmetatable(t, mt)
    return t
end

gGlobalSoundSource = create_read_only_table({ x = 0, y = 0, z = 0 })

print(gGlobalSoundSource.x .. " " .. gGlobalSoundSource.y .. " " .. gGlobalSoundSource.z) -- 0 0 0

gGlobalSoundSource.x = 12
gGlobalSoundSource.y = 2
gGlobalSoundSource.z = 7

print(gGlobalSoundSource.x .. " " .. gGlobalSoundSource.y .. " " .. gGlobalSoundSource.z) -- 0 0 0

rawset(gGlobalSoundSource, "x", 12)
rawset(gGlobalSoundSource, "y", 2)
rawset(gGlobalSoundSource, "z", 7)

print(gGlobalSoundSource.x .. " " .. gGlobalSoundSource.y .. " " .. gGlobalSoundSource.z) -- 12 2 7
```

Isaac0-dev (changes requested):

src/pc/lua/smlua_functions.c
PeachyPeachSM64: sure I can

src/pc/lua/smlua_functions.c
PeachyPeachSM64: yeah but there is no table type in C
and if I define the function without any parameter, the Lua call would fail since it wouldn't accept any parameter at all
because of this:
```c
    int top = lua_gettop(L);
    if (top != 0) {
        LOG_LUA_LINE("Improper param count for '%s': Expected %u, Received %u", "table_copy", 0, top);
        return 0;
    }
```

autogen/lua_constants/built-in.lua
PeachyPeachSM64: the point of `rawset` is to bypass the metamethod `__newindex`
the goal here is to prevent users from *accidentally* overriding "constants", and nobody uses `rawset` by accident

I think I could raise an error instead of silently doing nothing, so that the user is at least informed that they try to modify a read-only variable

btw, you can already do the same with `gGlobalSyncTable` and `gPlayerSyncTable`, which are supposed to be read-only outside of their hidden `_table` field:
```lua
gGlobalSyncTable.test = "using metamethods"
print(gGlobalSyncTable.test) -- using metamethods (uses _table)
rawset(gGlobalSyncTable, "test", "not using metamethods")
print(gGlobalSyncTable.test) -- not using metamethods (no longer uses _table)

gPlayerSyncTable.test = "aaaa"
print(gPlayerSyncTable.test) -- nil
rawset(gPlayerSyncTable, "test", "aaaa")
print(gPlayerSyncTable.test) -- aaaa
```
(you can even destroy the global sync table with this)
```lua
rawset(gGlobalSyncTable, "_table", nil)
print(gGlobalSyncTable.test)
-- [LUA] [string "math.randomseed(get_time())..."]:5: attempt to index a nil value (local '_table')
```

src/pc/lua/smlua_functions.c
Isaac0-dev: it may be possible to implement the same thing as `smlua_to_lua_function` for tables or something.
but you're right, that makes my idea less viable

autogen/lua_constants/built-in.lua
Isaac0-dev: I didn't realise the sync tables have the same issue, i never use them lol

yes that's the point of `rawset`, but i guess my concern is just that people can break other mods using this.
but if that's what someone wants, they can easily do that and there's many ways. just pointing out what i noticed

src/pc/lua/smlua_functions.c
PeachyPeachSM64: I completely forgot `smlua_to_lua_function` existed

PeachyPeachSM64: actually I get more downsides implementing these functions this way:
- push the table twice before and after the function call, because tables are not C objects and cannot be manipulated outside of the Lua stack
- autogen struggles to recognize this type (and it needs patching here and there to work properly)
- that's not really smlua, since it's just basic Lua functionality that's not part of the standard

so I'll keep the current implementation

autogen/lua_constants/built-in.lua
```diff
@@ -167,6 +202,48 @@ function math.hypot(a, b)
...
+--- @return number
+--- Returns 1 if `x` is positive or zero, -1 otherwise
+function math.sign(x)
+    return x >= 0 and 1 or -1
+end
+
+--- @param x number
+--- @return number
+--- Returns 1 if `x` is positive, 0 if it is zero, -1 otherwise
+function math.sign0(x)
```
Isaac0-dev, line 215: Would math.sign_zero be more appropriate?

src/pc/lua/smlua_utils.c
```diff
@@ -751,15 +751,21 @@ void smlua_logline(void) {
...
+                    if (++slashCount == 2) {
+                        folderStart = p + 1;
+                        break;
+                    }
                 }
             }
+        } else {
+            // That's not a filepath
+            // It also explains why sometimes the whole gSmluaConstants string was printed to the console
+            snprintf(info.short_src, sizeof(info.short_src), "gSmluaConstants");
```
Isaac0-dev, line 768: Good catch

src/pc/lua/smlua_functions.c
```diff
@@ -40,6 +40,116 @@ bool smlua_functions_valid_param_range(lua_State* L, int min, int max) {
...
+        return 0;
+    }
+
+    // Cache to prevent copying the same table twice
+    lua_newtable(L);
+    int idxCache = lua_gettop(L);
+
+    table_deepcopy_table(L, 1, idxCache);
+
+    lua_remove(L, idxCache);
```
Isaac0-dev, line 147: Close call?

Isaac0-dev:
Looks pretty much ready

src/pc/lua/smlua_functions.c
PeachyPeachSM64: wdym?

autogen/lua_constants/built-in.lua
PeachyPeachSM64: gotta keep the one-word style for math functions

src/pc/lua/smlua_functions.c
Isaac0-dev: Nevermind sorry

PR #822, opened 2025-05-24
Fix Chat Scrolling

Cooliokid956:
and more

### DIFF
--- a/src/pc/djui/djui_chat_box.c
+++ b/src/pc/djui/djui_chat_box.c
@@ -469,12 +469,13 @@ static bool djui_chat_box_input_on_key_down(UNUSED struct DjuiBase* base, int sc
             sent_history_reset_navigation(&sentHistory);
             djui_chat_box_input_escape(gDjuiChatBox->chatInput);
             return true;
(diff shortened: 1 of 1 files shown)


src/pc/djui/djui_chat_box.c
```diff
@@ -458,18 +458,18 @@ static bool djui_chat_box_input_on_key_down(UNUSED struct DjuiBase* base, int sc
...
             sent_history_reset_navigation(&sentHistory);
             djui_chat_box_input_enter(gDjuiChatBox->chatInput);
-            break;
+            return true;
         case SCANCODE_ESCAPE:
             reset_tab_completion_all();
             sent_history_reset_navigation(&sentHistory);
             djui_chat_box_input_escape(gDjuiChatBox->chatInput);
-            break;
-        default: {
```
AgentXLP, line 472: When you remove these braces mac compilation breaks because declaring a variable inside switch cases isn't allowed for some reason, I think something to do with how mac gcc handles scope

AgentXLP (approved):

PR #825, opened 2025-05-25
Fix dpad logic in menus

Gumbo64:
Without this change, there is no way to get the cursor onto the ">" button in the public lobbies pagination without using a mouse, because the dpad/arrowkey logic skips over it strangely. Another alternative solution could be that the rules button and the back/reset buttons swap places.

### DIFF
--- a/src/pc/djui/djui_cursor.c
+++ b/src/pc/djui/djui_cursor.c
@@ -85,8 +85,8 @@ static void djui_cursor_move_check(s8 xDir, s8 yDir, struct DjuiBase** pick, str
             if (*pick == NULL) {
                 *pick = base;
             } else {
-                f32 pickDist = djui_cursor_base_distance(*pick, xDir, yDir ? 1.0f : 2.0f);
-                f32 baseDist = djui_cursor_base_distance(base,  xDir, yDir ? 1.0f : 2.0f);
+                f32 pickDist = djui_cursor_base_distance(*pick, xDir ? 1.0f : 1.2f, yDir ? 1.0f : 2.0f);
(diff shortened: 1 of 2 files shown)


src/pc/djui/djui_cursor.c
```diff
@@ -85,8 +85,8 @@ static void djui_cursor_move_check(s8 xDir, s8 yDir, struct DjuiBase** pick, str
             if (*pick == NULL) {
                 *pick = base;
             } else {
-                f32 pickDist = djui_cursor_base_distance(*pick, xDir ? 1.0f : 2.0f, yDir ? 1.0f : 2.0f);
-                f32 baseDist = djui_cursor_base_distance(base,  xDir ? 1.0f : 2.0f, yDir ? 1.0f : 2.0f);
+                f32 pickDist = djui_cursor_base_distance(*pick, xDir, yDir ? 1.0f : 2.0f);
+                f32 baseDist = djui_cursor_base_distance(base,  xDir, yDir ? 1.0f : 2.0f);
```
PeachyPeachSM64: this change creates other bugs:
in Options -> Controls -> N64 binds, pressing D-pad left always puts the cursor to the leftmost button, no matter the current position

it also breaks completely mouse to cursor position:
still in this menu, move the mouse over any button and press a D-pad direction: it will move the cursor to another button randomly instead of an adjacent one

PeachyPeachSM64 (changes requested):

src/pc/djui/djui_cursor.c
Gumbo64: thank you for noticing. you're right lol i forgot how ternaries worked for a second. should be fixed now

PeachyPeachSM64 (approved):

PR #826, opened 2025-05-26
smlua event hooks refactor

PeachyPeachSM64:
No performance improvement, no optimization, just plain refactor for the sake of readability and maintainability.
Hooks functions are autogened because copy/pasting the same piece of code dozens of times is bad.
Error message is now `[LUA] Failed to call the callback for hook HOOK_NAME` instead of the hook enum index.

Also adding https://github.com/coop-deluxe/sm64coopdx/pull/834, and overridable `pos` in `HOOK_ON_NAMETAGS_RENDER` (suggested by @Beckowl)

Some observations:
- Most hooks never return before calling all callbacks for a given event, except some that return on first successful call or after setting the ouput. That's why there is a `HOOK_RETURN_*` parameter for each hook. This is really inconsistent and some of these hooks don't even need an early return (probably due to some copy/paste sh*t when implementing new hooks). But I kept it so it doesn't break any mod.
- `HOOK_ON_HUD_RENDER` and `HOOK_ON_HUD_RENDER_BEHIND` cannot be autogened because *backwards compatibility*.
- Naming hook functions `smlua_call_event_hooks_<HOOK_NAME>` isn't very consistent with the coding style, but at least it's explicit enough.
- What's the point of `HOOK_ON_DJUI_THEME_CHANGED`?

## A new (proper) way to implement hooks

To create a new hook, add this line to `src/pc/lua/smlua_hook_events.inl`:
```c
SMLUA_EVENT_HOOK(HOOK_NAME, HOOK_RETURN_*, paramType paramName, ..., OUTPUT outputType *outputName, ...)
```
- As usual, add your hook name to the `LuaHookedEventType` enum.
- `HOOK_RETURN_*` has 3 possible values, check `LuaHookedEventReturn` enum for description.
- There can be any number of parameters or outputs.
- Prefix outputs with `OUTPUT` (before the type). Outputs should always be pointers.
- Don't forget to run autogen. If a type doesn't exist in autogen, add it to `SMLUA_TYPES` in `autogen/gen_hooks.py`. Use other types as examples to properly define it.
- Hook function name is `smlua_call_event_hooks_<HOOK_NAME>`. These functions always return a boolean. When they return `true` is described in the `LuaHookedEventReturn` enum.


autogen/gen_hooks.py
```diff
@@ -0,0 +1,308 @@
...
+
+    parameters = ", ".join(token.replace("OUTPUT ", "") for token in tokens[3:])
+    inputs = []
+    outputs = []
+    for token in tokens[3:]:
+        param = [x.strip() for x in token.split() if x.strip()]
+        if len(param) < 2:
+            continue
+
+        is_output = param[0] == "OUTPUT"
```
Isaac0-dev, line 214: nice trick

Isaac0-dev (commented):
this is very cool, awesome work!

why not define a macro that takes the hook name and concatinates it to the function name, so `smlua_call_event_hooks(HOOK_ON_SYNC_VALID)` becomes `smlua_call_event_hooks_HOOK_ON_SYNC_VALID()`? might fix the code style issue.

Can you now use the constant names in `smlua_hook_events.inl` for `enum LuaHookedEventType`? that could save the amount of steps you need to add a hook

PeachyPeachSM64:
> why not define a macro that takes the hook name and concatinates it to the function name, so smlua_call_event_hooks(HOOK_ON_SYNC_VALID) becomes smlua_call_event_hooks_HOOK_ON_SYNC_VALID()? might fix the code style issue.

damn that's exactly what I was thinking

> Can you now use the constant names in smlua_hook_events.inl for enum LuaHookedEventType? that could save the amount of steps you need to add a hook

yes, but autogen wouldn't be able to find them

src/pc/nametags.c
```diff
@@ -84,21 +84,24 @@ void nametags_render(void) {
...
-            smlua_call_event_hooks_int_params_ret_string(HOOK_ON_NAMETAGS_RENDER, i, &hookedString);
+            const char* hookedString = NULL;
+            smlua_call_event_hooks(HOOK_ON_NAMETAGS_RENDER, i, pos, &hookedString);
             if (hookedString) {
                 snprintf(name, MAX_CONFIG_STRING, "%s", hookedString);
             } else {
                 snprintf(name, MAX_CONFIG_STRING, "%s", np->name);
                 name_without_hex(name);
             }
+            if (!djui_hud_world_pos_to_screen_pos(pos, out)) {
```
Isaac0-dev, line 99: what's this new `djui_hud_world_pos_to_screen_pos` for?

autogen/gen_hooks.py
```diff
@@ -0,0 +1,311 @@
...
+        lua_getglobal(L, "gMarioStates");
+        lua_pushinteger(L, {name}->playerIndex);
+        lua_gettable(L, -2);
+        lua_remove(L, -2);
+"""
+    },
+    "structObject*": {
+"input": """
+        // push {name}
+        smlua_push_object(L, LOT_OBJECT, {name}, NULL);
```
Isaac0-dev: for these manual struct definitions that only contain a `smlua_push_object`, can't you just assume the LOT exists, and make it do a `smlua_push_object` by default for structs? if you need, you could move `gen_hooks.py` in `autogen.sh` to after `convert_structs.py` so that you can check if the LOT exists, but honestly that's up to the person adding the hook imo, it'll throw an error at compile time anyway.

Isaac0-dev (changes requested):
can you fix merge conflicts too, I just merged dj's pr which adds a hook.

src/pc/nametags.c
PeachyPeachSM64: `out` needs to be recalculated if `pos` is modified inside HOOK_ON_NAMETAG_RENDER

Isaac0-dev: ah, could do a memcmp then to see if it was modified. idk if its actually worth it so its not vital

autogen/gen_hooks.py
PeachyPeachSM64: done

PR #834, opened 2025-05-30
Add ``ModelExtendedId`` param to ``HOOK_OBJECT_SET_MODEL`` 

Beckowl:
The ``HOOK_OBJECT_SET_MODEL`` event only provides you a regular model ID, which isn't very useful in Lua. While it is technically possible to get the extended ID using ``get_obj_model_extended`` inside the hook, doing so kinda defeats the purpose of having the hook in the first place. This PR adds a third parameter ``ModelExtendedId`` to the hook

I also exposed ``smlua_model_util_ext_id_to_id`` and ``smlua_model_util_id_to_ext_id``.


Isaac0-dev:
Don't expose those functions because they may be removed later

PR #835, opened 2025-05-31
Add Scroll Binds

Cooliokid956:
and fix a five year oob array index/write


src/pc/controller/controller_bind_mapping.c
```diff
@@ -106,11 +106,17 @@ const char* translate_bind_to_name(int bind) {
...
-        snprintf(name, 8, "Mouse %d", bind - VK_BASE_SDL_MOUSE);
-        return name;
+        switch (mouse_button) {
+            case 1: return "L Mouse";
+            case 2: return "M Mouse";
+            case 3: return "R Mouse";
+            case 6: return "Scroll Up";
+            case 7: return "Scroll Down";
+            default: {
+                snprintf(name, 8, "Mouse %d", mouse_button);
```
PeachyPeachSM64: why 8? `name` is 11 chars long

src/pc/controller/controller_sdl2.c
```diff
@@ -285,9 +285,7 @@ static void controller_sdl_read(OSContPad *pad) {
     u32 buttons_down = 0;
     for (u32 i = 0; i < num_joy_binds; ++i)
         if (joy_buttons[joy_binds[i][0]])
-            buttons_down |= joy_binds[i][1];
-
-    pad->button |= buttons_down;
+            pad->button |= joy_binds[i][1];
```
PeachyPeachSM64: why that change?

PeachyPeachSM64 (changes requested):

src/pc/controller/controller_bind_mapping.c
Isaac0-dev: "Scroll Down" is 11 chars long but this only needs to be 8, assuming mouse_button is 1 digit (which it should be).
No need to allow for more chars than expected.

Cooliokid956: ask dj I didn't type that number 😭

src/pc/controller/controller_sdl2.c
Cooliokid956: this section is functionally (and now) identical to L194-196

PeachyPeachSM64: no, you're not setting `buttons_down` anymore, meaning the following section will have no effect

PeachyPeachSM64 (changes requested):

src/pc/controller/controller_sdl2.c
Cooliokid956: @PeachyPeachSM64 is that good

src/pc/controller/controller_sdl2.c
```diff
@@ -282,15 +282,12 @@ static void controller_sdl_read(OSContPad *pad) {
...
         if (joy_buttons[joy_binds[i][0]])
-            buttons_down |= joy_binds[i][1];
+            pad->button |= joy_binds[i][1];
 
-    pad->button |= buttons_down;
-
-    const u32 xstick = buttons_down & STICK_XMASK;
-    const u32 ystick = buttons_down & STICK_YMASK;
+    const u32 xstick = pad->button & STICK_XMASK;
+    const u32 ystick = pad->button & STICK_YMASK;
```
PeachyPeachSM64: !! That's a regression !!
pad->button is a u16, meaning STICK bits will never be set

there is also a bug that needs fixing in this function
buttons_down is supposed to catch those STICK bits, but it only checks joystick binds, not mouse binds (meaning you can't move up with a mouse click for example)
a potential fix would be to move buttons_down all the way up, replace all occurrences of pad->button with buttons_down, then do pad->button |= buttons_down at the end of the function

PeachyPeachSM64 (changes requested):

src/pc/controller/controller_sdl2.c
Cooliokid956: Okay I have reverted and fixed that bug, Tested

xLuigiGamerx:
I think this is absolute cinema

idiotforge:
LGTM🔥

PR #837, opened 2025-06-01
Fix vec3f bugs

PeachyPeachSM64:
- Fix script errors when using read-only `gGlobalSoundSource` with `play_sound` functions
- Vec-type parameters are no longer pushed to their corresponding Lua table after the function call, unless the parameter is prefixed with `OUT` in the function C definition
ex:
  - `func(Vec3f dest, Vec3f src)` -> `src` and `dest` are not updated
  - `func(OUT Vec3f dest, Vec3f src)` -> only `dest` is updated after the call to `func`

This also slightly improves performance, as parameters that are never modified are no longer pushed after a function call (and makes it safe to use Vec constants as such parameters)
- remove `vec3f_copy_2`
- add `object_face_angle_to_vec3s`, `vec3s_to_object_face_angle`, `object_move_angle_to_vec3s`, `vec3s_to_object_move_angle`


autogen/lua_constants/built-in.lua
```diff
@@ -61,7 +61,7 @@ end
 -----------
 
 --- @type Vec3f
-gGlobalSoundSource = { x = 0, y = 0, z = 0 }
+gGlobalSoundSource = create_read_only_table({ x = 0, y = 0, z = 0 })
```
Isaac0-dev, line 64: why was this changed in the first place?

docs/lua/functions-3.md
```diff
@@ -537,7 +537,7 @@ Converts an object's position to a `Vec3f` format. Useful for aligning object be
 - None
 
 ### C Prototype
-`void object_pos_to_vec3f(Vec3f dst, struct Object *o);`
+`void object_pos_to_vec3f(OUT Vec3f dst, struct Object *o);`
```
Isaac0-dev, line 540: probably should remove `OUT` from parameter names from the "C prototype", it would only lead to confusion for mod devs

autogen/lua_constants/built-in.lua
PeachyPeachSM64: it created script errors with play_sound trying to modify the second parameter

PR #842, opened 2025-06-07
R - OPTIONS prompt

kingthememer:
removed the djui format R Button - Options prompt in place of a vanilla format R - OPTIONS prompt handled alongside all the vanilla pause text and R_TRIG detection
image


Beckowl:
Doesn't this break language support? The vanilla font doesn't have as many characters that the DJUI one does.

kingthememer:
> Doesn't this break language support? The vanilla font doesn't have as many characters that the DJUI one does.

for this specific peice of text, yes but i feel like that doesnt really matter as much in the grand scheme of things
the rest of the menu text doesnt have language support outside of what the vanilla pal build already had

eros71-dev:
Honestly I do believe it should be handled with DJUI, because we also should make it get proper translations like the rest of the Co-op DX additions

kingthememer:
> Honestly I do believe it should be handled with DJUI, because we also should make it get proper translations like the rest of the Co-op DX additions to the vanilla game

the way i see it, its supposed to be a part of the vanilla pause screen
if we were talking about something in the djui menu id be more understanding, that should have localizability so more can actually set up/host/join the game itself, but this is an ingame thing like the added EXIT TO CASTLE text (which funnily enough is also on the pause screen)

Beckowl:
The Czech, Japanese, Portuguese, Russian, and Spanish translations contain non-ASCII characters in the ``R_BUTTON`` text. What happens when you enable these translations?

kingthememer:
> The Czech, Japanese, Portuguese, Russian, and Spanish translations contain non-ASCII characters in the `R_BUTTON` text. What happens when you enable these translations?

oh no i understand what youre saying, yeah those translations wont do anything
my point is, they dont affect ANYTHING in the vanilla pause menu, even other extras like the `EXIT TO CASTLE` text, so why should they need to affect this one other string at all?

the pr makes it seem like that options text was in the game from the start in a way, obviously this isnt true but you get the idea

eros71-dev:
Okay, after talking it out, yeah I do see that the one we have right now does feel out of place, specially with the lack of a fade in

And it may improve performance every so slightly? Not sure, I do know DJUI makes the game run slower when rendering stuff

Only downside is translations, me personally I don't have any problems with this PR now, if for the sake of consistency, hopefully we can maybe do some sort of (crappy probably) workaround to load the translated string based on the language setting, a little ugly unless we make another way to load the R_BUTTON string from the lang files when opening up the pause menu or something, maybe store it in a variable and then use that, updating it whenever the language is changed

That sounds reasonable and fixes both problems I would say

PR #843, opened 2025-06-07
Add HOOK_ON_ADD_SURFACE and HOOK_ON_CLEAR_AREAS

djoslin0:
HOOK_ON_ADD_SURFACE allows mods to read both static and dynamic collision geometry

HOOK_ON_CLEAR_AREAS gives an obvious entry point to erase anything related to collisions in a mod


Isaac0-dev:
It looks like the dynamic parameter should be treated as a boolean, it's so annoying that Lua treats 0 as truthy

djoslin0:
yeah that's a good idea. I changed `dynamic` to bool

PR #846, opened 2025-06-08
replace "auto" djui scale functionality with 4x native resolution

kingthememer:
with the djui scale set to "auto", all menu graphics will scale to the ingame resolution rather than being locked to the system
this way the menu can look consistent across all resolutions and window sizes

preview with a medium sized window:
image

preview with an extremely small window:
image


Isaac0-dev:
we'll try it out

PR #847, opened 2025-06-11
Add a safer version of Lua's require()

djoslin0:
I didn't add standard Lua require() because I've always been afraid of it. I'm not sure we can guarantee which files it will read (or not read).

Instead, here is a custom implementation. It should work more or less the same and allow for more modular code.

For backwards compatibility reasons, all of the lua files in the base mod folder will be loaded as in the past. Aka one at a time and alphabetically.

However, now coop will look for Lua files in subdirectories and will load them in when another Lua file calls require().

The file search order is more reasonable than normal Lua require(). It will first look for files relative to the currently running script. If there is no matching relative file, it will pick from any Lua file that is in any of the mod's subdirectories.

### DIFF
--- a/src/pc/lua/smlua_require.c
+++ b/src/pc/lua/smlua_require.c
@@ -90,62 +90,18 @@ static struct ModFile* smlua_find_mod_file(const char* moduleName) {
     return bestPick;
 }
 
-static int smlua_load_mod_file(lua_State* L, struct ModFile* file) {
-    if (!file->cachedPath) {
-        LOG_LUA("Lua 'require' error: module '%s' not found in mod files", file->relativePath);
-        return LUA_ERRFILE;
-    }
-
-    FILE* f = f_open_r(file->cachedPath);
-    if (!f) {
-        LOG_LUA("Lua 'require' error: cannot open '%s': file not found", file->cachedPath);
-        return LUA_ERRFILE;
-    }
-
-    // get file size
-    f_seek(f, 0, SEEK_END);
-    size_t length = f_tell(f);
-    f_rewind(f);
-
-    // allocate buffer
-    void* buffer = calloc(length + 1, 1);
-    if (!buffer) {
-        f_close(f);
-        f_delete(f);
-        LOG_LUA("Lua 'require' error: cannot allocate buffer for '%s'", file->cachedPath);
-        return LUA_ERRMEM;
-    }
-
-    // read file
-    if (f_read(buffer, 1, length, f) < length) {
-        free(buffer);
-        f_close(f);
-        f_delete(f);
-        LOG_LUA("Lua 'require' error: cannot read '%s'", file->cachedPath);
-        return LUA_ERRFILE;
-    }
-
-    f_close(f);
-    f_delete(f);
-
-    // load the buffer as lua code
-    int result = luaL_loadbuffer(L, buffer, length, file->relativePath);
-    free(buffer);
-
-    return result;
-}
-
 static int smlua_custom_require(lua_State* L) {
     const char* moduleName = luaL_checkstring(L, 1);
 
-    if (!gLuaActiveMod) {
+    struct Mod* activeMod = gLuaActiveMod;
+    if (!activeMod) {
         LOG_LUA("require() called outside of mod context");
         return 0;
     }
 
     // create registry key for this mod's loaded table
     char registryKey[SYS_MAX_PATH + 16] = "";
-    snprintf(registryKey, sizeof(registryKey), "mod_loaded_%s", gLuaActiveMod->relativePath);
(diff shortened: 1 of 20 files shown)


djoslin0:
Here's how to use it.
```
-- file at src/test.lua

local function on_update()
    djui_chat_message_create('updating...')
end

hook_event(HOOK_UPDATE, on_update)

return 123
```


```
-- file at main.lua

-- load and execute the file
local test_variable =  require('src/test')

-- test_variable is equal to 123

-- further invokations of require() don't re-run the file, they just return the stored reference
test_variable =  require('test')
test_variable =  require('test.lua')
test_variable =  require('src/test.lua')
test_variable =  require('src/test')
-- all of those are loading the same file
```

This allows for way better organization and usage of libraries.

src/pc/lua/smlua_require.c
```diff
@@ -0,0 +1,239 @@
...
+            bestPick = file;
+            bestRelativeDepth = relativeDepth;
+            bestTotalDepth = totalDepth;
+        }
+    }
+
+    return bestPick;
+}
+
+static int smlua_load_mod_file(lua_State* L, struct ModFile* file) {
```
Isaac0-dev: `smlua_load_mod_file` looks like a rewrite of `smlua_load_script`. could it not be adapted to work for both types of script loading?

djoslin0:
For example, this would be a bit more advanced but more reasonable.
```
-- src/pool_array.lua

local PoolArray = {}
PoolArray.__index = PoolArray

--- Create a new PoolArray instance.
-- @param allocation_fn  function() → new element
-- @param overwrite_fn   function(arg) → replacement element (optional)
function PoolArray:create(allocation_fn, overwrite_fn)
    local o = {
        items               = {},
        count               = 0,
        allocation_function = allocation_fn,
        overwrite_function  = overwrite_fn,
    }
    setmetatable(o, self)
    return o
end

--- Reset the pool (so next :new() will start re-using from 1).
function PoolArray:clear()
    self.count = 0
end

--- Grab the next item from the pool, allocating or overwriting as needed.
-- @param arg  passed to overwrite_function (if any)
function PoolArray:new(arg)
    self.count = self.count + 1
    if self.count > #self.items then
        self.items[self.count] = self.allocation_function()
    end
    if self.overwrite_function then
        self.items[self.count] = self.overwrite_function(arg)
    end
    return self.items[self.count]
end

return PoolArray
```
```
-- main.lua

local PoolArray = require("pool_array")

-- create a pool that allocates tables and re-initializes them
local myPool = PoolArray:create(function() return {} end)


local function on_sync_valid()
    -- outputs 0
    djui_chat_message_create('item count before: ' .. myPool.count)

    -- add to pool
    local obj = myPool:new()
    obj.foo = "bar"

    -- outputs 1
    djui_chat_message_create('item count after: ' .. myPool.count)

    -- clear the pool
    myPool:clear()

    -- outputs 0
    djui_chat_message_create('item count clear: ' .. myPool.count)

end

hook_event(HOOK_ON_SYNC_VALID, on_sync_valid)
```

src/pc/lua/smlua_require.c
djoslin0: I'll take a look, I should probably reuse the existing one

Isaac0-dev: when reading your implmentation i did feel like yours does some things better so it may be worth carrying them over if it's compatible.

src/pc/lua/smlua_require.c
```diff
@@ -0,0 +1,179 @@
...
+        lua_pop(L, 1);
+        lua_newtable(L);
+        lua_pushvalue(L, -1);
+        lua_setfield(L, LUA_REGISTRYINDEX, registryKey);
+    }
+
+    // find the file in mod files
+    struct ModFile* file = smlua_find_mod_file(moduleName);
+    if (!file) {
+        LOG_LUA("module '%s' not found in mod files", moduleName)
```
Isaac0-dev: missing a semicolon, its a macro without any issues but consistency is nice

src/pc/lua/smlua_require.c
```diff
@@ -0,0 +1,179 @@
...
+        return 0;
```
Isaac0-dev, line 120: you need to pop the table when you return here (use `LUA_STACK_CHECK_*` macros to test this)

Isaac0-dev:
looks pretty good other than those 2 things

src/pc/lua/smlua_require.c
djoslin0: fixed

src/pc/lua/smlua_require.c
djoslin0: fixed

djoslin0:
addressed comments and plumbed the rest of the hooks to know their ModFile

PR #849, opened 2025-06-14
type checking for dynos gfx symbol parsing

Isaac0-dev:
previously, you could do things like `gsSPDisplayList(123)` to crash the game via DynOS bins or `gfx_set_command`. because it's constructing a pointer out of 123. This is just one of many issues that could be created by providing the wrong type for a gfx macro. This is a slightly hacky solution which keeps track of the types of all gfx commands that dynos supports.

### DIFF
--- a/data/dynos_bin_gfx.cpp
+++ b/data/dynos_bin_gfx.cpp
@@ -9,6 +9,27 @@ extern "C" {
 #include "include/macros.h"
 }
 
+#define GFX_PARAM_TYPE_INT 'i'
+#define GFX_PARAM_TYPE_STR 's'
+#define GFX_PARAM_TYPE_PTR 'p'
+#define GFX_PARAM_TYPE_VTX 'v'
+#define GFX_PARAM_TYPE_TEX 't'
+#define GFX_PARAM_TYPE_GFX 'g'
+
+#define GFX_PARAM_TYPE_IS_POINTER(paramType) ( \
+    paramType == GFX_PARAM_TYPE_PTR || \
+    paramType == GFX_PARAM_TYPE_VTX || \
+    paramType == GFX_PARAM_TYPE_TEX || \
+    paramType == GFX_PARAM_TYPE_GFX    \
+)
+
+#define GFX_PARAM_TYPE_IS_INT_OR_CONSTANT(paramType) ( \
+    paramType == GFX_PARAM_TYPE_INT || \
+    paramType == GFX_PARAM_TYPE_STR    \
+)
+
+typedef char GfxParamType;
+
 static std::map<std::string, std::pair<Gfx *, u32>> sGfxCommandCache;
 static char *sGfxCommandErrorMsg = NULL;
 static u32 sGfxCommandErrorSize = 0;
@@ -426,16 +447,14 @@ s64 DynOS_Gfx_ParseGfxConstants(const String& _Arg, bool* found) {
     return 0;
 }
 
-static s64 ParseGfxSymbolArg(GfxData* aGfxData, DataNode<Gfx>* aNode, u64* pTokenIndex, const char *aPrefix, const char type) {
+static s64 ParseGfxSymbolArg(GfxData* aGfxData, DataNode<Gfx>* aNode, u64* pTokenIndex, const char *aPrefix, const GfxParamType aParamType) {
     assert(aPrefix != NULL);
     if (pTokenIndex != NULL) { CHECK_TOKEN_INDEX(*pTokenIndex, 0); }
     String _Token = (pTokenIndex != NULL ? aNode->mTokens[(*pTokenIndex)++] : "");
     String _Arg("%s%s", aPrefix, _Token.begin());
-    bool isPtrType = (type == 'p' || type == 't' || type == 'g' || type == 'v');
-    bool isIntOrConstant = (type == 's' || type == 'i'); // Allow strings as constants for integer parameters
 
     // Integers
-    if (isIntOrConstant) {
+    if (GFX_PARAM_TYPE_IS_INT_OR_CONSTANT(aParamType)) {
         bool integerFound = false;
         s64 integerValue = DynOS_Misc_ParseInteger(_Arg, &integerFound);
         if (integerFound) {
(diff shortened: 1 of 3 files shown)


data/dynos_bin_gfx.cpp
```diff
@@ -426,29 +426,33 @@ s64 DynOS_Gfx_ParseGfxConstants(const String& _Arg, bool* found) {
...
-    for (auto& _Node : aGfxData->mRawPointers) {
-        if (_Arg == _Node->mName) {
-            return (s64) _Node->mData;
+    if (type == 'p') {
+        for (auto& _Node : aGfxData->mRawPointers) {
+            if (_Arg == _Node->mName) {
+                return (s64) _Node->mData;
+            }
         }
     }
```
PeachyPeachSM64: raw pointers are used internally by gfx_set_command
checking type here breaks this function completely

data/dynos_bin_gfx.cpp
```diff
@@ -461,125 +465,144 @@ static s64 ParseGfxSymbolArg(GfxData* aGfxData, DataNode<Gfx>* aNode, u64* pToke
...
+        s64 rdValue = DynOS_RecursiveDescent_Parse(_Arg.begin(), &rdSuccess, DynOS_Gfx_ParseGfxConstants);
+        if (rdSuccess) {
+            return rdValue;
+        }
+    }
+
+    // Allow NULL for pointer types
+    if ((type == 'p' || type == 't' || type == 'g' || type == 'v') && _Arg == "NULL") {
+        return 0;
     }
```
PeachyPeachSM64: 0 is also a valid pointer constant

data/dynos_bin_gfx.cpp
```diff
@@ -1167,26 +1194,63 @@ static String ResolveParam(lua_State *L, GfxData *aGfxData, u32 paramIndex, char
             [] (lua_State *L, u32 paramIndex) { return (Gfx *) smlua_to_cobject(L, paramIndex, LOT_GFX); },
             [&aGfxData] (Gfx *gfx) { return CreateRawPointerDataNode(aGfxData, gfx); }
         );
+
+        // Unimplemented pointer type
+        case 'p': {
+            PrintDataErrorGfx("  ERROR: Unimplemented pointer type: '%c'", paramType);
+            return "";
+        }
```
PeachyPeachSM64: what's the point?
`%p` is not valid, it shouldn't give a different error message

data/dynos_bin_gfx.cpp
```diff
@@ -718,52 +741,56 @@ static void ParseGfxSymbol(GfxData* aGfxData, DataNode<Gfx>* aNode, Gfx*& aHead,
...
     // Uses macro iterators to dynamically handle the correct number of parameters
-#define HANDLE_PARAM(paramNum) s64 _Arg##paramNum = ParseGfxSymbolArg(aGfxData, aNode, &aTokenIndex, "");
+#define HANDLE_PARAM(paramNum) s64 _Arg##paramNum = ParseGfxSymbolArg(aGfxData, aNode, &aTokenIndex, "", types[paramNum - 1]);
 #define GET_ARG(paramNum) _Arg##paramNum
 #define CALL_SYMB(symb, ...) symb(__VA_ARGS__)
 #define define_gfx_symbol(symb, params, addPtr, ...)                                                 \
 if (_Symbol == #symb) {                                                                              \
-    REPEAT(HANDLE_PARAM, params);                                                                    \
+    static const char types[] = { __VA_ARGS__ };                                              \
+    REPEAT(HANDLE_PARAM, params);                                                      \
```
PeachyPeachSM64: misaligned backslashes

data/dynos_bin_gfx.cpp
```diff
@@ -718,52 +741,56 @@ static void ParseGfxSymbol(GfxData* aGfxData, DataNode<Gfx>* aNode, Gfx*& aHead,
...
     if (addPtr) { aGfxData->mPointerList.Add(aHead); }                                               \
     Gfx _Gfx[] = { CALL_SYMB(symb, LIST_ARGS(GET_ARG, params)) };                                    \
     memcpy(aHead, _Gfx, sizeof(_Gfx));                                                               \
     aHead += (sizeof(_Gfx) / sizeof(_Gfx[0]));                                                       \
     return;                                                                                          \
 }
+#define define_gfx_symbol_manual(...)
 #include "gfx_symbols.h"
+define_gfx_symbol(gsSPVertexNonGlobal, 3, true, 'v', 'i', 'i');
```
PeachyPeachSM64: ?

data/dynos_bin_gfx.cpp
```diff
@@ -1167,26 +1194,63 @@ static String ResolveParam(lua_State *L, GfxData *aGfxData, u32 paramIndex, char
...
+struct ParamInfo {
+    u8 count;
+    char types[];
+};
+
+static struct ParamInfo *get_gfx_command_param_data(const char *command) {
+#define define_gfx_symbol(symb, params, addPtr, ...)                                                       \
+static struct ParamInfo info_##symb = { .count = params, .types = { __VA_ARGS__ } };                       \
+static_assert(COUNT_ARGS(__VA_ARGS__) == (params + 1), "Parameter count does not match for gfx command."); \
+if (!strncmp(#symb, command, strlen(#symb))) { return &info_##symb; }
```
PeachyPeachSM64: indent these for better clarity

data/dynos_bin_gfx.cpp
```diff
@@ -1167,26 +1194,63 @@ static String ResolveParam(lua_State *L, GfxData *aGfxData, u32 paramIndex, char
...
+    if (!paramInfo) { PrintDataErrorGfx("  ERROR: Unknown gfx command: %s", command); return ""; }
+
     std::string output;
+    u8 paramNum = 0;
     for (u32 paramIndex = 3; *command; command++) {
         char c = *command;
         if (c == '%') {
             char paramType = *(++command);
+            char expectedType = paramInfo->types[paramNum];
+            if (paramNum <= paramInfo->count && expectedType != paramType) {
```
PeachyPeachSM64: the bounds check is useless if it is done after reading from the buffer

PeachyPeachSM64 (changes requested):

data/dynos_bin_gfx.cpp
Isaac0-dev: good catch. problem is, it could mean you can do `gsSPDisplayList(some_vtx_buffer)`

data/dynos_bin_gfx.cpp
Isaac0-dev: whoops, that was for testing

data/dynos_bin_gfx.cpp
Isaac0-dev: some of the macros that we support take pointer types that we don't natively support for `gfx_set_command`, like `gsSPLight`.

data/dynos_bin_gfx.cpp
Isaac0-dev: that check is there to prevent it throwing this error, when the actual problem is that they had too many parameters

Isaac0-dev:
@PeachyPeachSM64 thanks for your changes, I have applied them. Are we good to merge this?

PeachyPeachSM64 (approved):

PR #850, opened 2025-06-14
Mod file system

PeachyPeachSM64:
Mod file system

A persistent storage per mod with extended capabilities:
- create/move/copy up to 256 named files, binary or text
- read and write any kind of data up to a total size of 16MB per mod, and possibility to read other mods files
- allow users to handle errors, most functions have a boolean return value indicating success or failure, errors do not stop code execution

Added a developer script to turn a directory and its content into a .modfs file, or extract files from a .modfs file

Extras:
- `lua_Integer` and `lua_Number` added to autogen
- optional parameters added to autogen


ThePlayerRolo:
wait so this could allow writing save files? That could  extremely in handy for some ideas, like coopdx event mods. (Egg hunt!!!)

PeachyPeachSM64:
yes, that's a free 16MB storage where you can put (almost) anything

ThePlayerRolo:
Letss goooo, Maybe i could implement a Roblox Style Egg hunt with saving

xLuigiGamerx:
hmmm, I'm intrigued, does this allow stuff like loading sound files from other mods or creating an `.ini` reader and does this work with any file format, like can i create a custom `.anim` format that has model animation data that i can load within a mod?

PR #851, opened 2025-06-14
Add mod development mode

djoslin0:
Mod development mode can only be enabled on direct connection hosts.

With mod development mode on you can press the L bind while paused to
reload the active mods. This reload will rescan the directories for
the active modes and thus refresh their file caches.

Mod development mode also enables live lua module reloading. Lua
modules are the lua files loaded through require(). Any time a lua module
is updated, coop will live reload the functions that changed and do its best
to maintain the previous variable states.

Keep in mind, live reloading does execute the new version of the file so if
that file changes the game state outside of functions it will apply that change
every time the file is reloaded.

Example of L-bind reload in action:
https://www.youtube.com/watch?v=PQlu82FzgWs

Example of live module reload in action:
https://www.youtube.com/watch?v=PIDtm4u0FY4


Beckowl:
Could we enable the `debug`, `io`, and `os` libraries, and add `load`, `loadfile`, `loadstring`, and `dofile` in development mode?

djoslin0:
I didn't want to enable libraries that aren't allowed to be used normally, that could be a larger discussion. I wouldn't want to add them myself since I think people may accidentally leave code in a release that causes errors outside of mod dev mode.

I definitely wouldn't want `os` to be loaded though. I go out of my way to disable that when I make dev builds since it opens a massive attack vector. (You could unknowingly enable some mod with a backdoor in it)

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+            lua_settable(L, dstIdx);  // dstTable[key] = newVal
+        }
+
+        // pop oldVal and newVal (leave key for next lua_next)
+        lua_pop(L, 2);
+    }
+}
+
+static void smlua_reload_module(lua_State *L, struct Mod* mod, struct ModFile *file) {
+    LUA_STACK_CHECK_BEGIN();
```
Isaac0-dev: all uses of the `LUA_STACK_CHECK_*` macros now need to provide the Lua state as a parameter because coroutines create a new Lua state. (there should be compile errors if you don't provide it?)

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+    lua_pushvalue(L, -1);  // duplicate the function closure (val)
+    // stack: ..., key, val, val_copy
+
+    int refRegistered = luaL_ref(L, LUA_REGISTRYINDEX);
+    // luaL_ref pops the copy, leaving the original 'val' in place
+    // stack: ..., key, val
+
+    int funcKeyRef = upval_references_deduplicate(L, sFuncKeyStr, refRegistered, isOld);
+
+    // now fnIdx points to the real function
```
Isaac0-dev: i found this comment strange, this code is actually storing the stack index of the reference to the function we're processing.

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+            return ref->reference;
+        }
+
+        lua_pop(L, 1); // pop only existFunc, leave newFunc for next iteration
+    }
+
+    lua_pop(L, 1); // pop newFunc
+
+    // allocate
+    UpvalReference *ref = calloc(1, sizeof(struct UpvalReference));
```
Isaac0-dev: use `malloc` here. `calloc` does 2 extra things on top of what `malloc` does.
1. it obviously provides a clean way to specify the number of objects to allocate.
2. it does `memset(ptr, 0, size);` on it. as in it clears all the garbage memory.
but both of these things you're not using, you're already overwriting all the garbage memory here by filling in all the fields.

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+
+        // now on top of the stack is the upvalue's 'value'
+        // we pin it in the registry
+        lua_pushvalue(L, -1);
+        int ref = luaL_ref(L, LUA_REGISTRYINDEX);
+        // pop the original
+        lua_pop(L, 1);
+
+        // allocate a new record
+        UpvalRecord *rec = calloc(1, sizeof(struct UpvalRecord));
```
Isaac0-dev: same here

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+#include "smlua.h"
+#include "pc/mods/mods.h"
+#include "pc/mods/mods_utils.h"
+
+#define LIVE_RELOAD_TICK_COUNT 15
+
+typedef struct UpvalRecord {
+    char *funcKeyStr;   // the table key under which the function lived
+    int funcKeyRef;     // registry ref for the key
+    char *name;         // the upvalue's name
```
Isaac0-dev, line 11: why not just have this name field always come with a small buffer of it's own so you don't have to `strdup` it?

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+    int   ref;          // registry reference to the upvalue's value
+    const void *id;     // the opaque upvalue‐cell pointer
+    int type;           // the Lua type code of that value (LUA_T*)
+    struct UpvalRecord *next;
+} UpvalRecord;
+
+typedef struct UpvalReference {
+    char *name;
```
Isaac0-dev, line 19: same with this

src/pc/mods/mod.c
```diff
@@ -8,6 +8,39 @@
...
+    const uint64_t EPOCH_DIFF = 116444736000000000ULL; // 100-ns from 1601 to 1970
+    uint64_t time100ns = ull.QuadPart;
+    return (time100ns - EPOCH_DIFF) / 10000000ULL;    // to seconds
+#else
+    struct stat st;
+    if (stat(file->cachedPath, &st) != 0) {
+        // error; errno is set
+        return 0;
+    }
+    return (uint64_t)st.st_mtime;
```
Isaac0-dev: in this function, just use `u64` instead of `uint64_t` since you return `u64`

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
+#include <stdbool.h>
+#include "smlua.h"
+#include "pc/mods/mods.h"
+#include "pc/mods/mods_utils.h"
+
+#define LIVE_RELOAD_TICK_COUNT 15
+
+typedef struct UpvalRecord {
+    char *funcKeyStr;   // the table key under which the function lived
```
Isaac0-dev, line 9: could also make this a small buffer to avoid `strdup`

src/pc/lua/smlua_live_reload.c
```diff
@@ -0,0 +1,461 @@
...
+    int   ref;          // registry reference to the upvalue's value
+    const void *id;     // the opaque upvalue‐cell pointer
+    int type;           // the Lua type code of that value (LUA_T*)
+    struct UpvalRecord *next;
+} UpvalRecord;
+
+typedef struct UpvalReference {
+    char *name;
+    int reference;
+    int active;
```
Isaac0-dev: looks like `active` can be a bool

src/pc/lua/smlua_live_reload.c
djoslin0: string names aren't always guaranteed (functions aren't always named) and why use more memory than necessary?
the live-reload stuff is a single execution on an event (files being saved), as opposed to constantly running every frame. So reusing buffers or pools and such aren't worth it here imo.

src/pc/lua/smlua_live_reload.c
djoslin0: see other comment

src/pc/lua/smlua_live_reload.c
djoslin0: see other comment

Isaac0-dev (approved):
looks good! there are still minor nitpicks I have, but this pr is still fine if it's merged as is imo. The functionality looks sound, and I can see you were very careful in implementing it.
those minor nitpicks are:
- in `smlua_live_reload.c`, function names and static variables etc don't use a naming scheme, which looks a little weird. That said, we'd have some pretty long names if you added a naming scheme, and the stuff that doesn't have one are local to that file.
- it would still be nice to let users map the mod reload button to whatever they want.

PR #852, opened 2025-06-15
Add get_game_tick_counter()

djoslin0:
Coop now maintains a counter that increments at the start of each game tick,

This is to be able to identify specific frames regardless of mod load, hook,
and execution order.

### DIFF
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -104,6 +104,9 @@ u8 gLuaVolumeLevel = 127;
 u8 gLuaVolumeSfx = 127;
 u8 gLuaVolumeEnv = 127;
 
+u64 gFrameCounterGame = 0;
+u64 gFrameCounterRender = 0;
+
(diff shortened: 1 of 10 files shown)


src/pc/pc_main.c
```diff
@@ -219,6 +222,7 @@ void produce_interpolation_frames_and_delay(void) {
         );
         gRenderingDelta = delta;
 
+        gFrameCounterRender++;
```
Isaac0-dev: How is this useful? There's no hook during interpolated frames

djoslin0: hmm, it looks like you're right. For some reason I thought the geo hooks were triggered during interpolation

djoslin0: I can remove the render counter

Isaac0-dev: i'm not sure, it could be removed to actual game render time, but i guess the regular update counter would do that fine too

PR #855, opened 2025-06-17
Increase MaxPlayer limit

Xenthio:
It would be nice to have more than just 16 players.


Xenthio:
Changes might have to be made to coopnet for this to work

PR #856, opened 2025-06-19
Fix bug in fixCollisionBugsRoundedCorners, add floorNormalMinY, ceilNormalMaxY

djoslin0:
There was a bug in resolving wall collisions with fixCollisionBugsRoundedCorners enabled.
Due to the way a check was written, it would often ignore walls when it shouldn't have.

I've also added gLevelValues.floorNormalMinY and gLevelValues.ceilNormalMaxY
These will values will be used to determine which surfaces will become a wall, floor, or ceiling

### DIFF
--- a/autogen/lua_definitions/structs.lua
+++ b/autogen/lua_definitions/structs.lua
@@ -1068,6 +1068,7 @@
 
 --- @class LevelValues
 --- @field public bubbleOnDeathBarrierInCapStages integer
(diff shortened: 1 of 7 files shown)


djoslin0:
Allows you to override which things are walls, ceilings, or floors. With some careful tuning dynamic objects feel a lot more correct.
image

PR #859, opened 2025-06-19
Add implicit string conversion to function params

Beckowl:
This updates ``smlua_to_string`` to align with Lua behavior, where many built-in Lua functions that receive string params implicitly convert values to strings (e.g. both ``print(1)`` and ``print("1")`` output ``1``)

Now code like the following works without requiring a tostring call:
```lua

    djui_chat_message_create(m.forwardVel) -- No need to tostring it
end

```

I don’t like having to call tostring every time i want to log something, so this change simplifies that. Lazy moment.


Beckowl:
nevermind

PR #860, opened 2025-06-20
Add `smlua_text_utils_dialog_get_text` and some other dialog functions

Beckowl:
This PR adds three new functions to `smlua_text_utils`:

`smlua_text_utils_dialog_get(dialogId)`
- Returns the DialogEntry* for the given dialog ID.

`smlua_text_utils_dialog_is_replaced(dialogId)`
- Returns true if the dialog has been replaced by a mod.

`smlua_text_utils_dialog_get_text(dialogId)`
- Returns the ASCII string of the dialog’s content.

This allows mods to check if a dialog has been replaced without overriding the global ``smlua_text_utils_dialog_replace``

# A little question for the reviewers:
How can I add `smlua_text_utils_dialog_get_text` to the Lua definitions?
Since it’s manually bound (in order to `free()` the returned string), it doesn't get picked up by autogen.


Beckowl:
I didn’t want to use a fixed-size buffer because `smlua_text_utils_dialog_replace` doesn’t enforce any size limit on custom dialogs. What should i do if a custom dialog is larger than the buffer? truncate it???

Beckowl:
That's nice, but i'd still need to manually bind it to avoid a double push, so it doesn't really solve the autogen problem:

```c
int smlua_func_smlua_text_utils_dialog_get_text(lua_State* L) {
    if (L == NULL) { return 0; }

    int top = lua_gettop(L);
    if (top != 1) {
        LOG_LUA_LINE("Improper param count for '%s': Expected %u, Received %u", "smlua_text_utils_dialog_get_text", 1, top);
        return 0;
    }

    int dialogId = smlua_to_integer(L, 1);
    if (!gSmLuaConvertSuccess) { LOG_LUA("Failed to convert parameter %u for function '%s'", 1, "smlua_text_utils_dialog_get_text"); return 0; }

    lua_pushstring(L, smlua_text_utils_dialog_get_text(dialogId)); // string is pushed twice

    return 1;
}
```

PeachyPeachSM64:
interesting, but why not adding a `const char *text` field directly to `struct DialogEntry` so you can `dialog.text` after calling `smlua_text_utils_dialog_get`? it would avoid to have two different functions
(for custom dialog the ascii string is already available and for original dialogs, the ascii form could be computed during the loading of assets)

Beckowl:
At first, `smlua_text_utils_dialog_get_text` was intended to be a general purpose SM64 string to ASCII converter, that you can use for things like `get_level_name_sm64()` and `get_star_name_sm64()`, which return `u8*` that Lua can’t dereference.

Making it check if a pointer actually came from a dialog was getting a bit tricky, so i narrowed it to just dialog IDs. That said, I still think a generic converter could be useful later.

As for adding a text field to dialog entry, that's a good idea, I'll look into it

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -22,8 +22,37 @@ static bool sReplacedActName[(COURSE_RR+2)*6] = { 0 };
...
 #define INVALID_COURSE_NUM(courseNum) (smlua_level_util_get_info_from_course_num(courseNum) == NULL && !COURSE_IS_VALID_COURSE(courseNum))
 
+extern const struct { const char *str; u8 c; u8 menu; } sSm64CharMap[];
 void convert_string_sm64_to_ascii(char *strAscii, const u8 *str64);
 
+static size_t measure_converted_sm64_string(const u8* str64) {
+    size_t len = 0;
+    
+    for (size_t i = 0; str64[i] != 0xFF; i++) {
+        for (int j = 0; sSm64CharMap[j].str != NULL; j++) {
```
PeachyPeachSM64: s32 instead of int

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -191,19 +233,37 @@ void smlua_text_utils_dialog_replace(enum DialogId dialogId, UNUSED u32 unused,
...
     }
 
+    free(dialog->text);
+
     dialog->unused = unused;
     dialog->linesPerBox = linesPerBox;
     dialog->leftOffset = leftOffset;
     dialog->width = width;
     dialog->str = smlua_text_utils_convert(str);
+    dialog->text = get_dialog_text_ascii(dialog);
```
PeachyPeachSM64: make a copy of str instead of translating back from sm64 text

PR #865, opened 2025-06-21
Fix HOOK_ALLOW_HAZARD_SURFACE and expose `SURFACE_IS_*` macros

Sunketchupm:
The hooks now run just before the player is about to be affected by said hazard surface instead of in a random spot every frame. The way I handled quicksand is pretty bad however and I would like a better solution if possible.


Sunketchupm:
May as well

PR #866, opened 2025-06-21
Add bytestring packets for mods to use

djoslin0:
Adds network_send_bytestring(), network_send_bytestring_to(), and HOOK_ON_PACKET_BYTESTRING_RECEIVE

This is as close to raw bytes as you can get in lua. It allows you to efficiently pack as much data as possible into each packet.

The existing network_send() was built for ease of use, but is quite inefficient when you want to send a lot of data. Each number or int field in the table amounts to 9 bytes for the key and 9 bytes for the value. (amounting to 18 bytes per value). Whereas with the new method you can pack that information into 1, 2, 4, or 8 bytes depending on your type's size.


PeachyPeachSM64:
what's the size limit of such a packet?

djoslin0:
3000 bytes is the total max size per packet but subtract 4 for the bytestring header, and 10 for the normal packet header.
So 2986 is left for the data

src/pc/network/packets/packet_lua_custom.c
```diff
@@ -132,3 +130,117 @@ void network_receive_lua_custom(struct Packet* p) {
...
+
+  ////////////////
+ // bytestring //
+////////////////
+
+#define MAX_BYTESTRING_LENGTH (PACKET_LENGTH - 15)
+
+void network_send_lua_custom_bytestring(bool broadcast) {
+    lua_State* L = gLuaState;
+    s32 paramIndex = 1;
```
Isaac0-dev, line 142: why have `paramIndex`? its less readable than just a magic value as long as you know what that arg is.
From my perspective I find it easier to see a magic value for Lua api calls because I'm expecting it.

Isaac0-dev (approved):
looks good.

just an observation though, if you call these functions from a coroutine it'll break, using the stack from the main Lua state. Not a big problem as these aren't the only functions that have the issue in smlua.

src/pc/network/packets/packet_lua_custom.c
djoslin0: I was just following the pattern from `network_send_lua_custom()`
I can remove it if you want

Isaac0-dev: ah didn't notice that. nah don't worry about it

PR #868, opened 2025-06-22
Fix interactions

Sunketchupm:
Fixes this bug

https://github.com/user-attachments/assets/471ddc2f-3a10-4591-a592-b608d39aeb1d


PeachyPeachSM64:
I don't know what it's fixing, but that reintroduces the bug where custom punch/kick/trip actions that set the MARIO_PUNCHING/KICKING/TRIPPING flags don't return the right interaction type

PR #870, opened 2025-06-25
Lighting engine improvements

djoslin0:
* Now objects will be affected by the lighting engine.
This is accomplished by passing the renderer the object's model matrix (uncombined with the view or projection).

* You can now setup the lighting engine mode to affect all shaded surfaces.
* Lighting can be affected by surface normals.
* You can control what type of tone mapping is applied.

```
added le_set_mode(mode)
    By default we retain the previous behavior.

    When set to LE_MODE_AFFECT_ALL_SHADED the lighting engine will
    affect every shaded material.

    This way we don't have to recompile every object and level that
    we want shaded with special coop-specific commands
```

`added le_get_mode()`

```
added le_set_tone_mapping(toneMapping)
    Tone mapping is what happens when a color value exceeds its 0-255
    range.

    By default we retain the current tone mapping (called
    LE_TONE_MAPPING_TOTAL_WEIGHTED).

    LE_TONE_MAPPING_WEIGHTED is now accessible, it was the tone
    mapping that was previously left out of the compile through ifdefs.

    LE_TONE_MAPPING_CLAMP is just simple additive with a clamp at a
    color value of 255.

    LE_TONE_MAPPING_REINHARD is reinhard tone mapping
    (vout = (vin + 1) / vin).
```

```
added le_set_light_use_surface_normals(id, useSurfaceNormals)
    By default lights retain their previous behavior (of ignoring
    surface normals).

    When enabled lights cast on one side of the object will not
    appear on the other side of the object.

    It is kind of like backface culling, but for lights.
```

```
added le_calculate_lighting_color_with_normal(pos, normal, outColor, lightIntensityScalar)
    It's just like le_calculate_lighting_color(), but you can pass
    in normals now.
```


djoslin0:
Example video showing the new commands
https://www.youtube.com/watch?v=TJ_qnzi8YHA

src/pc/gfx/gfx_pc.c
```diff
@@ -814,10 +853,20 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
...
             }
 
-            if (rsp.geometry_mode & G_LIGHTING_ENGINE_EXT) {
+            if ((le_get_mode() == LE_MODE_AFFECT_ALL_SHADED) || (rsp.geometry_mode & G_LIGHTING_ENGINE_EXT)) {
                 Color color;
                 CTX_BEGIN(CTX_LIGHTING);
-                le_calculate_lighting_color(((Vtx_t*)v)->ob, color, 1.0f);
+
+                Vec3f vpos    = { v->ob[0], v->ob[1], v->ob[2] };
+                Vec3f vnormal = { vn->n[0] / 127.0f, vn->n[1] / 127.0f, vn->n[2] / 127.0f };
```
PeachyPeachSM64: from @ManIsCat2 

instead of vn->n[] use nx, ny, nz for normals, that's the vtx normals, and its modified when packed normals is used (packed normals uses normals, its essentially Vertex Colors and Lighting)

(defined at line 791)

PeachyPeachSM64 (changes requested):

AgentXLP:
image
image
If you don't create any lights, the lighting will be very broken by default for some reason. I think this might have been an issue before too but I'm not entirely sure.

I think I remember trying a global mode initially and trying to get vertex colored surfaces to use the lighting effect but it inadvertently made all of the HUD and menu elements also have it which led me to making a geometry mode. I wonder if there's a way around that.

Some suggestions:
* Make lights using surface normals a global setting or at least on by default, I'm okay with some default settings being changed around.
* Make lights affect objects even when in default mode, this is the behavior I intended originally but couldn't due to not knowing how to handle the object matrices.
* Replace the original lighting engine demo with your new bowling lighting demo, I find it really interesting and honestly a little more than my example map.

This is amazing work nonetheless and you managed to crack things like the object matrices that I couldn't figure out no matter how much I tried, good job!

djoslin0:
> If you don't create any lights, the lighting will be very broken by default for some reason. I think this might have been an issue before too but I'm not entirely sure.
Yeah this was an issue before too, I'll see if it's a quick fix or not

>I think I remember trying a global mode initially and trying to get vertex colored surfaces to use the lighting effect but it inadvertently made all of the HUD and menu elements also have it which led me to making a geometry mode. I wonder if there's a way around that.
Yeah that's tricky, I haven't looked too deeply into this side of things since I'm mainly trying to use the lighting engine in my custom level. So for now I went with only applying the lighting effects to shaded surfaces. If we try to apply them to unlit surfaces then the lighting engine will dim things that should be bright like lamps and such. I'm not sure what a good solution here would be. Unfortunately vanilla uses unshaded surfaces a lot in the levels, so this isn't ideal but I think it's a good compromise.

> Make lights using surface normals a global setting or at least on by default, I'm okay with some default settings being changed around.
I can do that, I just didn't want to potentially break anything you had cooking. I'll make it use surface normals by default with a toggle per light to disable the behavior. I did notice that vertex colored levels like the black mesa one have their surface normals replaced with color information. So in those cases the lights will still ignore normals.

>Make lights affect objects even when in default mode, this is the behavior I intended originally but couldn't due to not knowing how to handle the object matrices.
Will do.

> Replace the original lighting engine demo with your new bowling lighting demo, I find it really interesting and honestly a little more than my example map.
The bowling mod is a lot bigger than a lighting demo, it's a lua physics engine and gamemode. I'll see if I can find some time to make a more concise example.

> This is amazing work nonetheless and you managed to crack things like the object matrices that I couldn't figure out no matter how much I tried, good job!
Thanks! The object matrices thing was a gigantic pain in the ass. It took me a long time to figure out that I couldn't get the information out of the existing matrices in `gfx_pc.c`

djoslin0:
Coolio pointed out that my changes didn't properly light characters, so I fell back into matrix hell.
Now instead of storing each object's model matrix, we store the inverse of the camera matrix.
Then we strip the camera's view matrix off of the model-view matrix that we had in order to obtain the model matrix.

This means animated characters and nested objects will receive the correct lighting information now.

image

AgentXLP:
I think the tonemapping options are good, I wasn't sure whether to use weighted or total weighted because they looked similar while I was testing and just picked one and left the other under an ifdef since I didn't think people would care about it.
Weighted looks the best and I think making it a default would be good, if it messes up the lighting engine demo I can just make it use total weighted again.

src/pc/gfx/gfx_pc.c
djoslin0: done

djoslin0:
I think weighted looks a little bit better than total weighted because I didn't notice any dimming when I was testing that one. I'll use that as the default.

I think I've addressed all of the comments now 👍

src/engine/math_util.c
```diff
@@ -781,6 +781,63 @@ OPTIMIZE_O3 void mtxf_inverse(OUT Mat4 dest, Mat4 src) {
     mtxf_copy(dest, buf);
 }
 
+/**
+ * Compute the inverse of 'src' and put it into 'dest' but it can be a non-affine matrix.
+ */
+OPTIMIZE_O3 bool mtxf_inverse_non_affine(OUT Mat4 dest, Mat4 src) {
+    // augmented matrix [ src | I ]
+    f32 aug[4][8];
+    for (int i = 0; i < 4; i++) {
```
PeachyPeachSM64: change all int to s32

src/engine/math_util.c
```diff
@@ -781,6 +781,63 @@ OPTIMIZE_O3 void mtxf_inverse(OUT Mat4 dest, Mat4 src) {
...
+
+    // forward elimination
+    for (int k = 0; k < 4; k++) {
+        // find pivot row
+        int piv = k;
+        for (int i = k + 1; i < 4; i++) {
+            if (fabsf(aug[i][k]) > fabsf(aug[piv][k])) { piv = i; }
+        }
+
+        if (fabsf(aug[piv][k]) < 1e-6f) { return false; }
```
PeachyPeachSM64: FLT_EPSILON instead of 1e-6f

src/engine/math_util.c
```diff
@@ -781,6 +781,63 @@ OPTIMIZE_O3 void mtxf_inverse(OUT Mat4 dest, Mat4 src) {
     mtxf_copy(dest, buf);
 }
 
+/**
+ * Compute the inverse of 'src' and put it into 'dest' but it can be a non-affine matrix.
+ */
+OPTIMIZE_O3 bool mtxf_inverse_non_affine(OUT Mat4 dest, Mat4 src) {
```
PeachyPeachSM64, line 788: what algorithm is it? did you made it up?
could be good to add a reference if not

src/game/rendering_graph_node.c
```diff
@@ -609,6 +613,16 @@ static void geo_process_camera(struct GraphNodeCamera *node) {
         mtxf_copy(gCamera->mtx, gMatStack[gMatStackIndex]);
     }
 
+    // compute inverse matrix for lighting engine
+    if (le_is_enabled()) {
+        Mat4 invCameraMatrix;
+        if (mtxf_inverse_non_affine(invCameraMatrix, gCamera->mtx)) {
+            Mtx *invMtx = alloc_display_list(sizeof(Mtx));
+            mtxf_to_mtx(invMtx, invCameraMatrix);
+            gSPMatrix(gDisplayListHead++, VIRTUAL_TO_PHYSICAL(invMtx), G_MTX_INVERSE_CAMERA_EXT);
```
PeachyPeachSM64: VIRTUAL_TO_PHYSICAL: unused

src/pc/gfx/gfx_pc.c
```diff
@@ -627,7 +631,18 @@ static void calculate_normal_dir(const Light_t *light, Vec3f coeffs, bool applyL
 }
 
 static void OPTIMIZE_O3 gfx_sp_matrix(uint8_t parameters, const int32_t *addr) {
+
     Mat4 matrix;
+
+    // remember inverse camera matrix to use for the lighting engine
+    if (parameters == G_MTX_INVERSE_CAMERA_EXT) {
+        if (addr) {
+            memcpy(sInverseCameraMatrix, addr, sizeof(matrix));
```
PeachyPeachSM64: sizeof(sInverseCameraMatrix) would be better

src/pc/gfx/gfx_pc.c
```diff
@@ -678,6 +694,39 @@ static float gfx_adjust_x_for_aspect_ratio(float x) {
     return x * gfx_current_dimensions.x_adjust_ratio;
 }
 
+#include "src/game/camera.h"
```
PeachyPeachSM64: move the #include at the top of the file

AgentXLP:
https://github.com/user-attachments/assets/6aa81e9c-742f-472d-b90e-00d0ff6a18ad

I found a way to do lighting on vertex colored surfaces and without it looking bad too. HUD/menus and surfaces that don't use shade color (like using environment color) are unaffected.

On my local clone, I added a new LEMode:
```c
enum LEMode {
    LE_MODE_AFFECT_ALL_SHADED_AND_COLORED,
    LE_MODE_AFFECT_ALL_SHADED,
    LE_MODE_AFFECT_ONLY_GEOMETRY_MODE,
};
```
And made it the default, hopefully that name is good enough. If an author wants to have an unlit surface they can either:
* Use `LE_MODE_AFFECT_ALL_SHADED` instead
* Use `LE_MODE_AFFECT_ONLY_GEOMETRY_MODE` instead
* Use `gsSPVertexNonGlobal` instead of `gsSPVertex` so it can't be vertex colored by Lua or the lighting engine
* Use environment/prim color or anything in the combiner that isn't Shade Color

Also, I made `gsSPVertexNonGlobal` to make on screen elements and some other things not be affected by Lua vertex coloring and that's the crux of this implementation.

I can probably commit my changes after this is merged, but I'm gonna look into fixing the lighting being really broken by default if you don't spawn any lights in, that was an issue before as well I remember now and I want to finally fix that.

src/pc/gfx/gfx_pc.c
```diff
@@ -814,21 +863,37 @@ static void OPTIMIZE_O3 gfx_sp_vertex(size_t n_vertices, size_t dest_index, cons
...
             }
 
-            if (rsp.geometry_mode & G_LIGHTING_ENGINE_EXT) {
+            if (le_is_enabled() && ((le_get_mode() == LE_MODE_AFFECT_ALL_SHADED) || (rsp.geometry_mode & G_LIGHTING_ENGINE_EXT))) {
                 Color color;
                 CTX_BEGIN(CTX_LIGHTING);
-                le_calculate_lighting_color(((Vtx_t*)v)->ob, color, 1.0f);
+
+                Vec3f vpos    = { v->ob[0], v->ob[1], v->ob[2] };
+                Vec3f vnormal = { nx / 127.0f, ny / 127.0f, nz / 127.0f };
```
PeachyPeachSM64: does vnormal need to be of length 1?
because vertex normal is not always of length 127, so vec3f_normalize might be needed

PeachyPeachSM64 (changes requested):

src/engine/math_util.c
djoslin0: It's Gauss-Jordan elimination, I'll add it to the comment

src/pc/gfx/gfx_pc.c
djoslin0: Oops, this was a leftover from trying a bunch of different things. I'll remove it

src/pc/gfx/gfx_pc.c
djoslin0: The normal doesn't need to be normalized at that step since we normalize it in `le_calculate_lighting_color_with_normal()`. I'll remove the division here since it isn't needed anymore.

djoslin0:
@AgentXLP Ooo, very interesting. If your method is only affecting surfaces that use `shade` in their color combiners, I think it could replace my new mode too. I want to ensure that there is a way to keep full-bright stuff, but all of my full-bright stuff doesn't use shade... and I can't think of something that would.

djoslin0:
I do think we should probably introduce something like `le_engine_enable()` to enable or disable the lighting engine. Otherwise vanilla will be constantly trying to call lighting engine stuff, when lighting engine stuff is very much not the norm. But we can think that through after the changes merge 😄

AgentXLP:
Alright, I'll just make it work under `LE_MODE_AFFECT_ALL_SHADED` and remove the other one.
And yeah, adding a function to enable or disable it cleanly would be better than trying to handle it automatically imo.

Also, I went about implementing the lighting for vertex colored geometry in an interesting way.
I wanted to retain the color on things like red coins so I made not using the geometry mode multiply the vertex color over the calculated lighting and I made using the geometry mode factor the vertex color into the lighting.
There's probably a better way to go about this, but red coins stay red and the dark spot in the Sunset Canals demo map lights up too.
```c
        // if lighting engine is enabled and we should affect all vertex colored surfaces or the lighting engine geometry mode is on
        } else if (le_is_enabled() && (affectAllVertexColored || (rsp.geometry_mode & G_LIGHTING_ENGINE_EXT))) {
            Color color;
            CTX_BEGIN(CTX_LIGHTING);

            Vec3f vpos = { v->ob[0], v->ob[1], v->ob[2] };

            // transform vpos to world space
            gfx_local_to_world_space(vpos, NULL);

            // do multiplication based lighting instead of additive based lighting if we're not using the lighting engine geometry mode,
            // this is my compromise for retaining vertex colors vs lighting up darker surfaces.
            // if retaining color is the most important like on a red coin, don't use the lighting engine geometry mode
            // if lighting up darker surfaces like in a map with prebaked lighting is the most important, use the lighting engine geometry mode
            if (affectAllVertexColored && !(rsp.geometry_mode & G_LIGHTING_ENGINE_EXT)) {
                le_calculate_lighting_color(vpos, color, 1.0f);
            } else {
                le_calculate_vertex_lighting((Vtx_t*)v, vpos, color);
            }

            CTX_END(CTX_LIGHTING);

            // combine the colors
            if (affectAllVertexColored && !(rsp.geometry_mode & G_LIGHTING_ENGINE_EXT)) {
                d->color.r = (v->cn[0] * color[0] / 255.0f) * vertexColorCached[0];
                d->color.g = (v->cn[1] * color[1] / 255.0f) * vertexColorCached[1];
                d->color.b = (v->cn[2] * color[2] / 255.0f) * vertexColorCached[2];
            } else {
                if (luaVertexColor) {
                    d->color.r = color[0] * vertexColorCached[0];
                    d->color.g = color[1] * vertexColorCached[1];
                    d->color.b = color[2] * vertexColorCached[2];
                } else {
                    d->color.r = color[0];
                    d->color.g = color[1];
                    d->color.b = color[2];
                }
            }
```

PeachyPeachSM64 (approved):

PR #873, opened 2025-06-28
Regenerate DynOS assets when source files are modified

djoslin0:
Previously to recompile DynOS assets you had to remove the bin/lvl/bhv/tex/col/etc files. Now the game will compare the last-modified-timestamps of the generated/compiled assets vs the other files that are within those directories. If the presumed- source files have a later modified timestamp DynOS will regenerate those assets.

While this results in scanning the attributes of files more, it also prevents parsing files unnecessarily. Previously actors would always parse their source files and build up GfxData unnecessarily.

### DIFF
--- a/data/dynos_bin_col.cpp
+++ b/data/dynos_bin_col.cpp
@@ -684,19 +684,24 @@ DataNode<Collision>* DynOS_Col_LoadFromBinary(const SysPath &aFilename, const ch
  // Generate //
 //////////////
 
+static String GetActorFolder(const Array<Pair<u64, String>> &aActorsFolders, u64 aModelIdentifier) {
+    for (const auto &_Pair : aActorsFolders) {
+        if (_Pair.first == aModelIdentifier) {
+            return _Pair.second;
+        }
+    }
+    return String();
+}
+
 void DynOS_Col_Generate(const SysPath &aPackFolder, Array<Pair<u64, String>> _ActorsFolders, GfxData *_GfxData) {
     for (auto &_ColNode : _GfxData->mCollisions) {
         String _ColRootName = _ColNode->mName;
-
-        // If there is an existing binary file for this collision, skip and go to the next actor
         SysPath _ColFilename = fstring("%s/%s.col", aPackFolder.c_str(), _ColRootName.begin());
-        if (fs_sys_file_exists(_ColFilename.c_str())) {
 
-            // Compress file to gain some space
-            if (configCompressOnStartup && !DynOS_Bin_IsCompressed(_ColFilename)) {
-                DynOS_Bin_Compress(_ColFilename);
-            }
-
+        // If there is an existing binary file for this collision, skip and go to the next collision
+        String _ActorFolder = GetActorFolder(_ActorsFolders, _ColNode->mModelIdentifier);
+        SysPath _SrcFilename = fstring("%s/%s/collision.inc.c", aPackFolder.c_str(), _ActorFolder.begin());
+        if (DynOS_GenFileExistsAndIsNewerThanFile(_ColFilename, _SrcFilename)) {
             continue;
         }
 
--- a/data/dynos_bin_utils.cpp
+++ b/data/dynos_bin_utils.cpp
@@ -1,5 +1,157 @@
 #include "dynos.cpp.h"
 
+  ////////////////////////////////
+ // Should-generate-pack logic //
+////////////////////////////////
+
+static bool DynOS_PathHasExtension(const char *aPath, const char *aExtension) {
+    size_t _LenStr    = strlen(aPath);
+    size_t _LenSuffix = strlen(aExtension);
+
+    if (_LenSuffix > _LenStr) {
+        return false;
+    }
+
+    return strcmp(aPath + (_LenStr - _LenSuffix), aExtension) == 0;
+}
+
+static bool DynOS_PathHasExtensions(const char *aPath, const char * const aExtensions[], s32 aExtensionCount) {
+    for (s32 i = 0; i < aExtensionCount; i++) {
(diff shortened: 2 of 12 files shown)


djoslin0:
Example video showing it in action:
https://www.youtube.com/watch?v=i_kLuKzFFbI

The video shows detection and auto-recompiling dynos for levels and actors. It does it for textures and behaviors too though.

src/pc/mods/mod.c
```diff
@@ -17,31 +17,6 @@
...
-    return (time100ns - EPOCH_DIFF) / 10000000ULL;    // to seconds
-#else
-    struct stat st;
-    if (stat(file->cachedPath, &st) != 0) {
-        // error; errno is set
-        return 0;
-    }
-    return (u64)st.st_mtime;
-#endif
-}
```
PeachyPeachSM64, line 43: please keep the comments in fs_sys_get_modified_time so we don't have to guess (or google) what the magic number is

djoslin0: alright I'll add them back in

djoslin0: done

data/dynos_bin_utils.cpp
```diff
@@ -1,5 +1,157 @@
...
+    size_t _LenSuffix = strlen(aExtension);
+
+    if (_LenSuffix > _LenStr) {
+        return false;
+    }
+
+    return strcmp(aPath + (_LenStr - _LenSuffix), aExtension) == 0;
+}
+
+static bool DynOS_PathHasExtensions(const char *aPath, const char * const aExtensions[], s32 aExtensionCount) {
```
Isaac0-dev: you can use an std::vector to remove the need to provide the array count as a parameter here

data/dynos_bin_utils.cpp
```diff
@@ -1,5 +1,157 @@
...
+    for (s32 i = 0; i < aExtensionCount; i++) {
+        if (DynOS_PathHasExtension(aPath, aExtensions[i])) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+static void DynOS_GetMTimeInFolderSplitByExtensions(const SysPath &aPath, const char * const aExtensions[], s32 aExtensionCount, u64 *aLatestMTimeExt, u64 *aLatestMTimeNonExt) {
```
Isaac0-dev: same here

data/dynos_bin_utils.cpp
```diff
@@ -1,5 +1,157 @@
...
+                _LatestMTimeSubDir = MAX(_LatestMTimeSubDir, _PathMTime);
+            }
+        }
+        closedir(_DirPath);
+    }
+
+    return _LatestMTimeSubDir;
+}
+
+bool DynOS_ShouldGeneratePack(const SysPath &aPackFolder, const char * const aGenFileExtensions[], s32 aGenFileExtensionCount) {
```
Isaac0-dev: same here

data/dynos_bin_utils.cpp
```diff
@@ -1,5 +1,157 @@
...
+        if (DynOS_PathHasExtension(aPath, aExtensions[i])) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+static void DynOS_GetMTimeInFolderSplitByExtensions(const SysPath &aPath, const char * const aExtensions[], s32 aExtensionCount, u64 *aLatestMTimeExt, u64 *aLatestMTimeNonExt) {
+    DIR *_DirPath = opendir(aPath.c_str());
```
Isaac0-dev, line 29: these functions, `DynOS_GetMTimeInFolder*` all seem to have very similar logic. Can you move it to one static helper function where possible?

data/dynos_bin_col.cpp
```diff
@@ -684,19 +684,24 @@ DataNode<Collision>* DynOS_Col_LoadFromBinary(const SysPath &aFilename, const ch
  // Generate //
 //////////////
 
+static String GetActorFolder(const Array<Pair<u64, String>> &aActorsFolders, u64 aModelIdentifier) {
```
Isaac0-dev: rather than duplicating `GetActorFolder`, move it to the utils file

Isaac0-dev (changes requested):

data/dynos_bin_utils.cpp
djoslin0: I'm pretty sure std::vector would use the heap here which seems unnecessary when the arrays and array counts are static

data/dynos_bin_utils.cpp
djoslin0: While they all accumulate modified times, they accumulate them in different ways. The only thing that could be common would be the directory walking code, but I think that would make the code harder to read

djoslin0:
I can make the other two changes if required, just wanted to give my thoughts first

data/dynos_bin_utils.cpp
Isaac0-dev: ah, good point. here's 2 alternatives that shouldn't:
- `std::array`
- `std::initializer_list`

data/dynos_bin_utils.cpp
Isaac0-dev: thats true, i didn't look closely enough and it'd look weird to attempt it anyway

data/dynos_bin_utils.cpp
Isaac0-dev: @djoslin0 can these alternatives be used?

djoslin0: Oops, I didn't notice the in-thread-reply.
I've added initializer_list to dynos and replaced the array count passing for the new functions

djoslin0:
Added initializer_list

PR #875, opened 2025-06-28
Add chat scaling and opacity

Blockyyy:
This adds a way to change the chat opacity and its scaling
While i was at it i also made the chat support the djui themes

Languages are a work in progress so it would be appreciated if people speaking a supported language could help me complete this setting, if nobody answers hopefully using a translator will be enough (so i got English, French, Spanish, Italian, Dutch and Portuguese)

https://github.com/user-attachments/assets/a9a463b3-4e8c-4780-83c2-3dad80a261d6

### DIFF
--- a/lang/Portuguese.ini
+++ b/lang/Portuguese.ini
@@ -185,6 +185,11 @@ MARIO_THEME = "Mario"
 ODYSSEY_THEME = "Odyssey"
 FILE_SELECT_THEME = "Seletor de Arquivo"
 
+[CHAT_DISPLAY]
+TEXT_OPACITY = "Opacidade de texto no chat"
+BACKGROUND_OPACITY = "Opacidade do fundo do chat"
+CHAT_SCALE = "Escala de chat"
+
 [DYNOS]
 DYNOS = "DYNOS"
 LOCAL_PLAYER_MODEL_ONLY = "Apenas modelo do jogador local"
--- a/lang/Dutch.ini
+++ b/lang/Dutch.ini
@@ -185,6 +185,11 @@ MARIO_THEME = "Mario"
 ODYSSEY_THEME = "Odyssey"
 FILE_SELECT_THEME = "Bestand selecteren"
 
+[CHAT_DISPLAY]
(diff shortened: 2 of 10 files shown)


lang/Portuguese.ini
```diff
@@ -185,6 +185,11 @@ MARIO_THEME = "Mario"
 ODYSSEY_THEME = "Odyssey"
 FILE_SELECT_THEME = "Seletor de Arquivo"
 
+[CHAT_DISPLAY]
+TEXT_OPACITY = "Opacidade de texto no chat"
+BACKGROUND_OPACITY = "Opacidade do fundo do chat"
+CHAT_SCALE = "Escala de chat"
```
mingokrb:
```suggestion
TEXT_OPACITY = "Opacidade do texto do chat"
BACKGROUND_OPACITY = "Opacidade do fundo do chat"
CHAT_SCALE = "Tamanho do chat"
```
there you go

lang/German.ini
```diff
@@ -185,6 +185,11 @@ MARIO_THEME = "Mario"
 ODYSSEY_THEME = "Odyssey"
 FILE_SELECT_THEME = "Dateiauswahl"
 
+[CHAT_DISPLAY]
+TEXT_OPACITY = "Chat-Textopazität"
+BACKGROUND_OPACITY = "Opazität des Chat-Hintergrunds"
+CHAT_SCALE = "Chat-Skala"
```
DM-kun: I think "Chat-Größe" would work better here

DMMaster636 (changes requested):

lang/German.ini
Blockyyy: i would like to fix it but i have no idea what that means

DM-kun: if you directly translate it, it'll give you "chat size"
while "Chat-Skala" does work, it seems a little unfitting for the context, which is why "Chat-Größe" makes more sense here

the rest of it seems fine i guess, that's how i would at least translate it
(i am german btw, just wanted to mention that)

PR #878, opened 2025-06-29
Fix warping from one custom level to another

djoslin0:
The explanation is in the code comments, but previously we only used one slot for custom levels. So when you warped from one to another it would override the current custom level's warp information with the one being loaded. Now we use two slots to avoid that.

> 
>         // This requires some explaination...
>         // It's a bit of a hack but it works.
> 
>         // DynOS's arrays for level information are LEVEL_MAX in size.
>         // LEVEL_MAX is based on the max number of vanilla levels.
> 
>         // So when we want to warp to a custom level we load the
>         // data into a slot that vanilla left unused.
> 
>         // We need two unused slots because we may warp from one
>         // custom level to another.
>

### DIFF
--- a/data/dynos_level.cpp
+++ b/data/dynos_level.cpp
@@ -381,7 +381,7 @@ void DynOS_Level_ParseScript(const void *aScript, s32 (*aPreprocessFunction)(u8,
 // Level Script Utilities
 //
 
-static s32 mCustomLevelSlot[2] = { 0 };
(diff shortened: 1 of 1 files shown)


Isaac0-dev:
i agree that it's a hack but the good news is I can't think of any situations this would actually break something. 
It's not very future proof, should 	someone decide to add custom level support for `DynOS_Level_GetCollision`, but I think it's fine.

PR #882, opened 2025-07-02
Translation error fixed  Spanish.ini

andromux:
Translation correction: "Tirantes" → "Overol"

The previous translation "tirantes" (suspenders/straps) was incorrect. Mario wears "overol" (overalls) - the complete garment consisting of pants with a bib front and suspenders. "Tirantes" refers only to the shoulder straps themselves, not the entire clothing item.


eros71-dev:
Or "Mono" depending on the country/region. Mostly was chosen as a neutral word, though you are correct.

PR #883, opened 2025-07-05
Mirrors

PeachyPeachSM64:
Fix mirror room and add mirror objects and features.

More info [here](todo).

Also come some changes and improvements:
- `Vec2` functions
- Geometry mode `G_CULL_INVERT_EXT`: invert culling when set (front becomes back and vice versa)
- `HOOK_MIRROR_MARIO_RENDER` is now deprecated and this hook doesn't run anymore
- Player lights (color parts) are no longer shared between objects: Models that use `geo_mario_set_player_colors` geo asm still pull colors from networkplayer palette, but if the palette is modified during rendering, only the graph nodes that have not been processed yet are affected by the changes. Thanks to that change, in the example mod, the player and their reflection can have  different palettes in the mirror room while "Strange reflections" is enabled.

Some preview of the example mod:

|mirror1|mirror2|
|-|-|

https://github.com/user-attachments/assets/b778a400-a8bf-4149-879d-304c11b31252

**Note**: the example mod requires a file `dbg.lua` that doesn't come with this PR. This file is a WIP that belongs to @djoslin0. This file isn't mandatory to play the mod, but the "Show mirror hitboxes" mod menu element won't work without it.


ThePlayerRolo:
YO THAT EXAMPLE IS SO COOL. WE GOT SOME DS type stuff

Baconator2558:
`G_CULL_INVERT_EXT` Siiiiiick

Isaac0-dev:
This is an awesome feature thanks for adding it to coop!
Can we have an option to toggle between vanilla mirror and enhanced mirror, while still having the enhanced mirror available to mods without an override?
Why remove the hook? There are actually plenty of mods that use that hook.

PeachyPeachSM64:
> Very cool!
> 
> You're allowed to distribute `dbg.lua` with mods if you want. I've semi-released it [here](https://github.com/djoslin0/coop-lua-libraries/tree/main/dbg) (but haven't posted publicly as I'm waiting for the next coop release) The linked one has an updated header comment.

Thanks! I'll update the example to include the dbg lib

Isaac0-dev:
i'm noticing some flashing in the mirror when i roll in OMM here. i think it's Mario's model.

here's some screenshots, but please do test it yourself to really get an idea of it:
image


frame 1 (what it looks like before):
Pasted image

frame 2:
Pasted image (2)

frame 3:
Pasted image (4)

and then it's gone.

docs/lua/examples/mirrors-demo/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Finally, the global setting `gBehaviorValues.MirrorsMaxReflections` controls how many times a real object can be reflected through mirrors. Defaults to `8`.
+
+<br>
+
+## Limitations
+
+- Only objects are reflected through mirrors. Level geometry must be manually reflected by the level designer to create the illusion of mirroring.
+- Shadows are not reflected by vertical mirrors. Shadow code is too hard to deal with.
+- Supports frame interpolation, to some extent. Don't stack too much mirrors in the same place.
```
Isaac0-dev, line 72: > Supports frame interpolation, to some extent. Don't stack too much mirrors in the same place.

Is this comment still true? I assume this was the same issue that was fixed by the geo interp pr?

src/game/mirror.c
```diff
@@ -0,0 +1,641 @@
...
+    struct Object obj[1];
+    struct Object *mirror;
+    struct MirrorObject *next;
+    struct MirrorObject *children;
+    u8 reflection;
+    bool rendered;
+};
+
+static struct GrowingArray *sMirrorObjectPool = NULL;
+static struct MirrorObject *sMirrorObjects[OBJECT_POOL_CAPACITY] = {[0 ... OBJECT_POOL_CAPACITY - 1] = NULL};
```
Isaac0-dev, line 76: this is confusing programming when these two do the same thing:

```C
static struct MirrorObject *sMirrorObjects[OBJECT_POOL_CAPACITY];
static struct MirrorObject *sMirrorObjects[OBJECT_POOL_CAPACITY] = { 0 };
```

src/game/mirror.c
```diff
@@ -0,0 +1,641 @@
...
+    cur_obj_set_behavior(bhvMirror);
+    cur_obj_hide();
+}
+
+  ////////////////////
+ // Mirror objects //
+////////////////////
+
+struct MirrorObject {
+    struct Object obj[1];
```
Isaac0-dev, line 67: what? why not `struct Object obj`?
This implies that you're trying to do something fancy with `obj` here, but it doesn't look like you are!

src/game/mirror.c
```diff
@@ -0,0 +1,641 @@
...
+void geo_process_mirrors(struct Object *obj) {
+
+    // Reject inactive objects
+    // Invisible objects can still be mirrored (for example to make an object visible only in the mirror)
+    if (!(obj->header.gfx.node.flags & GRAPH_RENDER_ACTIVE) || obj->header.gfx.sharedChild == NULL) {
+        return;
+    }
+
+    // Reject mirrors, lights and camera lakitu
+    if (obj->behavior == bhvMirror ||
```
Isaac0-dev, line 511: you could use a switch case here

src/game/mirror.c
```diff
@@ -0,0 +1,641 @@
...
+    struct MirrorObject *next;
+    struct MirrorObject *children;
+    u8 reflection;
+    bool rendered;
+};
+
+static struct GrowingArray *sMirrorObjectPool = NULL;
+static struct MirrorObject *sMirrorObjects[OBJECT_POOL_CAPACITY] = {[0 ... OBJECT_POOL_CAPACITY - 1] = NULL};
+
+INLINE static s32 get_object_slot(struct Object *obj) {
```
Isaac0-dev, line 78: This is a general utility function and I think it belongs somewhere else.

Isaac0-dev (commented):
I know it's wip, just wanted to take a look so here's some things I noticed.

I don't agree with entirely removing the original mirror code. While your implementation is fundamentally better than the original, the original had issues with it that were visible in the vanilla game anyway. You couldn't see particles from Mario correctly reflected in the mirror. 
So adding this as a compulsory feature seems like we're making coop be an enhanced version that doesn't let you play as though it were the original game.
I just think it should be an option. This should definitely be enabled by default, just having the option is my point. 
Furthermore, keeping `HOOK_MIRROR_MARIO_RENDER` seems a logical followup if it's going to be an option. Just talking about mod compatibility here. 

Of course I'm not completely against removing the old ways since this is absolutely better. But I just wanted to throw it out there. Please consider it.

docs/lua/examples/mirrors-demo/README.md
PeachyPeachSM64: yeah it needs updating

src/game/mirror.c
PeachyPeachSM64: it's explicitely setting all elements to NULL
but since NULL is 0 anyway, I'll change that

src/game/mirror.c
PeachyPeachSM64: yeah sorry, some trick I like to allow me to use obj as a pointer
I'll change that

src/game/mirror.c
PeachyPeachSM64: I'm pretty sure you can't switch case on pointers
also some of them have extra requirements, like bhvCloud and bhvCloudPart

src/game/mirror.c
PeachyPeachSM64: I agree

src/game/mirror.c
Isaac0-dev: Yeah, you can cast to uintptr_t, but at that point it's messy. Ignore my comment

PR #887, opened 2025-07-08
[WIP] User input for mods

Sunketchupm:
This adds in HOOK_ON_KEY_PRESSED and HOOK_ON_KEY_RELEASED. I considered making this a function similar to the mouse input functions we have and have tried it, but encountered problems where the passed in inputs overrides the previous passed in inputs. It may be better to make this its own special hook function instead of another event hook.

What still needs to be solved here is converting the passed in scancode into characters. I tried using the lua utf8 library for this but that does not seem to work.


Sunketchupm:
I've been thinking about this for a little while and I have no idea how to pull it off

PR #888, opened 2025-07-08
Fix buffer overruns in update checker's network-facing code.

bradleycha:
Pull request #885, rebased against and targetting the ```dev``` branch.

### DIFF
--- a/src/pc/update_checker.c
+++ b/src/pc/update_checker.c
@@ -49,8 +49,10 @@ void parse_version(const char *data) {
     u8 len = strlen(VERSION_IDENTIFIER);
     version += len;
     const char *end = strchr(version, '"');
-    memcpy(sRemoteVersion, version, end - version);
-    sRemoteVersion[end - version] = '\0';
+    size_t version_length = (size_t)(end - version);
+    if (version_length > sizeof(sRemoteVersion) - 1) { return; }
+    memcpy(sRemoteVersion, version, version_length);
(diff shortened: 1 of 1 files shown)


bradleycha:
After further testing, I discovered commit 01eb5c3 correctly downloads the entire file on Linux, which lets data be any arbitrary length.  I was able to easily overwrite important data and get SIGSEGV, and the only modifications to the code made were to use http instead of https for local testing, but this could easily be circumvented in a real attack.

src/pc/update_checker.c
```diff
@@ -49,8 +49,10 @@ void parse_version(const char *data) {
     u8 len = strlen(VERSION_IDENTIFIER);
     version += len;
     const char *end = strchr(version, '"');
-    memcpy(sRemoteVersion, version, end - version);
-    sRemoteVersion[end - version] = '\0';
+    size_t version_length = (size_t)(end - version);
+    if (version_length > sizeof(sRemoteVersion) - 1) { return; }
```
PeachyPeachSM64: please use camelCase for variable names

PeachyPeachSM64 (changes requested):

src/pc/update_checker.c
djoslin0: I recommend merging, we can fix the variable names after if it's a big enough deal. Idk if bradley is coming back

PeachyPeachSM64 (approved):

PR #891, opened 2025-07-10
Implement Allocate_Sequence()

ThePlayerRolo:
With the Help of ManIsCat2, This function allows mods to set there sequence id to a new one, similar to allocate_mario_action. This way no mod's  sequence ids will overlap each other.

Heres an Example Of How it works:

```lua
-- name: Allocate Sequence Mod Test
-- descripton: For Testing


local MY_SEQ = allocate_sequence()
local MY_SEQ2 = allocate_sequence()

log_to_console(tostring(MY_SEQ), CONSOLE_MESSAGE_INFO) --Prints 36
log_to_console(tostring(MY_SEQ2), CONSOLE_MESSAGE_INFO) --Prints 37

smlua_audio_utils_replace_sequence(MY_SEQ, 0x25, 74, "BooWaltz") --Set this to a M64  In Sounds
smlua_audio_utils_replace_sequence(MY_SEQ2, 0x25, 74, "CybersWorld") --Set this to a M64  In Sounds


    if m.action  ==  ACT_GROUND_POUND then
        if (m.controller.buttonPressed & B_BUTTON) ~= 0 then 
            play_cap_music(MY_SEQ2)
        end 
        if (m.controller.buttonPressed & A_BUTTON) ~= 0 then 
            play_cap_music(MY_SEQ)
        end 
    end 
end

hook_event(HOOK_MARIO_UPDATE, mario_update)```

https://youtu.be/Tbuuk2IeKps

### CODE REVIEW

File: src/pc/lua/utils/smlua_audio_utils.c [hunk 1]
```diff
@@ -4,6 +4,8 @@
 #define STB_VORBIS_HEADER_ONLY
 #include "pc/utils/stb_vorbis.c"
 
+#define MAX_CUSTOM_SEQS 1024
```
PeachyPeachSM64: no, that won't work
sequences are already limited by MAX_AUDIO_OVERRIDE
don't define a new constant when one already exists for the same purpose

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -29,6 +31,8 @@ struct AudioOverride {
 
 struct AudioOverride sAudioOverrides[MAX_AUDIO_OVERRIDE] = { 0 };
 
+static u32 sCustomSeqsCount = 0;
```
PeachyPeachSM64: why not start at SEQ_COUNT?
then you don't have to account for vanilla sequences
I would also suggest to rename it to `sCustomSeqIndex`
u32 is pointless, sequenceIds are u8

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -619,3 +624,12 @@ void smlua_audio_custom_deinit(void) {
...
 }
+
+
+u32 allocate_sequence(void) {
+    if ((sCustomSeqsCount + SEQ_COUNT) < MAX_CUSTOM_SEQS) {
+        return (++sCustomSeqsCount + SEQ_COUNT);   
+    }
+    LOG_ERROR("Cannot allocate more custom sequences.");
+    return 0;
+}
```
PeachyPeachSM64: see comment above
also u32 is pointless, sequenceIds are u8
btw increment is wrong, it should be varname++

src/pc/lua/utils/smlua_audio_utils.h
```diff
@@ -76,4 +76,7 @@ void audio_custom_update_volume(void);
 void audio_custom_shutdown(void);
 void smlua_audio_custom_deinit(void);
 
+/* |description|Allocates an `audio` sequence ID|descriptionEnd| */
+u32 allocate_sequence(void);
```
PeachyPeachSM64: move this to below `smlua_audio_utils_replace_sequence`
I would also suggest to rename it to `smlua_audio_utils_allocate_sequence` to keep consistent naming
and remove back quotes around audio, they are usually here to indicate a parameter name

PeachyPeachSM64 (changes requested):

ThePlayerRolo:
Should be ready for review again

src/pc/lua/utils/smlua_audio_utils.h
```diff
@@ -8,6 +8,8 @@ void smlua_audio_utils_reset_all(void);
 bool smlua_audio_utils_override(u8 sequenceId, s32* bankId, void** seqData);
 /* |description|Replaces the sequence corresponding to `sequenceId` with one called `m64Name`.m64 with `bankId` and `defaultVolume`|descriptionEnd| */
 void smlua_audio_utils_replace_sequence(u8 sequenceId, u8 bankId, u8 defaultVolume, const char* m64Name);
+/* |description|Allocates an `audio` sequence ID|descriptionEnd| */
```
PeachyPeachSM64: change description to `Allocates a new sequence ID`

PeachyPeachSM64 (changes requested):

ThePlayerRolo:
Should be ready for review again

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -154,6 +157,14 @@ void smlua_audio_utils_replace_sequence(u8 sequenceId, u8 bankId, u8 defaultVolu
     LOG_LUA_LINE("Could not find m64 at path: %s", m64path);
 }
 
+u8 smlua_audio_utils_allocate_sequence(void) {
+    if (sCustomSeqIndex< MAX_AUDIO_OVERRIDE) {
```
AgentXLP: Needs a space

AgentXLP (approved):
Looks good

PeachyPeachSM64:
Now that I think about it, what's the point of this?
As far as I know, nothing prevents doing something like `smlua_audio_utils_replace_sequence(0x60, bank, volume, "whatever")`, even if sequence 0x60 doesn't exist
There is no allocation, it's already allocated, but "unused"
If the point is to find an empty slot, then something like `smlua_audio_utils_find_unused_sequence` that would look for the first `override->enabled == false`, starting from SEQ_COUNT

PeachyPeachSM64:
> > Now that I think about it, what's the point of this? As far as I know, nothing prevents doing something like `smlua_audio_utils_replace_sequence(0x60, bank, volume, "whatever")`, even if sequence 0x60 doesn't exist There is no allocation, it's already allocated, but "unused" If the point is to find an empty slot, then something like `smlua_audio_utils_find_unused_sequence` that would look for the first `override->enabled == false`, starting from SEQ_COUNT would fit better
> 
> would we like loop from SEQ_COUNT to MAX_AUDIO_OVERRIDES in a for loop? If so, then the sCustomSeqCount varaible probably isn't needed from what i understand

yes in that case the sCustomSeqCount variable isn't needed anymore
since you'll be looping from SEQ_COUNT (included) to MAX_AUDIO_OVERRIDES (excluded) and return the first index where override->enabled is false

ThePlayerRolo:
> yes in that case the sCustomSeqCount variable isn't needed anymore since you'll be looping from SEQ_COUNT (included) to MAX_AUDIO_OVERRIDES (excluded) and return the first index where override->enabled is false

what index are you reffering to? is there like a table of audios?

ThePlayerRolo:
> yeah it's sAudioOverrides I mean something like this should work
> 
> ```c
> for (u8 seqId = SEQ_COUNT; seqId < MAX_AUDIO_OVERRIDES; seqId++) {
>     if (!sAudioOverrides[seqId].enabled) {
>         return seqId;
>     }
> }
> ```

lemme try that

PeachyPeachSM64:
> > Now that I think about it, what's the point of this? As far as I know, nothing prevents doing something like `smlua_audio_utils_replace_sequence(0x60, bank, volume, "whatever")`, even if sequence 0x60 doesn't exist There is no allocation, it's already allocated, but "unused" If the point is to find an empty slot, then something like `smlua_audio_utils_find_unused_sequence` that would look for the first `override->enabled == false`, starting from SEQ_COUNT would fit better
> 
> What if a mod that uses that function loads before a mod that doesn't, and the second mod replaces the sequence from the first? Not sure how common this would be, but it's something I think is worth considering.

you can't do anything about that, unless you tell `smlua_audio_utils_replace_sequence` to throw an error if the sequence id you pass is already used
but that would probably break rom-hacks or anything that replaces vanilla sequences

PeachyPeachSM64:
> > you can't do anything about that, unless you tell `smlua_audio_utils_replace_sequence` to throw an error if the sequence id you pass is already used but that would probably break rom-hacks or anything that replaces vanilla sequences
> 
> So then wouldn't it be better if you could allocate a new sequence that other mods can't replace? Unless I'm misunderstanding the original concept of this PR.

That would be ideal, a separate space where you could actually allocate sequences
Then ids would be unique and mods couldn't steal these slots without calling smlua_audio_utils_allocate_sequence first

ThePlayerRolo:
> > > yes in that case the sCustomSeqCount variable isn't needed anymore since you'll be looping from SEQ_COUNT (included) to MAX_AUDIO_OVERRIDES (excluded) and return the first index where override->enabled is false
> > 
> > 
> > what index are you reffering to? is there like a table of audios? EDIT: I think i found it: sAudioOverrides[] would we have i be sAudioOverrides? and if override is not enabled return that value?
> 
> yeah it's sAudioOverrides I mean something like this should work
> 
> ```c
> for (u8 seqId = SEQ_COUNT; seqId < MAX_AUDIO_OVERRIDES; seqId++) {
>     if (!sAudioOverrides[seqId].enabled) {
>         return seqId;
>     }
> }
> ```

Updated to be similar to this, but made it do SEQ_COUNT + 1 to avoid errors, the test script has also changed:
`-- name: Allocate Sequence Mod Test
-- descripton: For Testing


local MY_SEQ = smlua_audio_utils_allocate_sequence()
log_to_console(tostring(MY_SEQ), CONSOLE_MESSAGE_INFO) --Prints 36
smlua_audio_utils_replace_sequence(MY_SEQ, 0x25, 74, "BooWaltz") --Set this to a M64  In Sounds
local MY_SEQ2 = smlua_audio_utils_allocate_sequence()
log_to_console(tostring(MY_SEQ2), CONSOLE_MESSAGE_INFO) --Prints 37
smlua_audio_utils_replace_sequence(MY_SEQ2, 0x25, 74, "CybersWorld") --Set this to a M64  In Sounds


    if m.action  ==  ACT_GROUND_POUND then
        if (m.controller.buttonPressed & B_BUTTON) ~= 0 then 
            play_cap_music(MY_SEQ2)
        end 
        if (m.controller.buttonPressed & A_BUTTON) ~= 0 then 
            play_cap_music(MY_SEQ)
        end 
    end 
end

hook_event(HOOK_MARIO_UPDATE, mario_update)`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -154,6 +154,16 @@ void smlua_audio_utils_replace_sequence(u8 sequenceId, u8 bankId, u8 defaultVolu
...
 }
 
+u8 smlua_audio_utils_allocate_sequence(void) {
+    for (u8 seqId = SEQ_COUNT + 1; seqId < MAX_AUDIO_OVERRIDE; seqId++) {
+        if (!sAudioOverrides[seqId].enabled) {
+            return seqId;
+        }
+    }
+    LOG_ERROR("Cannot allocate more custom sequences.");
+    return 0;
```
djoslin0: Should probably not return `0` since that corresponds to `SEQ_SOUND_PLAYER`
1. If a mod calls `allocate_sequence()` when all slots are full, it gets `0`
2. The mod then calls `smlua_audio_utils_replace_sequence(0, ...)`
3. This would override `SEQ_SOUND_PLAYER`

It should return `MAX_AUDIO_OVERRIDE` instead of `0` so that `smlua_audio_utils_replace_sequence()` will error properly in this scenario

djoslin0 (changes requested):

src/pc/lua/utils/smlua_audio_utils.c
Isaac0-dev: dj has a good point here
```suggestion
    return MAX_AUDIO_OVERRIDE;
```

PR #892, opened 2025-07-13
Fix colon mapping in sSm64CharMap

Beckowl:
The ellipsis doesn’t map to `0xE6`, it’s actually just three `0x3F` (period) characters.
`0xE6` is used for the colon, so this just fixes the mapping

### DIFF
--- a/src/game/level_info.c
+++ b/src/game/level_info.c
@@ -40,7 +40,7 @@ const struct { const char *str; u8 c; u8 menu; } sSm64CharMap[] = {
     { "z", 0x3D, 0 },
 
     // Punctuation
-    { "...", 0xE6, 0 }, // ellipsis
+    { ":",   0xE6, 0 }, // colon
     { ")(",  0xE2, 0 }, // close-open parentheses
     { "<<",  0xF5, 0 }, // double quote open
     { ">>",  0xF6, 0 }, // double quote close


Beckowl:
Might as well just make `smlua_text_utils_dialog_replace` use `ascii_to_sm64_char` (from level_info.h) instead of `str_ascii_to_dialog` (from ingame_menu.c)??? they do the same but some mappings are completely different. Is this a leftover from before dialogs were loaded at runtime??

This makes dialogEntries' text field inconsistent

Beckowl:
Peachy pointed out the other redundant mappings so i got rid of them. I set the menu params to true in the calls from ingame_menu because i don't see a reason to not allow some characters in menus it's a bit dumb

PeachyPeachSM64 (approved):

src/game/level_info.c
```diff
@@ -67,16 +67,17 @@ const struct { const char *str; u8 c; u8 menu; } sSm64CharMap[] = {
...
-    { "<->", 0xE4, 0 }, // left-right arrow
+    { "+",   0xE4, 0 }, // left-right arrow
     { "^",   0x50, 0 }, // up arrow
     { "|",   0x51, 0 }, // down arrow
     { "<",   0x52, 0 }, // left arrow
     { ">",   0x53, 0 }, // right arrow
-    { "+",   0xF9, 1 }, // coin
-    { "@",   0xFA, 1 }, // star filled
+    { "$",   0xF9, 1 }, // coin
+    { "★",   0xFA, 1 }, // star filled
```
Isaac0-dev, line 76: can you fix this indentation here

Isaac0-dev (changes requested):
sorry to be so picky but

src/game/level_info.c
Beckowl: it is the exact same as all the other characters, the star is just wider for some reason

Isaac0-dev: oh you're right, and you can't fix it in any way either. sorry about that.

PR #894, opened 2025-07-16
Add default pvp type (Revamped update)

Sunketchupm:
Added a new pvp type called Default, which aims to be the simplest and default version of pvp.

Default:
* Ground pounds deal 3 damage
* Kicks, breakdances, slidekicks, twirls, and punches deal 2 damage
* Everything else deals 1 damage

Revamped also got an overhaul:
* Removed all knockback distance changes, as they made little sense to begin with
* Punches will now deal 3 damage instead of 2 if moving fast enough, since at such a speed a punch combo is impossible anyways
* Breakdances now always do hard knockback
* Cannon shots deal 4 damage (suggestion by a few people)
* Ground pounds deal scaling damage, where the more you fall, the more damage it deals. Deals a base of 3 damage.
* Slidekicks deal scaling damage, where the faster you move, the more damage it deals. Deals a base of 1 damage.

Overall:
* Improved the rollout hitbox so that it doesn't last nearly as long
* Moves that are attacks are now prevented from bouncing


Blockyyy:
EXCOOP PVP, YOU'RE BACK

EmilyEmmi:
Should Classic be given a different name, since it's no longer the same as the previous version's Classic? Perhaps "Old" or "Excoop"? People might enable it assuming its the same as the current "Classic".

AgentXLP:
We don't need three PVP modes, one was already enough let alone two and this is just overkill now.

Sunketchupm:
I've said a few times to some people that revamped (and pvp modes in general) is just a mod that's built into coop as I couldn't figure out a good way to make them work as a mod. If I could figure how to make it work well, then I do plan to just remove the pvp modes.
So yeah I agree 3 is a lot.

AgentXLP:
Alright, that can work

PR #895, opened 2025-07-19
Add functions to enable/disable mods and a function to commit changes by restarting the server

Sploder-Saptarshi:
This will allow for extremely dynamic servers for hosting gamemodes and stuff. You can host multiple gamemodes and switch between them with a vote by players!

Demo of what's possible if this is merged:

https://github.com/user-attachments/assets/963e94c6-6a90-47dd-af9e-666c77905a06

### DIFF
--- a/src/pc/lua/smlua.h
+++ b/src/pc/lua/smlua.h
@@ -15,6 +15,7 @@
 #include "smlua_functions_autogen.h"
 #include "smlua_hooks.h"
 #include "smlua_sync_table.h"
+#include "smlua_mod_funcs.h"
 
 #include "pc/debuglog.h"
 #include "pc/djui/djui_console.h"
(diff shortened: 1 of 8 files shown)


Isaac0-dev:
this is a level of power i don't think mods should have personally

Sploder-Saptarshi:
> this is a level of power i don't think mods should have personally

Fair point. Will it be better if we have an explicit command-line argument to make these functions available so they can't be used by default?

Isaac0-dev:
This just makes me concerned that the feature would become useless to most folks if we added the cli option.

Sploder-Saptarshi:
> This just makes me concerned that the feature would become useless to most folks if we added the cli option.

I mean, it's really just needed on the client hosting the lobby, aka the "server" and hosters can manually enable it

PR #896, opened 2025-07-21
Add  ``smlua_text_utils_allocate_dialog``

Beckowl:
Mods can now create fully custom dialog boxes, without replacing any vanilla dialogs, as follows:

```lua
local dialogID = smlua_text_utils_allocate_dialog()

smlua_text_utils_dialog_replace(dialogID, 1, 6, 30, 200, "Hello, World!\n\nHello from dialog ID " .. dialogID)

-- show the dialog box
hook_event(HOOK_ON_LEVEL_INIT, function()
    create_dialog_box(dialogID)
end)
```

image

---

## Important changes:

- The old ``seg2_dialog_table`` and related seg2 dialog entry tables have been removed, all dialog data is now in ``gDialogTable``
- Some function signatures were changed to use ``s32`` (instead of ``u8`` or ``s16``) for dialog ids


Beckowl:
Just keep in mind that I did very little testing, it seems to work fine but i can't test it more throroughly right now

src/game/ingame_menu.c
```diff
@@ -128,7 +129,7 @@ f32 gDialogBoxOpenTimer = DEFAULT_DIALOG_BOX_ANGLE;
 f32 gDialogBoxScale = DEFAULT_DIALOG_BOX_SCALE;
 s16 gDialogScrollOffsetY = 0;
 s8 gDialogBoxType = DIALOG_TYPE_ROTATE;
-s16 gDialogID = -1;
+s32 gDialogID = -1;
```
PeachyPeachSM64: DIALOG_NONE

src/game/ingame_menu.c
```diff
@@ -1172,29 +1173,29 @@ bool handle_dialog_hook(s16 dialogId) {
     return true;
 }
 
-void create_dialog_box(s16 dialog) {
+void create_dialog_box(s32 dialog) {
     if (handle_dialog_hook(dialog) && gDialogID == -1) {
```
PeachyPeachSM64: DIALOG_NONE

src/game/ingame_menu.c
```diff
@@ -1172,29 +1173,29 @@ bool handle_dialog_hook(s16 dialogId) {
...
         gDialogID = dialog;
         gDialogBoxType = DIALOG_TYPE_ROTATE;
     }
 }
 
-void create_dialog_box_with_var(s16 dialog, s32 dialogVar) {
+void create_dialog_box_with_var(s32 dialog, s32 dialogVar) {
     if (handle_dialog_hook(dialog) && gDialogID == -1) {
```
PeachyPeachSM64: DIALOG_NONE

src/game/ingame_menu.c
```diff
@@ -1172,29 +1173,29 @@ bool handle_dialog_hook(s16 dialogId) {
...
         gDialogID = dialog;
         gDialogVariable = dialogVar;
         gDialogBoxType = DIALOG_TYPE_ROTATE;
     }
 }
 
-void create_dialog_inverted_box(s16 dialog) {
+void create_dialog_inverted_box(s32 dialog) {
     if (handle_dialog_hook(dialog) && gDialogID == -1) {
```
PeachyPeachSM64: DIALOG_NONE

src/game/ingame_menu.c
```diff
@@ -1172,29 +1173,29 @@ bool handle_dialog_hook(s16 dialogId) {
...
         gDialogID = dialog;
         gDialogBoxType = DIALOG_TYPE_ZOOM;
     }
 }
 
-void create_dialog_box_with_response(s16 dialog) {
+void create_dialog_box_with_response(s32 dialog) {
     if (handle_dialog_hook(dialog) && gDialogID == -1) {
```
PeachyPeachSM64: DIALOG_NONE

src/game/ingame_menu.c
```diff
@@ -1957,35 +1958,13 @@ void render_dialog_entries(void) {
...
-    }
-    dialogTable = segmented_to_virtual(seg2_dialog_table);
-#endif
-    dialog = segmented_to_virtual(dialogTable[gDialogID]);
+    struct DialogEntry *dialog = dialog_table_get(gDialogID);
 
-    // if the dialog entry is invalid, set the ID to -1.
-    if (segmented_to_virtual(NULL) == dialog) {
+    if (dialog == NULL) {
         gDialogID = -1;
```
PeachyPeachSM64: DIALOG_NONE

src/game/ingame_menu.c
```diff
@@ -2278,28 +2257,11 @@ void do_cutscene_handler(void) {
...
-    }
-#else
-    dialogTable = segmented_to_virtual(seg2_dialog_table);
-#endif
-    dialog = segmented_to_virtual(dialogTable[gDialogID]);
 
-    str = sOverrideDialogString ? sHookString : segmented_to_virtual(dialog->str);
+    struct DialogEntry *dialog = dialog_table_get(gDialogID);
+
+    str = sOverrideDialogString ? sHookString : (u8*)dialog->str;
```
PeachyPeachSM64: use const u8*, then remove cast

src/game/level_update.c
```diff
@@ -299,7 +299,7 @@ void stub_level_update_1(void) {
 
 void load_level_init_text(u32 arg) {
     s32 gotAchievement;
-    u32 dialogID = gCurrentArea->dialog[arg];
+    s32 dialogID = gCurrentArea->dialog[arg];
```
PeachyPeachSM64, line 302: area->dialog is still u8
I don't know if it's currently possible, but what if a custom level wants to use a custom dialog id > 255?

src/pc/dialog_table.h
```diff
@@ -0,0 +1,27 @@
...
+#include "PR/ultratypes.h"
+#include "game/ingame_menu.h"
+
+#define MAX_ALLOCATED_DIALOGS 2147483647 // max s32
+
+typedef struct {
+    struct DialogEntry **data;
+    s32 size;
+    s32 capacity;
+} DialogTable;
```
PeachyPeachSM64: why not using GrowingArray?

src/pc/dialog_table.h
```diff
@@ -0,0 +1,27 @@
+#ifndef DIALOG_TABLE_H
+#define DIALOG_TABLE_H
+
+#include "PR/ultratypes.h"
+#include "game/ingame_menu.h"
+
+#define MAX_ALLOCATED_DIALOGS 2147483647 // max s32
```
PeachyPeachSM64: a bit too much
also pointless, capacity will explode at count around 1.07M due to overflow (if ones computer has enough RAM to allocate this)

src/pc/dialog_table.c
```diff
@@ -0,0 +1,134 @@
...
+    DialogTable *table = &gDialogTable;
+
+    table->capacity = 256;
+    table->size = 0;
+    table->data = malloc(sizeof(struct DialogEntry*) * table->capacity);
+
+    if (!table->data) {
+        LOG_ERROR("Failed to allocate dialog table");
+        exit(EXIT_FAILURE);
+    }
```
PeachyPeachSM64: this is growing_array_init

src/pc/dialog_table.c
```diff
@@ -0,0 +1,134 @@
...
+            LOG_LUA_LINE_WARNING("Failed to resize dialog table");
+            return -1;
+        }
+
+        table->data = newData;
+        table->capacity = newCapacity;
+    }
+
+    s32 index = table->size;
+    table->data[table->size++] = dialog;
```
PeachyPeachSM64: this is growing_array_alloc
don't reinvent the wheel
also as stated before, since your capacity is signed, it will overflow at size around INT_MAX/2

src/pc/dialog_table.c
```diff
@@ -0,0 +1,134 @@
...
+        free((u8*)dialog->str);
+        free(dialog->text);
+
+        if (!IS_CUSTOM_DIALOG(i)) {
+            const struct DialogEntry *dialogOrig = smlua_text_utils_dialog_get_unmodified(i);
+
+            memcpy(dialog, dialogOrig, sizeof(struct DialogEntry));
+            dialog->text = get_dialog_text_ascii(dialog); 
+        }
+        else {
```
PeachyPeachSM64: don't do this

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -281,16 +244,22 @@ void smlua_text_utils_dialog_replace(enum DialogId dialogId, UNUSED u32 unused,
...
+}
+
+s32 smlua_text_utils_allocate_dialog(void) {
+    struct DialogEntry* dialog = calloc(1, sizeof(struct DialogEntry));
+
+    dialog->replaced = true;
 
-    return sReplacedDialog[dialogId];
+    return dialog_table_add(dialog);
 }
```
PeachyPeachSM64: this is cool stuff, but what if I want to have a fixed dialogId?
for NPCs for example, dialogId is fixed to the object behavior params when the level is generated
(but the dialog itself doesn't need to exist when the level is loaded)
indeed a custom NPC behavior would solve the issue, but that sounds a bit inconvenient

PeachyPeachSM64 (changes requested):

src/pc/lua/utils/smlua_text_utils.c
Beckowl: then mod loading order would mess it up?? you can't have fixed dialog ids

Beckowl: mods could also conflict with each other i don't think that would work

src/pc/dialog_table.h
Beckowl: I didnt know about it :p

src/game/level_update.c
Beckowl: mod loading order would mess up constant ids

src/pc/lua/utils/smlua_text_utils.c
PeachyPeachSM64: with either ways nothing would prevent a mod from using an already allocated dialog id though
and we can't do nothing about this
but being able to choose which id the user wants could be a plus
but that's just my opinion

Beckowl: a way this could work is:
- each mod has its own growing dialog table
- objects/levels have a mod index field
- get dialog would return ``DIALOG_COUNT - index`` from the mod's dialog table if index >= DIALOG_COUNT??

PeachyPeachSM64: mmh that seems tricky
what if for example a mod references an object (let's say a bobomb buddy) spawned by another mod, retrieve its dialog id and then call create_dialog_box or whatever with that id?
how will the game know it's from that specific object from the other mod?

PeachyPeachSM64: maybe the way it could be solved is by allowing lua variables in beh params in objects, like how it is done for the model and behavior
then we could have something like this:
```c
...
OBJECT(MODEL_BOBOMB_BUDDY, ..., myBobombBuddyBehParams, bhvBobombBuddy),
...
```
and in a lua file:
```lua
local myBobombBuddyDialogId = smlua_text_utils_allocate_dialog()
myBobombBuddyBehParams = myBobombBuddyDialogId << 16
```

Beckowl: that's good yeah
i don't know how the level system works though, so i'll have to look into that

Beckowl:
Other than the fact that ``HOOK_ON_DIALOG`` half works, it should be fine now

I don't know where/what to modify to make the level object macro work with lua variables though, i need some help with that

src/pc/dialog_table.c
```diff
@@ -0,0 +1,100 @@
...
+void dialog_table_init(void) {
+    gDialogTable = growing_array_init(gDialogTable, 256);
+
+    for (u32 i = 0; i < DIALOG_COUNT; i++) {
+        const struct DialogEntry* dialogOrig = smlua_text_utils_dialog_get_unmodified(i);
+        struct DialogEntry* dialog = dialog_table_alloc(NULL);
+
+        if (!dialog) {
+            LOG_ERROR("Failed to allocate DialogEntry %d", i);
+            exit(EXIT_FAILURE);
```
PeachyPeachSM64: use `sys_fatal` with the error message instead of just closing itself

src/pc/dialog_table.c
```diff
@@ -0,0 +1,100 @@
...
+struct DialogEntry* dialog_table_alloc(s32 *dialogId) {
+    DialogTable *table = gDialogTable;
+
+    if (table->count >= MAX_ALLOCATED_DIALOGS) {
+        LOG_LUA_LINE_WARNING("Dialog limit reached! (%d max)", MAX_ALLOCATED_DIALOGS);
+        if (dialogId) *dialogId = DIALOG_NONE;
+        return NULL;
+    }
+
+    struct DialogEntry* alloc = growing_array_alloc(table, sizeof(struct DialogEntry));
```
PeachyPeachSM64: not really a good name, `dialog` would be better

PeachyPeachSM64 (changes requested):

Beckowl:
I tried to add lua behavior parameters yesterday and made these changes:

```diff
+++ b/data/dynos_bin_lvl.cpp
@@ -733,6 +733,7 @@ static void ParseLevelScriptSymbol(GfxData* aGfxData, DataNode<LevelScript>* aNo
 
         bool foundModel = true;
         bool foundBeh = true;
+        bool foundBehParam = true;
         LevelScript model    = ParseLevelScriptSymbolArgInternal(aGfxData, aNode, aTokenIndex, &foundModel);
         LevelScript posX     = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
         LevelScript posY     = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
@@ -740,9 +741,13 @@ static void ParseLevelScriptSymbol(GfxData* aGfxData, DataNode<LevelScript>* aNo
         LevelScript angleX   = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
         LevelScript angleY   = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
         LevelScript angleZ   = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
-        LevelScript behParam = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
+        LevelScript behParam = ParseLevelScriptSymbolArgInternal(aGfxData, aNode, aTokenIndex, &foundBehParam);
         LevelScript beh      = ParseLevelScriptSymbolArgInternal(aGfxData, aNode, aTokenIndex, &foundBeh);
 
+        if (!foundBehParam) {
+            behParam = DynOS_Lua_RememberVariable(aGfxData, aHead + 4, aNode->mTokens[topTokenIndex + 7]);
+        }
+
         if (foundModel && foundBeh) {
             aGfxData->mPointerList.Add(aHead + 5);
             LevelScript _Ls[] = { OBJECT(model, posX, posY, posZ, angleX, angleY, angleZ, behParam, beh) };
@@ -769,6 +774,7 @@ static void ParseLevelScriptSymbol(GfxData* aGfxData, DataNode<LevelScript>* aNo

         bool foundModel = true;
         bool foundBeh = true;
+        bool foundBehParam = true;
         LevelScript model    = ParseLevelScriptSymbolArgInternal(aGfxData, aNode, aTokenIndex, &foundModel);
         LevelScript posX     = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
         LevelScript posY     = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);
@@ -780,6 +786,11 @@ static void ParseLevelScriptSymbol(GfxData* aGfxData, DataNode<LevelScript>* aNo
         LevelScript beh      = ParseLevelScriptSymbolArgInternal(aGfxData, aNode, aTokenIndex, &foundBeh);
         LevelScript acts     = ParseLevelScriptSymbolArg(aGfxData, aNode, aTokenIndex);

+        if (!foundBehParam) {
+            behParam = DynOS_Lua_RememberVariable(aGfxData, aHead + 4, aNode->mTokens[topTokenIndex + 7]);
+        }
+
+
         if (foundModel && foundBeh) {
             aGfxData->mPointerList.Add(aHead + 5);
             LevelScript _Ls[] = { OBJECT_WITH_ACTS(model, posX, posY, posZ, angleX, angleY, angleZ, behParam, beh, acts) };
```

the levels compile fine, and levels that use constant behavior params still work, but when i try loading a level that uses a Lua behavior param, i get this in the console:
```
Didn't expect a pointer while reading level script: level_bob_entry, 1447122252
```
I don't really know how dynos works so this is as far as i could get for now

Beckowl:
Requesting review so i can get some guidance with the lua variables part

Beckowl:
last dialog addition

Beckowl:
got something to work

https://github.com/user-attachments/assets/91ce9fc8-fcd0-4066-b944-2d0007ca0337

PeachyPeachSM64:
sorry for the late review
I like the `flags` parameter in the level command, so I extended it to all OBJECT fields
removed `level_cmd_place_object_ext`, `level_cmd_place_object_ext2` and made `level_cmd_place_object_ext_lua_params` more generic, because there was so much duplicated code
same for `smlua_func_level_script_parse_callback`
I wish I could delete OBJECT_EXT and OBJECT_EXT2, but retro-compatibility must be ensured

PeachyPeachSM64 (approved):
this is starting to get out of this PR initial scope, so we should stop here (partially my fault I admit)

Isaac0-dev (changes requested):
I'm sorry but this PR contains far more than just "Add smlua_text_utils_allocate_dialog".
It's more of a misc changes PR at this point, which is really hard to effectively review and test. Are you okay with splitting this PR into parts? I know you've already got approvals from Peachy, but even so I think it should be split up. I remember times when I used to do this and it always ended in a mess and reviewers often ended up merging with a more hopeful attitude than real confidence.

Beckowl:
I’ve reverted the level macro changes. Aside from ``smlua_text_utils_dialog_restore``, the changes should now all be related to allocate dialog. This still touches a lot of files, but most of those edits are just for normalizing dialogID types, so i believe it should be ok now

src/pc/pc_main.c
```diff
@@ -446,6 +448,7 @@ void* main_game_init(UNUSED void* dummy) {
 
     LOADING_SCREEN_MUTEX(loading_screen_set_segment_text("Loading ROM Assets"));
     rom_assets_load();
+    dialog_table_init();
```
Isaac0-dev: can't you put `dialog_table_init` and `dialog_table_shutdown` inside the `smlua_text_utils_` functions? these functions exist because of the smlua text utils

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -281,16 +244,41 @@ void smlua_text_utils_dialog_replace(enum DialogId dialogId, UNUSED u32 unused,
...
 
-    sReplacedDialog[dialogId] = true;
+void smlua_text_utils_dialog_restore(enum DialogId dialogId) {
+    if (!IS_VALID_VANILLA_DIALOG(dialogId)) return;
+
+    struct DialogEntry *dialog = smlua_text_utils_dialog_get(dialogId);
+
+    if (!dialog->replaced) return;
+
+    const struct DialogEntry *dialogOrig = smlua_text_utils_dialog_get_unmodified(dialogId);
```
Isaac0-dev, line 260: seems to me at this point we need 2 DialogEntry structs. one that's for the original vanilla copies of a dialog, and the other for modified dialogs.
because currently, we're taking up 2 more fields in the `DialogEntry` structure which is just wasting memory for the original copies, they only exist as a copy for restoring later anyway. 
it's not actually going to hurt anyone if this isn't done, it's just bad practice imo. It's fine if you don't agree with me though.

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -281,16 +244,41 @@ void smlua_text_utils_dialog_replace(enum DialogId dialogId, UNUSED u32 unused,
...
+    free(dialog->text);
+
+    memcpy(dialog, dialogOrig, sizeof(struct DialogEntry));
+    dialog->text = get_dialog_text_ascii(dialog);
 }
 
 bool smlua_text_utils_dialog_is_replaced(enum DialogId dialogId) {
-    if (dialogId >= DIALOG_COUNT) {
-        return false;
+    if (!IS_VALID_DIALOG(dialogId)) return false;
```
Isaac0-dev: at least throughout this file, please keep the code style consistent. put braces around these 1 liner conditions

src/pc/dialog_table.h
```diff
@@ -0,0 +1,23 @@
+#ifndef DIALOG_TABLE_H
+#define DIALOG_TABLE_H
+
+#include "PR/ultratypes.h"
+#include "game/ingame_menu.h"
+
+#define MAX_ALLOCATED_DIALOGS 32767
```
Isaac0-dev: just a nitpick, i think `0x8000` is easier to identify as half the max of a u16

src/audio/external.c
```diff
@@ -2455,10 +2457,10 @@ void set_sound_moving_speed(u8 bank, u8 speed) {
 /**
  * Called from threads: thread5_game_loop
  */
-void play_dialog_sound(u8 dialogID) {
+void play_dialog_sound(s32 dialogID) {
     s32 speaker;
 
-    if (dialogID >= DIALOG_COUNT) {
+    if (!IS_VALID_DIALOG(dialogID) || IS_CUSTOM_DIALOG(dialogID)) {
```
Isaac0-dev: `!IS_VALID_VANILLA_DIALOG(dialogID)` should be fine here

include/object_fields.h
```diff
@@ -970,7 +970,7 @@
 #define /*0x10C*/ oTiltingPyramidMarioOnPlatform OBJECT_FIELD_S32(0x21)
 
 /* Toad Message */
-#define /*0x108*/ oToadMessageDialogId       OBJECT_FIELD_U32(0x20)
+#define /*0x108*/ oToadMessageDialogId       OBJECT_FIELD_S32(0x20)
```
Isaac0-dev, line 973: just acknowledging that this may break mods that use this with high values

Isaac0-dev (changes requested):
looks mostly good, these are just my nitpicks

PR #897, opened 2025-07-24
splitscreen local multiplayer

Isaac0-dev:
up to 4 players can play together on the same coop client. network functionality is currently broken.
there are plenty of issues with this, but as of now it is playable.
mod support is very unreliable, and will likely never reach a point of perfection simply because mods don't expect multiplayer to be handled in this way.
this was implemented in a very hacky way, by swapping mario states and player indices around. this method is used to avoid breaking some mods and needing to swap out much of coop's mario related code.

There are some working builds of this project available [here](https://github.com/Isaac0-dev/sm64coopdx/actions/workflows/build-coop.yaml?query=branch%3Asplitscreen), just choose the most recent successful build and download the artifact for your machine.

long term goals for this project:
- fully fleshed out 4 player couch vanilla sm64 coop experience
- full network compatibility, syncing all 4 players with the network, pretending each one is a different client.
- very limited mod support, some mods will need dedicated support for it. most mods won't work at all though

being my favourite coop mod, i've added some manual support for omm rebirth using some hacks, although it still is not perfect.
I'm not sure if i'll personally continue this project which is why i'm opening a pull request here so anyone can keep it going if they wish (although you'd have to deal with my terrible code). 
If you decide to continue on it, please let me know about your progress! I'd still love to see this finished one day.

Here is a short term todo list that I had written some time ago:
```
- fix player 1 broken torso tilt
- add a real controller system and menu
- fix hud builtin, star display
- fix holp pushing only player 1
- deprecate matrixPtr and matrixPtrPrev
- fix transitions being so fast
- add multiple rooms rendering at once
- exit painting position
- fix infinite cappy jumps on local players
- fix omm hud
- omm cappy homing
- interpolation: camera movement, skyboxes
- play sound effects are buggy, sometimes it plays the wrong one
- there are some spots with oHiddenStarLastInteractedObject == &gMarioStates[0], predicted desync
```


Isaac0-dev:
if anyone decides to try this, you may want to limit your fps to 30 to disable interpolation. i forgot that i left interpolation active. At the moment, it interpolates the camera positions between each player. The experience is very buggy as i mentioned.

djoslin0:
WIP & Stale. And should probably not be merged into the main repo. "Proper" splitscreen support would require existing mods to work. There are ways to do this, but it isn't easy.

AgentXLP:
> WIP & Stale. And should probably not be merged into the main repo. "Proper" splitscreen support would require existing mods to work. There are ways to do this, but it isn't easy.

Yeahhh sorry but I don't see this being merged into the main repo and it hasn't been worked on in a long time either

PR #899, opened 2025-07-29
Add Option to Save Active Palette as Preset

Baconator2558:
Intended for use with mods like Character Select that override the player's palette. This menu option lets players easily save those palettes as presets.


src/pc/djui/djui_panel_player.c
```diff
@@ -326,6 +334,58 @@ static void djui_panel_player_edit_palette_create(struct DjuiBase* caller) {
...
 
+static void djui_panel_player_name_active_palette(struct DjuiBase* caller) {
+    gDjuiInPlayerMenu = true;
+
+    struct DjuiThreePanel* panel = djui_panel_menu_create(DLANG(PLAYER, PALETTE), true);
+
+    // A bit of a gross hack to send out palette changes and update the palette preset selection box on unpause AND
+    // pressing the Back button
+    sSavedDestroy = panel->base.destroy;
+    panel->base.destroy = djui_panel_player_edit_palette_destroy;
```
PeachyPeachSM64, line 345: I was about to comment on this, but it seems to be a copy/paste of `djui_panel_player_edit_palette_create`
As long as it works, it's fine, I guess

Baconator2558: If you think it'd be better, I could change the comment to something like "Same hacky fix used in djui_panel_player_edit_palette_create"

PR #901, opened 2025-07-30
revise better frame pacing

Isaac0-dev:
improved frame pacing, where frame delaying is done just before displaying the frame.
so, the frame is fully rendered, and then the delay is done to wait the extra time before displaying the frame.
This is a far more reliable and consistent way to manage frame pacing, and it will help to keep the framerate from choking too badly when the game lags a little. 

The changes to `gfx_end_frame` is to minimize the steps necessary between delaying and actually displaying the frame. This is an optimization just to increase the accuracy of frame pacing even further.

Also, delta calculation is improved when the number of frames to draw is known.

### DIFF
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -223,7 +223,6 @@ static u32 get_refresh_rate() {
 
 void produce_interpolation_frames_and_delay(void) {
     u32 refreshRate = get_refresh_rate();
-    bool is30Fps = (refreshRate == FRAMERATE);
 
     gRenderingInterpolated = true;
 
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -266,7 +266,6 @@ void produce_interpolation_frames_and_delay(void) {
 
         // send the frame to the screen (should be directly after the delay for good frame pacing)
         gfx_display_frame();
-        gfx_display_frame(); // ! for debugging purposes
         sDrawnFrames++;
         if (configFramerateMode != RRM_UNLIMITED) { numFramesToDraw--; }
     } while ((curTime = clock_elapsed_f64()) < targetTime && numFramesToDraw > 0);
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -266,6 +266,7 @@ void produce_interpolation_frames_and_delay(void) {
 
         // send the frame to the screen (should be directly after the delay for good frame pacing)
         gfx_display_frame();
+        gfx_display_frame(); // ! for debugging purposes
         sDrawnFrames++;
         if (configFramerateMode != RRM_UNLIMITED) { numFramesToDraw--; }
     } while ((curTime = clock_elapsed_f64()) < targetTime && numFramesToDraw > 0);
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -205,8 +205,8 @@ static s32 get_num_frames_to_draw(f64 t, u32 frameLimit) {
 static u32 get_refresh_rate() {
     if (configFramerateMode == RRM_MANUAL) { return configFrameLimit; }
     if (configFramerateMode == RRM_UNLIMITED) { return 3000; } // Has no effect
-    static u32 refreshRate = 0;
 #ifdef HAVE_SDL2
+    static u32 refreshRate = 0;
     if (!refreshRate) {
         SDL_DisplayMode mode;
         if (SDL_GetCurrentDisplayMode(0, &mode) == 0) {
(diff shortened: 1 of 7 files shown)


Isaac0-dev:
thanks for reminding me that directx is actually an option we support.
From just looking at the code, my changes shouldn't break anything.
but I can't test it reliably myself right now.

Beckowl:
VSync now caps to 30 FPS instead of 60 FPS (single 60 Hz monitor):
image

---

60 FPS appears less smooth compared to 1.3.2, at least for me, you might not notice it depending on your setup

https://github.com/user-attachments/assets/52139c48-9ef5-4e09-b348-b916a1612d3b

Isaac0-dev:
djui should look smooth again (theoretically it should look smoother than before for screens with higher than 60hz)

i also tried to do something about this:
> VSync now caps to 30 FPS instead of 60 FPS (single 60 Hz monitor)

I guessed that this is more likely to do with Auto mode choosing the wrong refresh rate so I made some changes to it in case. But I do need more information though. Does Vsync "cap to 30" if you have manual mode on and the limiter set to 60?

Beckowl:
> djui should look smooth again (theoretically it should look smoother than before for screens with higher than 60hz)
> 
> i also tried to do something about this:
> 
> > VSync now caps to 30 FPS instead of 60 FPS (single 60 Hz monitor)
> 
> I guessed that this is more likely to do with Auto mode choosing the wrong refresh rate so I made some changes to it in case. But I do need more information though. Does Vsync "cap to 30" if you have manual mode on and the limiter set to 60?

Sorry for the late reply, I didn't have time to compile and test your changes earlier.

---

VSync caps to 30 if i have manual + limit set to 60:
image

If i set the mode to uncapped, it jumps to 50 fps but won't go higher:
image

Djui is so much smoother now. I actually feel i'm in control of the cursor now, lol

Beckowl:
Holding the disconnect button when you host a server will crash (I forgot to mention earlier, but this is opengl)
```
#0  gfx_sp_matrix (parameters=1 '\001', addr=<optimized out>) at src/pc/gfx/gfx_pc.c:670
        matrix = {{0, 0, 1, 0}, {-0.03125, 0.03125, 0, 1}, {0, -1, 0, 2.97400071e-36}, {
            -nan(0x7f02f9), 9.18340949e-41, 0.5, -1}}
#1  gfx_run_dl (cmd=0x14388bde0 <gGfxPools+70656>) at src/pc/gfx/gfx_pc.c:1679
        opcode = 218
#2  0x0000000140118b84 in gfx_run (commands=<optimized out>) at src/pc/gfx/gfx_pc.c:1972
No locals.
#3  0x00000001400fb2b6 in send_display_list (spTask=<optimized out>) at src/pc/pc_main.c:127
No locals.
#4  send_display_list (spTask=<optimized out>) at src/pc/pc_main.c:125
No locals.
#5  produce_interpolation_frames_and_delay () at src/pc/pc_main.c:253
        idealTime = <optimized out>
        delta = <optimized out>
        refreshRate = <optimized out>
        targetTime = 11.680511933333333
        numFramesToDraw = 100
        curTime = 11.647757
        loopStartTime = 11.647757
        expectedTime = 0
        framesDrawn = 1
        interpFrameTime = 0.00033333333333333332
#6  0x00000001400fb697 in produce_one_frame () at src/pc/pc_main.c:355
No locals.
#7  0x00000001400fbd89 in SDL_main (argc=<optimized out>, argv=<optimized out>)
    at src/pc/pc_main.c:599
        threadSuccess = <optimized out>
#8  0x00000001403a74dc in main_getcmdline ()
No symbol table info available.
#9  0x00000001400010c9 in __tmainCRTStartup ()
    at D:/W/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:236
        lock_free = 0x0
        fiberid = <optimized out>
        nested = 0
        startinfo = {newmode = 0}
        ret = 0
#10 0x0000000140001416 in mainCRTStartup () at D:/W/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:122
        ret = 255
```


https://github.com/user-attachments/assets/28440d61-6436-419d-ae9d-12a5c7f2b277

src/pc/pc_main.c
```diff
@@ -255,19 +255,19 @@ void produce_interpolation_frames_and_delay(void) {
...
                 precise_delay_f64(delay);
             }
-            numFramesToDraw--;
         }
 
         // send the frame to the screen (should be directly after the delay for good frame pacing)
         gfx_display_frame();
         sDrawnFrames++;
+        if (configFramerateMode != RRM_UNLIMITED) { numFramesToDraw--; }
     } while ((curTime = clock_elapsed_f64()) < targetTime && numFramesToDraw > 0);
```
PeachyPeachSM64: I don't get these changes.
While moving `expectedTime` calculation could result in micro(nano?)-seconds of difference in `now` calculation (and actually it's even up to the compiler to decide what to do first), moving `numFramesToDraw--` changes nothing.

PeachyPeachSM64 (changes requested):

src/pc/pc_main.c
```diff
@@ -205,12 +206,12 @@ static s32 get_num_frames_to_draw(f64 t, u32 frameLimit) {
...
     if (configFramerateMode == RRM_MANUAL) { return configFrameLimit; }
     if (configFramerateMode == RRM_UNLIMITED) { return 3000; } // Has no effect
-    static u32 refreshRate = 0;
 #ifdef HAVE_SDL2
+    static u32 refreshRate = 0;
     if (!refreshRate) {
         SDL_DisplayMode mode;
-        if (SDL_GetCurrentDisplayMode(0, &mode) == 0) {
-            refreshRate = (u32) mode.refresh_rate;
+        if (SDL_GetWindowDisplayMode(0, &mode) == 0) {
```
PeachyPeachSM64: are you sure about that change?
From the SDL doc, it is stated that `SDL_GetWindowDisplayMode: Query the display mode to use when a window is visible at fullscreen.`

PeachyPeachSM64: also it takes a `SDL_Window * window` as first parameter, not int

src/pc/pc_main.c
Isaac0-dev: Yeah, you're right and I made those changes fully aware of that. As I said in my commit name, its "paranoid". But I'm hoping to emphasise what parts of the code actually matter in terms of frame pacing.

src/pc/pc_main.c
Isaac0-dev: i now know that i didn't need to make that change anyway thanks to beckowl's tests. I'll just revert that change.

Isaac0-dev:
fixed the crash beckowl mentioned.

Beckowl:
> @Beckowl this is direct X you're talking about here right? Just wanting to confirm. If it is, my speculation is that direct X is doing frame delaying without detecting that coop is also delaying. So 2 systems delaying at the same time? Idk. I think I need to try to make my test environment as similar to yours as possible. I think I'll need to test on a real windows machine. But being certain that it's direct X that's the issue will be a huge clue.

No, this is all openGL. i downloaded from the artifacts just like you

Isaac0-dev:
Unfortunately I can't reproduce this issue you're talking about @Beckowl. I tried using Wine, my Windows 10 VM and a family member's Windows 10 computer. Each test I tried, I got perfect frame pacing as expected. I tried VSync on with manual/auto mode, and uncapped with vsync just produces a 60fps. 
So I believe I've run out of options to test locally at least. 

I don't want to merge this without at least understanding the issue you're experiencing. I could try making minor changes to the code and you could download the artifacts and test? Honestly without your help I can only speculate.

Isaac0-dev:
@Beckowl I just pushed a change and I'm wondering if you can test it please? I forgot that current builds of coop do a double buffer swap and I want to see if that was the reason

Isaac0-dev:
nevermind don't bother with that. i was wrong, and it won't help.

Isaac0-dev:
the latest change checks if vsync is going to delay for us. If it will, we don't do any delaying. This should hopefully fix the windows 11 windowed mode with vsync issue. But I can't test that.

Isaac0-dev:
Sure

Isaac0-dev:
The manual mode interp breaking is technically a bug but I don't know if it's worth fixing because it may compromise a feature, and it only happens if you go higher than your actual frame rate.

Beckowl:
> Now I'm confused. Are you getting the 30fps with vsync on in windowed mode anymore?

no, now it stays slightly below 60 fps (50-57 fps). I can reach 60+  if vsync is off. Everything else works fine

PR #908, opened 2025-08-07
fix require

Beckowl:
``require`` looked for module names by checking if a file's relative path ended with a string like ``/src/foo/bar.lua`` (note the slash). However, mod files are stored without a leading slash (e.g., ``src/foo/bar.lua``), so the match would fail

### DIFF
--- a/src/pc/lua/smlua_hooks.c
+++ b/src/pc/lua/smlua_hooks.c
@@ -100,6 +100,7 @@ int smlua_hook_event(lua_State* L) {
 
     hook->reference[hook->count] = ref;
     hook->mod[hook->count] = gLuaActiveMod;
+    hook->modFile[hook->count] = gLuaActiveModFile;
     hook->count++;
 
     return 1;
--- a/src/pc/lua/smlua_require.c
+++ b/src/pc/lua/smlua_require.c
@@ -114,14 +114,19 @@ static int smlua_custom_require(lua_State* L) {
 
     struct Mod* activeMod = gLuaActiveMod;
     if (!activeMod) {
-        LOG_LUA("require() called outside of mod context");
+        LOG_LUA_LINE("require() called outside of mod context");
+        return 0;
+    }
+
+    if (str_ends_with(moduleName, "/") || str_ends_with(moduleName, "\\")) {
+        LOG_LUA_LINE("cannot require a directory");
         return 0;
     }
 
     // find the file in mod files
     struct ModFile* file = smlua_find_mod_file(moduleName);
     if (!file) {
-        LOG_LUA("module '%s' not found in mod files", moduleName);
+        LOG_LUA_LINE("module '%s' not found in mod files", moduleName);
         return 0;
     }
 


PeachyPeachSM64:
Now, there is another problem

Since the file look-up code uses `str_ends_with`, it will incorrectly load the wrong lua file in this specific situation:
```lua
-- src/test.lua

print("src/test.lua")
```
```lua
-- src/another-test.lua

print("src/another-test.lua")
```
```lua
-- main.lua

local a = require("src/test.lua")
local b = require("test.lua")
```
Output:
```
src/test.lua
src/another-test.lua
```

Beckowl:
I rewrote ``smlua_find_mod_file``, require now works like this:
- By default, paths are relative to the current file’s directory.
- If you want to require relative to the mod root, prefix the path with a slash ``/`` (``\\`` works too). For example, ``require("/file4")``
- You can use ``..`` to go up directories, for example ``require("../file3")``

And now it won't load the wrong files

djoslin0:
Needs testing with windows/other. I thought that coop replaces all backslashes with forward slashes for the modfile's relativePath. 
You use different separators depending on platform which makes me think a mod will load correctly on one platform but not another. I'm not 100% sure on this though

Beckowl:
> Needs testing with windows and other OS (linux/mac). I thought that coop replaces all backslashes with forward slashes for the modfile's relativePath. You use different separators depending on platform which makes me think a mod will load correctly on one platform but not another. I'm not 100% sure on this though

I only did that because `normalize_path` does it that way. In theory, it should work since i normalized pretty much everything.
Id that doesn't work, i could do a fuzzy comparison instead

Beckowl:
it works fine on both windows and linux, remote mods work fine too

Beckowl:
I don’t think this should be called ``require`` since it works so differently from Lua’s, which is misleading because people will expect it to behave like Lua require. It also expects paths to be in a different format (relative paths, go up directories, ``/`` for absolute paths) which will mess with IDEs. Calling it ``import`` or something else would make more sense imo.

djoslin0:
You need to add this line, either I forgot it in my original review, or it was missed somewhere along the chain:
```diff
diff --git a/src/pc/lua/smlua_hooks.c b/src/pc/lua/smlua_hooks.c
index f08e156d3..0c69fd9bd 100644
--- a/src/pc/lua/smlua_hooks.c
+++ b/src/pc/lua/smlua_hooks.c
@@ -100,6 +100,7 @@ int smlua_hook_event(lua_State* L) {

     hook->reference[hook->count] = ref;
     hook->mod[hook->count] = gLuaActiveMod;
+    hook->modFile[hook->count] = gLuaActiveModFile;
     hook->count++;

     return 1;
```

The only other issue I found is that require should probably reject trailing '/'s:
`require('a/')` should not load `a.lua`

Other than that looks good. I replaced all 569 `require()`s in my bowling mod and it seems to work. I also tested a bunch of various paths and everything worked as expected

PeachyPeachSM64 (approved):

PR #913, opened 2025-08-12
GEO_BONE GeoLayout command

ManIsCat2:
Adds 2 new GeoLayout commands: GEO_BONE and GEO_BONE_WITH_SCALE. 
they support Animation, Translation, Rotation and Scale all at once.

the reason i split them into two commands is because originally, i made GEO_BONE to make decomp romhacks that use it,
easier to port, otherwise i would've added the scale param to GEO_BONE.

GEO_BONE doesn't let you select the default scale value, while GEO_BONE_WITH_SCALE does,
they both are Autogen'd and work with GEO_ASM.

i have also added another anim flag: ANIM_FLAG_BONE_TRANS, it allows for per bone rotation AND translation in animations, i added support for this in my fast64 fork. i was planning to do Scale but that would increase the anim size too much.


include/geo_commands.h
```diff
@@ -457,5 +457,27 @@ enum SkyBackgroundParams {
...
+ *   u32 displayList: display list segmented address
+ */
+#define GEO_BONE(layer, tx, ty, tz, rx, ry, rz, displayList) \
+    CMD_BBH(0x24, (0x00 | layer), 0x0000), \
+    CMD_HHHHHH(tx, ty, tz, rx, ry, rz), \
+    CMD_PTR(displayList)
+#define GEO_BONE_WITH_SCALE(layer, tx, ty, tz, rx, ry, rz, scale, displayList) \
+    CMD_BBH(0x24, (0x80 | layer), scale), \
+    CMD_HHHHHH(tx, ty, tz, rx, ry, rz), \
+    CMD_PTR(displayList)
```
PeachyPeachSM64: you're storing `scale` as a u16 but read it later as a 16.16 fixed point float later (u32), truncating the integer part
also, why no separate x, y, z scale?

src/engine/geo_layout.c
```diff
@@ -830,6 +831,38 @@ void geo_layout_cmd_node_generated_ext(void) {
...
+*/
+void geo_layout_cmd_bone(void) {
+    struct GraphNodeBone *graphNode;
+    Vec3s translation;
+    Vec3s rotation;
+    s32 drawingLayer = cur_geo_cmd_u8(0x01);;
+    f32 scale = 1.0f;
+    if (drawingLayer & 0x80) {
+        drawingLayer &= 0x0F;
+        scale = cur_geo_cmd_u32(0x02) / 65536.0f;
```
PeachyPeachSM64: you're reading scale+tx, due to how you packed the data

src/engine/geo_layout.c
```diff
@@ -830,6 +831,38 @@ void geo_layout_cmd_node_generated_ext(void) {
...
 }
 
+/*
+  0x24: Create a scene graph node that is rotated by the object's animation + an initial rotation.
+*/
+void geo_layout_cmd_bone(void) {
+    struct GraphNodeBone *graphNode;
+    Vec3s translation;
+    Vec3s rotation;
+    s32 drawingLayer = cur_geo_cmd_u8(0x01);;
```
PeachyPeachSM64: extra semi-colon

src/engine/graph_node.h
```diff
@@ -369,6 +370,16 @@ struct GraphNodeCullingRadius
     u8 pad1E[2];
 };
 
+/* Same as GraphNodeAnimatedPart but it has a Initial rotation value.*/
```
PeachyPeachSM64:
```
/** GraphNodeAnimatedPart with initial rotation and scale values.
 */
```

src/game/rendering_graph_node.c
```diff
@@ -1535,6 +1537,76 @@ void geo_process_held_object(struct GraphNodeHeldObject *node) {
     }
 }
 
+/**
+ * Render an animated part with an Initial rotation value.
+ */
```
PeachyPeachSM64:
```
/**
 * Render an animated part with initial rotation and scale values.
 */
```

src/game/rendering_graph_node.c
```diff
@@ -1535,6 +1537,76 @@ void geo_process_held_object(struct GraphNodeHeldObject *node) {
...
+
+    mtxf_rotate_xyz_and_translate(matrix, translation, rotation);
+    mtxf_scale_vec3f(matrix, matrix, scaleVec);
+    mtxf_mul(gMatStack[gMatStackIndex + 1], matrix, gMatStack[gMatStackIndex]);
+
+    mtxf_rotate_xyz_and_translate(matrix, translationPrev, rotationPrev);
+    /* From geo_process_scale:
+        TODO: this fails because multiple player models reuse the same scalenode
+        mtxf_scale_vec3f(matrix, matrix, prevScaleVec);
+    */
```
PeachyPeachSM64: TODO

PeachyPeachSM64 (changes requested):

src/game/rendering_graph_node.c
ManIsCat2: Thats also there in geo_process_scale though

PeachyPeachSM64: uh
how is it working for translation and rotation but not scale?
player models also share the same translation and rotation nodes, no?

src/game/rendering_graph_node.c
ManIsCat2: done

src/engine/graph_node.h
ManIsCat2: done

src/engine/geo_layout.c
ManIsCat2: oopsie

src/engine/geo_layout.c
ManIsCat2: fixed

include/geo_commands.h
ManIsCat2: fixed

src/game/rendering_graph_node.c
ManIsCat2: just tried it, when you punch/kick, the other players also have their hands and foot scaled up

Isaac0-dev: > uh how is it working for translation and rotation but not scale? player models also share the same translation and rotation nodes, no?

@PeachyPeachSM64 
yes, you're right. but there's a hack that disables interpolation of remote players rotation. see `geo_mario_tilt_torso` and `geo_mario_head_rotation` (search for the comment saying "only interpolate angles for the local player")
Translation is not unique to the player except for whole model translation obviously. But whole model translation and rotation is not shared because `GraphNodeObject`s are unique per object. 
The whole problem is interpolation. Coop relies on the `prev*` fields in a graph node's structure to be true to that object, so it assumes the same model will not be reused several times for different objects. 
Overall there's several issues here lol.

It seems the easy fix is to add the hack that's in those other functions i mentioned to `geo_mario_hand_foot_scaler`. Then you can re-enable proper interpolation of scale nodes. 
To be clear, `geo_process_bone` does NOT need the same hack. It should interpolate the scale correctly.
It'd be nice to clear this up properly with correct interpolation of remote players too eventually.

Isaac0-dev (approved):

PR #920, opened 2025-08-16
Static Object Collision

Cooliokid956:
# Static Object Collision
## `struct StaticObjectCollision *load_static_object_collision()`
does most of what load_object_collision_model does, but dumps its surfaces into the static surface partitions and returns a StaticObjectCollision which can later be fed into
## `void toggle_static_object_collision(struct StaticObjectCollision col, bool tangible)`
loops through all surfaces belonging to the SOC struct, ticking their `SURFACE_FLAG_INTANGIBLE` flag bit inversely to `tangible`
## `struct Surface *get_static_object_surface(struct StaticObjectCollision *col, u32 index)`
used to get surfaces belonging to `StaticObjectCollision`s

```lua
local col

-- in some object behavior
if not col then col = load_static_object_collision() end

-- to toggle collision
toggle_static_object_collision(col, ((get_global_timer() // 30) % 2 ~= 0))
```

### DIFF
--- a/src/pc/lua/smlua_utils.c
+++ b/src/pc/lua/smlua_utils.c
@@ -3,7 +3,6 @@
 #include "audio/external.h"
 
 u8 gSmLuaConvertSuccess = false;
-u16 gSmLuaFreeLot = LOT_SURFACE;
 
 #define VEC3F_BUFFER_COUNT 64
 static Vec3f sVec3fBuffer[VEC3F_BUFFER_COUNT] = { 0 };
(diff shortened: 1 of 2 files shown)


theclashingfritz (changes requested):
While I agree the ordering of a,b,g,r for RGBA colors is.. bad. It doesn't in my eyes alone vindicate disabling field sorting outright. Unless you provide more concrete examples I would recommended a struct exclusive disable of it.

PeachyPeachSM64:
please separate static object collision and doc changes into different PRs

Cooliokid956:
> please separate static object collision and doc changes into different PRs

Okay

PR #921, opened 2025-08-16
Restored OOB death + De-sorted fields + Read-Only marked as `protected`

Cooliokid956:
## Restored OOB death
this only happens when you're alone in a level
image image

Mario now bubbles (if enabled) when warping back to start after going OOB

---
## De-sorted fields
image

Nobody asked for this

---
## Added `protected` scope to field autogen
during the de-sorting tangent, I learned about the `protected` scope and added it to read-only fields
image image


src/game/mario.c
```diff
@@ -1574,12 +1574,14 @@ resetGoto:;
...
         m->faceAngle[1] = m->spawnInfo->startAngle[1];
+        if (mario_can_bubble(m)) mario_set_bubbled(m);
         struct Surface* floor = NULL;
         find_floor(m->pos[0], m->pos[1], m->pos[2], &floor);
         if (floor == NULL) {
-            level_trigger_warp(m, WARP_OP_DEATH);
+deathWarp:  level_trigger_warp(m, WARP_OP_DEATH);
         } else {
             goto resetGoto;
         }
```
PeachyPeachSM64: gotos are ugly
```c
} else if (!is_other_player_active()) {
    level_trigger_warp(m, WARP_OP_DEATH);
} else {
    vec3s_to_vec3f(m->pos, m->spawnInfo->startPos);
    m->faceAngle[1] = m->spawnInfo->startAngle[1];
    if (mario_can_bubble(m)) mario_set_bubbled(m);
    struct Surface* floor = NULL;
    find_floor(m->pos[0], m->pos[1], m->pos[2], &floor);
    if (floor == NULL) {
        level_trigger_warp(m, WARP_OP_DEATH);
    } else {
        return update_mario_geometry_inputs(m);
    }
}
```
and remove resetGoto

PeachyPeachSM64 (changes requested):

src/game/mario.c
Cooliokid956: Okay

autogen/lua_constants/built-in.lua
```diff
@@ -17,8 +17,8 @@ _ReadOnlyTable = {
         local _table = rawget(t, '_table')
         return _table[k]
     end,
-    __newindex = function (t,k,v)
-    end
+    __newindex = function (_,k,_) error('Attempting to modify key `' .. k .. '` of read-only table') end,
+    __metatable = false
```
Isaac0-dev, line 21: metatable should be `nil`, not `false`

src/game/mario.c
```diff
@@ -1573,17 +1572,20 @@ resetGoto:;
...
+    } else {  
+        vec3s_to_vec3f(m->pos, m->spawnInfo->startPos);  
+        m->faceAngle[1] = m->spawnInfo->startAngle[1];  
+        if (mario_can_bubble(m)) mario_set_bubbled(m);  
+        struct Surface* floor = NULL;  
+        find_floor(m->pos[0], m->pos[1], m->pos[2], &floor);  
+        if (floor == NULL) {  
+            level_trigger_warp(m, WARP_OP_DEATH);  
+        } else {  
+            return update_mario_geometry_inputs(m);
```
Isaac0-dev: this implies that you're returning the value that `update_mario_geometry_inputs` returns, but it doesn't return anything!
prefer:
```C
update_mario_geometry_inputs(m);
return;
```

src/game/mario.c
```diff
@@ -1573,17 +1572,20 @@ resetGoto:;
...
-        } else {
-            goto resetGoto;
-        }
-    }
+    } else if (!is_other_player_active()) {  
+        level_trigger_warp(m, WARP_OP_DEATH);  
+    } else {  
+        vec3s_to_vec3f(m->pos, m->spawnInfo->startPos);  
+        m->faceAngle[1] = m->spawnInfo->startAngle[1];  
+        if (mario_can_bubble(m)) mario_set_bubbled(m);
```
Isaac0-dev: wrap this in curly braces

Isaac0-dev (changes requested):

autogen/lua_constants/built-in.lua
PeachyPeachSM64: `__metatable = false` is a Lua thing, it's used to block `getmetatable` and `setmetatable`

Isaac0-dev: Yes, I'm saying it should be `__metatable = nil`
Because it's normally a table, not a Boolean.

PeachyPeachSM64: no, that has nothing to do with what `__metatable` is
`__metatable = false` is what makes the metatable immutable
otherwise you would be able to call `setmetatable` to override `__index`, `__newindex` and other metamethods

Isaac0-dev: oh i see. I was confused with the way we do it in C with userdata objects.

src/game/mario.c
Cooliokid956: okay!

src/game/mario.c
Cooliokid956: okay!

PR #922, opened 2025-08-18
Print lua log_to_console() messages to debug log, add LOG_WARNING()

bradleycha:
The lua function `log_to_console()` only logs to the game's djui console, but not to `stdout`.  This is important for logging with headless servers, as we can capture lua script messages in server logs.

This patch first adds `LOG_WARNING()` to the debug log functions (`LOG_INFO()`, `LOG_ERROR()`), alongside the corresponding config file key.  I grouped this in alongside this patch to avoid hacky stuff such as using `LOG_ERROR()` in-place of warning logs.

Using the above, this patch then logs all calls to `log_to_console()` to `stdout` via the debug log functions.  Each log message is prefixed with the file path to the mod calling `log_to_console()` to better clarify which mod is creating the log message.

### DIFF
--- a/autogen/convert_functions.py
+++ b/autogen/convert_functions.py
@@ -223,6 +223,7 @@ manual_index_documentation = """
    - [smlua_anim_util_register_animation](#smlua_anim_util_register_animation)
    - [level_script_parse](#level_script_parse)
    - [log_to_console](#log_to_console)
+   - [log_to_stdout](#log_to_stdout)
    - [add_scroll_target](#add_scroll_target)
    - [collision_find_surface_on_ray](#collision_find_surface_on_ray)
    - [cast_graph_node](#cast_graph_node)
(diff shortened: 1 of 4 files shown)


bradleycha:
> You can already use `print` in Lua to send messages to `stdout`. So you can do this in Lua:
> 
> ```lua
> function debug_print(...)
> 	log_to_console(...)
> 	print(...)
> end
> 
> debug_print("This message appears in stdout and the in game console.")
> ```
> 
> Perhaps I've misunderstood what you wanted though?

Using `print()` directly does work, but it would require manually formatting information like date, time, and location.  In addition, it would require editing the source code of mods which you want logging information from.  The benefit of these patches is it uses the game's existing logging code to solve all of the above.

For example:

`print("Hello")` displays "Hello" to `stdout`
`log_to_console("Hello")` displays "Wed Aug 20 19:27:00 2025 [00] [INFO] smlua_functions.c: mod.lua: Hello" to `stdout`

Isaac0-dev:
Having 2 places to log messages is handy.
You shouldn't force people to have those messages sent to stdout. 

So I think there's a few options:
- Expose a new function to Lua which prints to both stdout and the in game console.
- Create an option for developers to pick the behavior of `log_to_console`.
- Simply use the tools we have in Lua already - this is a very viable option, it's not that hard to patch a mod to print to both. for example:
```lua
local old_log_to_console = log_to_console
function log_to_console(str, ...)
    local dt = get_date_and_time()
    local year = dt.year + 1900
    local month = dt.month + 1
    local day = dt.day
    local hour = dt.hour
    local minute = dt.minute
    local second = dt.second
    local timestamp = string.format("%04d-%02d-%02d %02d:%02d:%02d", year, month, day, hour, minute, second)
    local debug_str = string.format("[%s] %s", timestamp, str)
    old_log_to_console(debug_str)
    print(debug_str)
end
```

bradleycha:
Revised version created in #926

PR #924, opened 2025-08-19
Static Object Collision

Cooliokid956:
# Static Object Collision
## `struct StaticObjectCollision *load_static_object_collision()`
does most of what load_object_collision_model does, but dumps its surfaces into the static surface partitions and returns a StaticObjectCollision which can later be fed into
## `void toggle_static_object_collision(struct StaticObjectCollision col, bool tangible)`
loops through all surfaces belonging to the SOC struct, ticking their `SURFACE_FLAG_INTANGIBLE` flag bit inversely to `tangible`
## `struct Surface *get_static_object_surface(struct StaticObjectCollision *col, u32 index)`
used to get surfaces belonging to `StaticObjectCollision`s

```lua
local col

-- in some object behavior
if not col then col = load_static_object_collision() end

-- to toggle collision
toggle_static_object_collision(col, ((get_global_timer() // 30) % 2 ~= 0))
```


PeachyPeachSM64:
> I'm also realizing there's the possibility of more dynamic collision being added mid-load via HOOK_ON_ADD_SURFACE

nope, recursive calls to `load_object_collision_model` are blocked

PeachyPeachSM64 (approved):

src/engine/surface_load.c
```diff
@@ -716,9 +762,12 @@ void load_object_surfaces(s16** data, s16* vertexData) {
 /**
  * Transform an object's vertices, reload them, and render the object.
  */
-void load_object_collision_model(void) {
+static void load_object_collision(bool isSOC) {
```
Isaac0-dev: This name change isn't ideal. `load_object_collision_model` is still an accurate descriptor for this function.
It should be `load_object_collision_model_` or `load_object_collision_model_internal`.

src/game/memory.c
```diff
@@ -220,6 +220,24 @@ void *growing_array_alloc(struct GrowingArray *array, u32 size) {
     return NULL;
 }
 
+void growing_array_move(struct GrowingArray *array, u32 from, u32 to, u32 count) {
+    if (array && array->buffer && count > 0 && (from + count) <= array->count && (to + count) <= array->count) {
+        void *temp[count];
```
Isaac0-dev: using the stack for this is convenient and fast but
this is considered bad practice because if count is too big, you may overflow the stack.

src/pc/lua/smlua_cobject.c
```diff
@@ -656,6 +656,12 @@ int smlua__eq(lua_State *L) {
     return 1;
 }
 
+int smlua__bnot(lua_State *L) {
```
Isaac0-dev, line 653: this is _bitwise_ not
were you expecting regular `not`?

src/game/memory.c
```diff
@@ -220,6 +220,24 @@ void *growing_array_alloc(struct GrowingArray *array, u32 size) {
     return NULL;
 }
 
+void growing_array_move(struct GrowingArray *array, u32 from, u32 to, u32 count) {
+    if (array && array->buffer && count > 0 && (from + count) <= array->count && (to + count) <= array->count) {
```
Isaac0-dev: rather than this complicated condition, separate each condition into conditional return statements.
and check `to == from`.

src/game/memory.c
```diff
@@ -220,6 +220,24 @@ void *growing_array_alloc(struct GrowingArray *array, u32 size) {
...
+        void *temp[count];
+
+        // Copy elements to move to temporary buffer
+        memcpy(temp, array->buffer + from, sizeof(void *) * count);
+
+        // Remove copied elements from the array
+        memmove(array->buffer + from, array->buffer + (from + count), sizeof(void *) * (array->count - (from + count)));
+
+        // Make place for the copied elements
+        memmove(array->buffer + (to + count), array->buffer + to, sizeof(void *) * (array->count - (to + count)));
```
Isaac0-dev, line 240: is this function assuming you're only moving left?
The logic is wrong unless you move left. 

Firstly, lets say you want to move right:
When you do the first `memmove`, the destination to write to has actually moved by `count`. 
so, add `if (to > from) { to -= count; }` to fix that.

Also, undefined behavior if the destination is inside the original location.

src/pc/lua/smlua_cobject.c
Cooliokid956: We were `~`.

src/game/memory.c
PeachyPeachSM64: you're right
not about the bad practice, but about the risk of overflowing the stack

src/pc/lua/smlua_cobject.c
PeachyPeachSM64: `not` has no metamethod equivalent

src/game/memory.c
PeachyPeachSM64: how is this complicated? that's just a bounds check

src/game/memory.c
PeachyPeachSM64: no, the logic is right.
the function assumes `to` is the destination index **after** `from + count` have been removed from the buffer.
You can't move inside the original location.

example:
```
buffer = { 1, 2, 3, 4, 5, 6 }
from = 2
to = 3
count = 2

step 1:
buffer = { 1, 2, 3, 4, 5, 6 }
temp = { 3, 4 }

step 2:
buffer = { 1, 2, 5, 6, _, _ }

step 3:
buffer = { 1, 2, 5, _, _, 6 }

step 4:
buffer = { 1, 2, 5, 3, 4, 6 }
```

src/game/memory.c
Isaac0-dev: i used the wrong word, i just mean the long list of conditions in one line

PeachyPeachSM64: not a fan of multiple returns for something like this
that would make it less readable
I guess we could put the condition on 2 lines instead of one

Isaac0-dev: I'm fine with that

src/game/memory.c
Isaac0-dev: The logic is indeed incorrect. `to` need to be an expected index in the _original_ array. Not a precomputed index for after removal of data.
There's no documentation for that expectation, and the expectation is unreasonable. 
You're asking the user to precompute the `to` value after the buffer is adjusted, even though the buffer mutation happens inside `growing_array_move`.
It's better to accept a `to` which is adjusted for the buffer in it's current state when the user calls `growing_array_move`.
Then, adjust `to` when the data is removed from the array.
The final `memcpy` can mask incorrect behavior in some situations, but it's still incorrect behavior.

src/pc/lua/smlua_cobject.c
Isaac0-dev: you should be able to allow:
```lua
if not col then col = load_static_object_collision() end
```
Without the bitwise not
This should be done by modifying `smlua__get_field` to push nil when the object is freed.

Cooliokid956: I would have to index `col` to invoke that metamethod. Nothing changes because the CObject has already been referenced and assigned to `col`

Your method would require me to
```lua
if not col or not col.index then col = load_static_object_collision() end
```

This would quickly get ugly when used across different types

PeachyPeachSM64: > you should be able to allow:
> 
> ```lua
> if not col then col = load_static_object_collision() end
> ```
> 
> Without the bitwise not This should be done by modifying `smlua__get_field` to push nil when the object is freed.

`not` does not call `smlua__get_field`
Behavior is defined by Lua: operator `not` doesn't call any metamethod, it's programmed to return `true` only if the object has type `nil` or if it is the boolean `false`

src/engine/surface_load.c
PeachyPeachSM64: done

src/game/memory.c
PeachyPeachSM64: done

src/game/memory.c
PeachyPeachSM64: done

src/game/memory.c
PeachyPeachSM64: done

src/game/memory.c
```diff
@@ -220,6 +220,32 @@ void *growing_array_alloc(struct GrowingArray *array, u32 size) {
     return NULL;
 }
 
+void growing_array_move(struct GrowingArray *array, u32 from, u32 to, u32 count) {
+    if (array && array->buffer && count > 0 &&
+        (to < from || to > from + count) &&
+        (from + count) <= array->count && to <= array->count) {
```
Isaac0-dev, line 226: Hi there, @PeachyPeachSM64 this is for you.
This condition: `to <= array->count` should be `(to + count) <= array->count`, or you'll allow writing out of bounds. 

No offence but overall, I think this function isn't very readable at all and in fact very confusing even if it works. 
This functionality can be achieved with only one `memmove`, but I'm happy to let it slide unless you agree that it's worth revising the structure of the function. 
It could also check for a situation where `from == to`, to avoid unnecessary work as an optimisation.

PeachyPeachSM64: Hi
> This condition: to <= array->count should be (to + count) <= array->count, or you'll allow writing out of bounds.

No? `to` is moved back in bounds if it is after `from`, you're the one who suggested to do that because the user shouldn't have to take the displacement into account

> I think this function isn't very readable at all and in fact very confusing even if it works.

How is it not readable and confusing? There are only 3 instructions, each one is commented and does exactly what the function is supposed to do:
- Remove elements to move
- Move buffer to fill the gap
- Insert elements at the new place

> This functionality can be achieved with only one memmove

I'd like to know how you would achieve that. Despite its name, `memmove` *doesn't* move memory, but copies while halding correctly overlapping.

> It could also check for a situation where from == to, to avoid unnecessary work as an optimisation.

It's already handled by the second condition: `(to < from || to > from + count)`

djoslin0: I think Peachy is correct here, while I do think all of the bounds checking is tricky (and I had to recheck quite a few times.. because I thought there was an issue) I don't see a bug here.

Isaac0-dev: @PeachyPeachSM64 I can see you've merged it already and I imagine you're probably sick of dealing with this function at this point but respectfully I ask that you look into it again.

It seems I was wrong indeed, there is no way to use this function to write out of bounds. Like you observed, that's because it moves `to` before using it.

But there is actually still a bug. When moving from left to right, the function incorrectly disallows moving elements to the index directly after itself. 
This is due to the condition `(to < from || to > from + count)`.
Here's an example of what we expect to happen:
```
from = 2
to = 4
count = 2
buffer before: { A, B, C, D, E, F }
expected buffer after: { A, B, E, F, C, D }
```
Our function will actually reject that situation because it's directly after itself, but that's a valid scenario!
Now, filling in our condition: `(to < from || to > from + count)`, we have `(4 < 2 || 4 > 2 + 2)`
`4 < 2` is false, `4 > 2 + 2` is false!
So, simply adding an `=` sign to the condition fixes that: `(to < from || to >= from + count)`

---

> 
> This functionality can be achieved with only one memmove

> I'd like to know how you would achieve that. Despite its name, memmove doesn't move memory, but > > copies while halding correctly overlapping.


It's interesting because your function prevents copying a block inside itself `(to < from || to > from + count)`, even though it's engineered in such a way that it's technically capable of copying a block inside itself. 
So if that's an expected limitation, you can indeed do one `memmove`.

I wrote, tested and profiled several different versions of this function in the hopes of making a more readable version without any logic errors.

Here is a version that performs only one `memmove` per call, and it's the fastest out of all the methods I've tested, while remaining logically sound.
In my opinion, it's also the most readable because it splits up the left and right operations.
```c
// This function cannot move elements to an overlapping region
// `to` is the index in the array to move to before any operations
bool growing_array_move(struct GrowingArray *array, u32 from, u32 to, u32 count) {
    if (!array || !array->buffer) { return false; }
    if (count == 0) { return true; }
    if (from >= array->count) { return false; }
    if (from + count > array->count) { return false; }
    if (to > array->count) { return false; }
    if (to + count > array->count) { return false; }
    if (from == to) { return true; }
    if (to >= from && to < from + count) { return false; }

    u32 blockBytes = sizeof(void *) * count;
    void **temp = malloc(blockBytes);
    if (!temp) { return false; }

    // Copy elements to move to temporary buffer
    memcpy(temp, array->buffer + from, blockBytes);

    // Make place for the copied elements
    if (to > from) { // From left to right
        u32 moveCount = to - from;
        memmove(array->buffer + from, array->buffer + from + count, sizeof(void *) * moveCount);
    } else { // From right to left
        u32 moveCount = from - to;
        memmove(array->buffer + to + count, array->buffer + to, sizeof(void *) * moveCount);
    }

    // Insert copied elements
    memcpy(array->buffer + to, temp, blockBytes);

    free(temp);
    return true;
}
```
																																																																																																																																																																																																													
Here's a version that supports (and allows) copying a block into itself. It's the most flexible of them all while remaining logically sound.
```C
// This function can move elements to an overlapping region
// `to` is the index in the array to move to before any operations
bool growing_array_move(struct GrowingArray *array, u32 from, u32 to, u32 count) {
    if (!array || !array->buffer) { return false; }
    if (count == 0) { return true; }
    if (from >= array->count) { return false; }
    if (from + count > array->count) { return false; }
    if (to > array->count) { return false; }
    if (to + count > array->count) { return false; }
    if (from == to) { return true; }

    void **temp = malloc(sizeof(void *) * count);
    if (!temp) { return false; }

    // Copy elements to move to temporary buffer
    memcpy(temp, array->buffer + from, sizeof(void *) * count);

    // Remove copied elements from the array
    u32 tailCount = array->count - (from + count);
    if (tailCount > 0) {
        memmove(array->buffer + from, array->buffer + from + count, sizeof(void *) * tailCount);
    }

    // Make place for the copied elements
    u32 numToShift = (array->count - count) - to;
    if (numToShift > 0) {
        memmove(array->buffer + to + count, array->buffer + to, sizeof(void *) * numToShift);
    }

    // Insert copied elements
    memcpy(array->buffer + to, temp, sizeof(void *) * count);

    free(temp);
    return true;
}
```

I don't know which to recommend because they both have pros and cons. If you'd like the code I've been using to test, feel free to ask.

Isaac0-dev: @PeachyPeachSM64 I will soon push a commit with the second version of the function that I suggested (the one that allows copying a block inside itself), as it is the most robust version. So this is a "speak now or forever hold your peace" moment.

PeachyPeachSM64: Hi @Isaac0-dev, I'm sorry, but this is getting ridiculous
Are we seriously still arguing over 5 perfectly fine lines of code after a month?

First of all,
> But there is actually still a bug. When moving from left to right, the function incorrectly disallows moving elements to the index directly after itself.
> This is due to the condition (to < from || to > from + count).
> Here's an example of what we expect to happen:
> 
> from = 2
> to = 4
> count = 2
> buffer before: { A, B, C, D, E, F }
> expected buffer after: { A, B, E, F, C, D }

Index 4 is E, not the end of the buffer. So the expected output is still `{ A, B, C, D, E, F }`. The one you get is if the user assumes the `to` index refers to the buffer `{ A, B, E, F }`, i.e. after the buffer to copy is removed. **The very logic you told me to not use because it would be too confusing for the user and asked me to change, which I did.**

> Our function will actually reject that situation because it's directly after itself, but that's a valid scenario!

No, it's not. Just as shown above, the buffer stays the same, so no need to move stuff around. The index directly after itself would be 5, i.e. between E and F.

> It's interesting because your function prevents copying a block inside itself (to < from || to > from + count), even though it's engineered in such a way that it's technically capable of copying a block inside itself.

It's definitely not. What does copying a block inside itself even mean?

> I wrote, tested and profiled several different versions of this function in the hopes of making a more readable version without any logic errors.

There is no (more) logic errors in the original function. Don't fix that isn't broken.

> Here is a version that performs only one memmove per call, and it's the fastest out of all the methods I've tested, while remaining logically sound.
In my opinion, it's also the most readable because it splits up the left and right operations.

The fastest out of what? I don't see any benchmark result.
Also, not only this function uses the logic you told me to not use, but it cannot move the buffer if the index lands between from and to? What's the point? This is deliberately introducing an unintended behavior!

> Here's a version that supports (and allows) copying a block into itself. It's the most flexible of them all while remaining logically sound.

Again, what does that even mean?
This function is a carbon copy of the existing one, minus the `if (to > from) { to -= count; }`. Nothing new here.

> I don't know which to recommend because they both have pros and cons. If you'd like the code I've been using to test, feel free to ask.

Finally, I ran the following tests:

```
from = 2
to = 3
count = 2
buffer before: { 0, 1, 2, 3, 4, 5 }
```

Using your logic, expected buffer would be: `{ 0, 1, 4, 2, 3, 5 }`
Result of your first function: `{ 0, 1, 2, 3, 4, 5 }`
Result of your second function: `{ 0, 1, 4, 2, 3, 5 }`

```
from = 2
to = 0
count = 2
buffer before: { 0, 1, 2, 3, 4, 5 }
```

Logic does not matter here, since it's moving to the left, expected buffer would be: `{ 2, 3, 0, 1, 4, 5 }`
Result of your first function: `{ 2, 3, 0, 1, 4, 5 }`
Result of your second function: `{ 2, 3, 0, 1, 4, 5 }`

First function fails for first case, because of that unintended behavior.
Code is overall correct, but uses the wrong logic.

Out of curiosity, I did a benchmark of existing function vs solution 2: https://quick-bench.com/q/FucokCzxLQwHqSLfCFZdw2o8xzo
Unsurprisingly, both functions run the same.
Regarding assembly size, your solution is a bit bigger.

> I will soon push a commit with the second version of the function that I suggested (the one that allows copying a block inside itself), as it is the most robust version.

Please, don't. I don't want to continue debating on this, this is going nowhere. I hope you understand.

Isaac0-dev: Hi again @PeachyPeachSM64, I understand your frustration and I wish to respect it.
As requested I won't push a commit to force my view.
It's my fault this argument has gone on so long, I apologise for both that and the long messages.
I am not seeking to continue our argument, but I still believe there's an issue with the current `growing_array_move`, so if anything I'd like to at least document this.
I do hope that a definitive conclusion will be made at some point.
Your last message helped me to see what we're actually disagreeing on.

I also realise that I contradicted myself. When I suggested adding `if (to > from) { to -= count; }`, that was a mistake. That change made the code do what you said it should (your words: "the function assumes `to` is the destination index after `from + count` have been removed from the buffer."), but I later changed my mind about what `to` should be, and I'll get into that.

I think there is an unstated disagreement on what `to` should be. 
Essentially the question is:
- Should `to` be relative to the array _before_ any operation is performed?
- **Or**, should `to` be relative to the array _after collapsing the removed block_?

I now believe `to` should refer to the state of the array **before any operations**. This is the standard and user friendly approach, as users shouldn’t have to mentally simulate what the position would be mid-function.
I failed to mention this when I realised it, which has understandably led to confusion from both of us.
Both functions I wrote are based on this belief.

---
> What does copying a block inside itself even mean?

That means a rightward operation where `to > from` and `to < from + count`.
We're moving a block to a destination that is inside the original block’s range.
You already showed an example of this in your first test, but here's one of my own:
```
buffer: {A, B, [C, D], E, F}
from: 2  (C)
count: 2 (block: [C, D])
to: 3
expected result (if using a version that allows this operation) : {A, B, E, [C, D], F}
```
In that example we've moved the buffer 1 space to the right even though the block's size is 2, so it's overlapping itself.

You were right to discard my claims to better performance. While I profiled each version I discussed, the difference is very marginal and inconsistent. I should not have brought up speed. Apologies for that.

> Also, not only this function uses the logic you told me to not use, but it cannot move the buffer if the index lands between from and to? What's the point? This is deliberately introducing an unintended behavior!

If you're talking about `if (to >= from && to < from + count) { return false; }`, that's basically the De Morgan's negation of `(to < from || to > from + count)`, in the current version.
I may have misunderstood what you're saying, but it ironically seems like the first function I suggested was criticised for "not supporting copying a block inside itself", but I included that limitation in that version intentionally because your version shares this issue. You asked for a version with one `memmove`, and I showed that; limitations and all.
Overall, the best version is the 2nd function I provided as it doesn't have this limitation; your first test reveals this.

I will also acknowledge that the only use of `growing_array_move` in the code base is `load_static_object_collision` in `surface_load.c`, and it never performs a rightward operation. It's always doing a right to left operation where `to <= from`.
This means that there will never be any bugs that arise from `load_static_object_collision`, using any version of `growing_array_move` that we have had so far. 
But I'm focusing on making sure the function is correct for any future uses of the function as well. If someone else uses the function later on, it should immediately behave as one would expect.

I want to reiterate that I'm not trying to argue.
I want the best correct outcome, not whatever makes me sound right.
If I'm wrong, so be it.

PR #926, opened 2025-08-21
Create Lua function log_to_stdout()

bradleycha:
Revised version of #922

Adds a new Lua function `log_to_stdout()` which sends log messages to the game's debug logger (`LOG_INFO`, `LOG_ERROR`, etc).

Of the options, I think it's best to make a new function.  If we modify the parameters for `log_to_console()`, this could potentially break existing mods which might have funky wrappers that break with a 3rd optional parameter.  I don't like having to replicate the formatting in Lua scripts, as we then have to duplicate the formatting code for every single mod that wants to print logging information.

### DIFF
--- a/src/pc/lua/smlua_functions.c
+++ b/src/pc/lua/smlua_functions.c
@@ -796,6 +796,37 @@ int smlua_func_log_to_console(lua_State* L) {
     return 1;
 }
 
+int smlua_func_log_to_stdout(lua_State* L) {
+    if (!smlua_functions_valid_param_range(L, 1, 2)) { return 0; }
+
+    int paramCount = lua_gettop(L);
+
+    const char* message = smlua_to_string(L, 1);
+    if (!gSmLuaConvertSuccess) { LOG_LUA("log_to_stdout: Failed to convert parameter 1 for function"); return 0; }
+
+    enum ConsoleMessageLevel level = CONSOLE_MESSAGE_INFO;
+    if (paramCount >= 2) {
+        level = smlua_to_integer(L, 2);
+        if (!gSmLuaConvertSuccess) { LOG_LUA("log_to_stdout: Failed to convert parameter 2 for function"); return 0; }
+    }
+
+    const char* origin = gLuaActiveModFile->relativePath;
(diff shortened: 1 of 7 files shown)


src/pc/lua/smlua_functions.c
```diff
@@ -796,6 +796,37 @@ int smlua_func_log_to_console(lua_State* L) {
...
+    const char* message = smlua_to_string(L, 1);
+    if (!gSmLuaConvertSuccess) { LOG_LUA("log_to_stdout: Failed to convert parameter 1 for function"); return 0; }
+
+    enum ConsoleMessageLevel level = CONSOLE_MESSAGE_INFO;
+    if (paramCount >= 2) {
+        level = smlua_to_integer(L, 2);
+        if (!gSmLuaConvertSuccess) { LOG_LUA("log_to_stdout: Failed to convert parameter 2 for function"); return 0; }
+    }
+
+    const char* origin = gLuaActiveModFile->relativePath;
```
djoslin0, line 813: Need to check for NULL
```C
const char* origin = (gLuaActiveModFile != NULL) ? gLuaActiveModFile->relativePath : "Unknown";
```

src/pc/lua/smlua_functions.c
```diff
@@ -796,6 +796,37 @@ int smlua_func_log_to_console(lua_State* L) {
...
+        case CONSOLE_MESSAGE_INFO:
+            LOG_INFO("%s: %s", origin, message);
+            break;
+        case CONSOLE_MESSAGE_WARNING:
+            LOG_WARNING("%s: %s", origin, message);
+            break;
+        case CONSOLE_MESSAGE_ERROR:
+            LOG_ERROR("%s: %s", origin, message);
+            break;
+    }
```
djoslin0, line 825: Should add a `default` case that throws an error if the passed in `level` is invalid

djoslin0 (changes requested):

Isaac0-dev:
This doesn't seem necessary because I think you can implement this functionality within Lua.

PR #931, opened 2025-08-25
Fix the FPS display for displaying FPS counts higher than 5 digits

Multi-Volt:
This is a very simple change (just increasing the fps display box width a bit).

Before:
Screenshot From 2025-08-25 11-51-19

After:
Screenshot From 2025-08-25 11-53-56


PeachyPeachSM64:
> couldn't you just render more digits in the existing black box? Or not enough space?

not enough space

maybe also consider capping the display to 99999

PR #936, opened 2025-08-30
Fix obj_init_animation

Baconator2558:
This is _long_ overdue.


src/game/object_helpers.c
```diff
@@ -811,7 +811,7 @@ void obj_set_gfx_pos_from_pos(struct Object *obj) {
 
 void obj_init_animation(struct Object *obj, s32 animIndex) {
     if (!o || !obj) { return; }
```
PeachyPeachSM64: might remove the !o as well

PeachyPeachSM64 (changes requested):

PR #947, opened 2025-09-12
custom player dialog

kikweeisW2012:
This is a version of sm64coopdx that I've been working on for a little while that instead of pulling all of the dialog assets from the rom file it instead pulls the dialogs that say the word "mario" from a custom variant table and instead of saying "Mario! It that really you?" it replaces the word "mario" as the player your playing as ("Toad! It that really you?"). I have added a dialog table of all of the dialogs that include the word "mario" and it tracks what player model is being used (CT_TOAD or CT_WALUIGI) and pulls the dialog depending on the model. There is a dialog ranking system so it pulls from mods first, then the player models, and then the original. This works globally so that it is specific for each player and it updates every frame so it stays in-sync. There is room left for future updates and I want to make it so it can track player models from mods automatically without creating dialog tables inside of the mods.
Screenshot (295)


kikweeisW2012:
👍> Why do you need to update every frame? You only need to update the dialogue when the string is actually getting pulled.

AgentXLP:
There are much better and cleaner ways to go about this, sorry.

PR #954, opened 2025-09-24
Ingame Chat – Major Overhaul & Quality of Life Improvements

iZePlayzYT:
#### Commands & Tab Completion

* **Reverse cycling**: Added **Shift+TAB** support for cycling **backwards** through commands, subcommands, and player names — similar to command prompts/terminals.
* **Help improvements**: Enhanced `/help`, `/?`, and `/` commands with better formatting and localized help text.
* **Command fix**: Updated `/psc` command registration so its usage displays as `[On|Off]` and tab completion works properly.
* **Command keybind**: Added new keybind to open chat with "/" pre-written for quick command access.
* **Tab completion preview**: Added preview for chat commands & subcommands.

#### UI & Chatbox

* **Chat UI redesign** with improved layout.
* **Character counter**: Added display of character amount in the chat input box.
* **Selection fix**: Fixed black text issue when selecting text in chat.
* **Prevent duplicate messages** in chat history.

#### Settings & Controls

* **Config options**:

  * Added **chat width setting**.
  * Added **hide chat setting**.
  * Added **hide char count setting**.
* **Improved checkbox text** for chat controls.

#### Translations

* Added **new and updated translations** across all supported languages.


sher-b:
thats not what is referred to by duplicate message. it's referring to duplicate messages when scrolling through your previous chat history, not the general chat history. its like minecraft, where you can still send duplicate messages and commands but you wont have to deal with each instance you sent the same message while scrolling

you should be able to find clips of it in #coop-discussion if youre still not convinced, but it should be clarified nonetheless

src/pc/djui/djui_slider.c
```diff
@@ -52,12 +73,43 @@ static void djui_slider_on_cursor_down(struct DjuiBase* base) {
...
     u32* value = slider->value;
-    f32 x = slider->rect->base.elem.x;
-    f32 w = slider->rect->base.elem.width;
-    f32 cursorX = gCursorX;
-    cursorX = fmax(cursorX, x);
-    cursorX = fmin(cursorX, x + w);
-    *value = ((cursorX - x) / w) * (max - min) + min;
+
+    bool fineAdjust = (gDjuiInputHeldShift != 0);
+    int newValue = (int)*value;
```
djoslin0, line 78: validate that `slider->value` is non-null before dereferencing.

src/pc/configfile.c
```diff
@@ -666,6 +687,23 @@ static void configfile_load_internal(const char *filename, bool* error) {
...
+                    configKeyChatCommand[0] = 0x002B; // '#' on QWERTZ (OEM_5 position)
+                    break;
+                default:
+                    configKeyChatCommand[0] = 0x0035; // '/' on US QWERTY
+                    break;
+            }
+#else
+            configKeyChatCommand[0] = 0x0035; // '/' default on non-Windows
+#endif
+        }
```
djoslin0, line 708: this code is very unusual... and feels misplaced
either set a default key the normal way or move this code into a relevant part of the code base. I don't think keyboard layout code belongs in configfile.c

It's also strange to only have this behavior on windows

src/pc/djui/djui_inputbox.c
```diff
@@ -447,8 +457,63 @@ static void djui_inputbox_render_char(struct DjuiInputbox* inputbox, char* c, f3
     *additionalShift += charWidth;
 }
 
+static void djui_inputbox_render_preview_text(struct DjuiInputbox* inputbox) {
```
djoslin0, line 460: dead code?
It looks like the only call site is commented out

djoslin0 (changes requested):
I recommend splitting this into smaller, focused PRs (e.g., tab completion changes, UI redesign, config changes, slider enhancements) for easier review. As-is it's hard to understand the full impact of the changes.

AgentXLP:
I agree with dj, split into smaller PRs this is a lot and has a sizable amount of features or little things I don't agree with

iZePlayzYT:
i will try do to this in august @AgentXLP @djoslin0

Issue #949, opened 2025-09-16
⭐ Star Symbol kerning is wrong with "Aliased" font.


bluelightzero:
<img width="668" height="46" alt="Image" src="https://github.com/user-attachments/assets/f8aa88ea-f2f4-4466-9211-beaf76f40e94" />

<img width="139" height="38" alt="Image" src="https://github.com/user-attachments/assets/6735a630-62e9-4b7c-88da-7949c8ada189" />

<img width="666" height="44" alt="Image" src="https://github.com/user-attachments/assets/471dabf9-abe4-4499-be67-d37dbb43ceac" />

<img width="104" height="39" alt="Image" src="https://github.com/user-attachments/assets/72b7dd24-8231-48cb-82a5-235970dfe04a" />

Notice how the P overlaps the star when on "Aliased".

It should have the same spacing as the "Normal" font.

Also, just a nitpick, but calling the font "aliased" is incorrect. It should probably be called "smooth" or "high-res", but that is a separate issue.

sher-b:
i mean you could make a pull request to fix the issue yourself if you've located it and its just about changing one integer

PR #956, opened 2025-09-24
Aliased font kerning fixes

ItsJaid:
Fixes the kerning for the colon and star symbol characters in the aliased DJUI font.

Original colon (with semicolon comparison... why did they have different kerning anyway?)
image
Fixed colon
image

Original star
image
Fixed star
image

Solves issue #949.


Beckowl:
by the way there's an issue with double quotation marks (`<<` & `>>`) im replaced dialogs where the top pixels are cut off, it doesn't happen in vanilla dialogs

perhaps you know how to fix that?

ItsJaid:
> by the way there's an issue with double quotation marks (`<<` & `>>`) in replaced dialogs where the top pixels are cut off, it doesn't happen in vanilla dialogs
> 
> perhaps you know how to fix that?

i do not, i don't even know how to code or understand code at all. only reason i was able to make this pr is because it was the easiest fix imaginable, especially after reading the issue i tagged in which op posted the fix

djoslin0:
looks good to me

PeachyPeachSM64 (approved):

PR #973, opened 2025-09-30
Allow clients to view server settings

EmeraldLoc:
Couple things about this:

- The client uses `gServerSettings` to get the current server settings, so if lua changes it, it will show what lua has, not what your config has, this can technically cause the server and the client to have different menus, but the client is more accurate
- This was put into a separate file (`djui_panel_client_server_settings`). Originally I was going to merge them into one file, but since they are both very different, doing a lot of checks to see if you're the server or not gets ugly fast, and I couldn't find a cleaner solution than just putting them in different files

Here's a video of it in action:

https://github.com/user-attachments/assets/b1ac91cc-4f9a-4f19-a736-562ff31b03ac


Amy54Desu:
This is a GOod idea, a lot of the time I do not know server settings (outside of player count) until I get a star or something

src/pc/djui/djui_panel_client_server_settings.c
```diff
@@ -0,0 +1,70 @@
...
+
+void djui_panel_client_server_settings_create(struct DjuiBase* caller) {
+    struct DjuiThreePanel* panel = djui_panel_menu_create(DLANG(HOST_SETTINGS, SETTINGS), false);
+    struct DjuiBase* body = djui_three_panel_get_body(panel);
+    {
+        char* iChoices[3] = { DLANG(HOST_SETTINGS, NONSOLID), DLANG(HOST_SETTINGS, SOLID), DLANG(HOST_SETTINGS, FRIENDLY_FIRE) };
+        struct DjuiSelectionbox* selectionbox1 = djui_selectionbox_create(body, DLANG(HOST_SETTINGS, PLAYER_INTERACTION), iChoices, 3, &gServerSettings.playerInteractions, NULL);
+        djui_base_set_enabled(&selectionbox1->base, false);
+
+        unsigned int sKnockbackIndex = (gServerSettings.playerKnockbackStrength <= 20)
```
djoslin0: must be declared as static instead of on the stack since the djui elements outlive the stack

src/pc/djui/djui_panel_client_server_settings.c
```diff
@@ -0,0 +1,70 @@
...
+                        : ((gServerSettings.playerKnockbackStrength <= 40) ? 1 : 2);
+        char* kChoices[3] = { DLANG(HOST_SETTINGS, WEAK), DLANG(HOST_SETTINGS, NORMAL), DLANG(HOST_SETTINGS, TOO_MUCH) };
+        struct DjuiSelectionbox* selectionbox2 = djui_selectionbox_create(body, DLANG(HOST_SETTINGS, KNOCKBACK_STRENGTH), kChoices, 3, &sKnockbackIndex, NULL);
+        djui_base_set_enabled(&selectionbox2->base, false);
+
+        char* pChoices[2] = { DLANG(HOST_SETTINGS, CLASSIC_PVP), DLANG(HOST_SETTINGS, REVAMPED_PVP) };
+        struct DjuiSelectionbox* selectionbox3 = djui_selectionbox_create(body, DLANG(HOST_SETTINGS, PVP_MODE), pChoices, 2, &gServerSettings.pvpType, NULL);
+        djui_base_set_enabled(&selectionbox3->base, false);
+
+        unsigned int sStayInLevelAfterStar = gServerSettings.stayInLevelAfterStar;
```
djoslin0: must be declared as static instead of on the stack since the djui elements outlive the stack
or it should use `gServerSettings.stayInLevelAfterStar` directly if possible?

djoslin0 (commented):
Other than these memory issues, looks good

djoslin0:
looks good to me

PR #981, opened 2025-10-14
Bring back unused missing_texture to DynOS

ManIsCat2:
uses the missing texture in the dynos texture code, now the parser will not stop generating if the texture couldnt be found, it will just replace it with the 4x4 pink-black missing texture

edit: this pr also adds shift S and T support in the pc renderer

### DIFF
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -1151,6 +1151,24 @@ static void OPTIMIZE_O3 gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t 
         if (use_texture) {
             float u = (v_arr[i]->u - rdp.texture_tile.uls * 8) / 32.0f;
             float v = (v_arr[i]->v - rdp.texture_tile.ult * 8) / 32.0f;
+
+            uint8_t shifts = rdp.texture_tile.shifts;
+            uint8_t shiftt = rdp.texture_tile.shiftt;
+            if (shifts != 0) {
+                if (shifts <= 10) {
+                    u /= 1 << shifts;
+                } else {
+                    u *= 1 << (16 - shifts);
+                }
+            }
+            if (shiftt != 0) {
+                if (shiftt <= 10) {
+                    v /= 1 << shiftt;
+                } else {
+                    v *= 1 << (16 - shiftt);
+                }
+            }
+
             if ((rdp.other_mode_h & (3U << G_MDSFT_TEXTFILT)) != G_TF_POINT) {
                 // Linear filter adds 0.5f to the coordinates (why?)
                 u += 0.5f;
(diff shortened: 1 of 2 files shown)


data/dynos_bin_tex.cpp
```diff
@@ -4,6 +4,15 @@ extern "C" {
 #include "stb/stb_image_write.h"
 }
 
+// 4x4 pink-black checkerboard texture to indicate missing textures
+#define MISSING_TEX_SIZE 64
+static const uint8_t sMissingTexture[MISSING_TEX_SIZE] = {
```
PeachyPeachSM64, line 9: why not keeping MISSING_W and MISSING_H? they are used later in the code
and use proper type for texture
`static const Texture sMissingTexture[MISSING_W * MISSING_H * 4]`

data/dynos_bin_tex.cpp
```diff
@@ -52,17 +62,22 @@ static TexData* LoadTextureFromFile(GfxData *aGfxData, const char* aFile) {
 
         // The file does not exist in either spot!
         if (!_File) {
-            PrintDataError("  ERROR: Unable to open file at \"%s\" or \"%s\"", _Filename.c_str(), _ActorFilename.c_str());
-            return NULL;
+            PrintError("  ERROR: Unable to open file at \"%s\" or \"%s\"", _Filename.c_str(), _ActorFilename.c_str());
+            _UnknownTexture = true;
```
PeachyPeachSM64, line 66: I don't agree with this change.
The parser should fail if the texture is missing.
I don't see any case where it is acceptable to replace a missing texture in a model with the pink-black checkerboard.

data/dynos_bin_tex.cpp
```diff
@@ -269,10 +284,20 @@ DataNode<TexData>* DynOS_Tex_Load(BinFile *aFile, GfxData *aGfxData) {
...
+                _Node->mData->mRawData   = Array<u8>(_RawData, _RawData + (_Node->mData->mRawWidth * _Node->mData->mRawHeight * 4));
+                free(_RawData);
+            } else {
+                // Handle missing texture
+                _Node->mData->mRawWidth  = 4;
+                _Node->mData->mRawHeight = 4;
+                _Node->mData->mRawFormat = G_IM_FMT_RGBA;
+                _Node->mData->mRawSize   = G_IM_SIZ_32b;
+
+                _Node->mData->mRawData = Array<u8>(sMissingTexture, sMissingTexture + MISSING_TEX_SIZE);
```
PeachyPeachSM64, line 299: this is where MISSING_W and MISSING_H are supposed to be used
also use `ARRAY_COUNT(sMissingTexture)` to point to the end of the buffer

src/pc/gfx/gfx_pc.c
```diff
@@ -1161,6 +1151,24 @@ static void OPTIMIZE_O3 gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t
...
+                }
+            }
+            if (shiftt != 0) {
+                if (shiftt <= 10) {
+                    v /= 1 << shiftt;
+                } else {
+                    v *= 1 << (16 - shiftt);
+                }
+            }
+
```
PeachyPeachSM64, line 1171: please move this change to a separate PR

PR #983, opened 2025-10-16
HOOK_OBJECT_UPDATE

Blockyyy:
should be self explanatory, will prevent having to loop through all the objects in lua

The before update hook might not be needed?, unsure about that will remove if needed


src/engine/behavior_script.c
```diff
@@ -1381,6 +1382,8 @@ cur_obj_update_begin:;
     smlua_call_behavior_hook(&gCurBhvCommand, gCurrentObject, false);
     gCurrentObject->curBhvCommand = gCurBhvCommand;
 
+    smlua_call_event_hooks(HOOK_BEFORE_OBJECT_UPDATE, gCurrentObject);
```
PeachyPeachSM64: at this point the object has already run its behavior code

src/engine/behavior_script.c
```diff
@@ -1293,6 +1293,7 @@ static BhvCommandProc BehaviorCmdTable[BEHAVIOR_CMD_TABLE_MAX] = {
 void cur_obj_update(void) {
     if (!gCurrentObject) { return; }
     // Don't update if dormant
+
```
PeachyPeachSM64: unnecessary new line
it separates the code from its comment

PR #986, opened 2025-10-18
djui_hud_is_mouse_locked()

Cooliokid956:
Adds a total of 1 function:
# `bool djui_hud_is_mouse_locked()`
Returns the DJUI mouse lock status

Tip: This won't return the true state of mouse lock, so make sure to use `get_first_person_enabled()` and `camera_config_is_mouse_look_enabled()` as well

---
This also fills in some descriptions left out by Isaac in his viewport/scissor functions

### DIFF
--- a/autogen/lua_definitions/functions.lua
+++ b/autogen/lua_definitions/functions.lua
@@ -3915,13 +3915,13 @@ function djui_hud_get_raw_mouse_y()
 end
 
 --- @return boolean
---- Returns if the cursor is hidden and constrained to the window
+--- Checks if the cursor is locked to the window
 function djui_hud_is_mouse_locked()
     -- ...
 end
 
 --- @param locked boolean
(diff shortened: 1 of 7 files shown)


src/pc/djui/djui_hud_utils.c
```diff
@@ -297,6 +297,10 @@ f32 djui_hud_get_raw_mouse_y(void) {
     return mouse_y;
 }
 
+bool djui_hud_get_mouse_locked(void) {
```
PeachyPeachSM64: djui_hud_is_mouse_locked sounds better imo

PeachyPeachSM64 (changes requested):

src/pc/djui/djui_hud_utils.c
Cooliokid956: true

PR #989, opened 2025-10-24
Some syncing fixes for Bowser and Heave Ho

EmilyEmmi:
The main focus is on Bowser, who is tied to the most infamous desyncs of all. Here are the changes I have made:

- Bowser's held loop now update o.parentObj to the marioObj of whichever player is holding him. When Bowser is thrown, the game calls cur_obj_set_pos_relative_to_parent to set Bowser's location, which assumes that the parent object is whichever Mario is holding him. This may not be the case due to multiplayer, so doing this ensures that the parent object is set correctly.
- The closest player to Bowser now sends a packet when he hits a mine, hopefully reducing how often Bowser fights get softlocked in multiplayer.
- Fixed a bug where if the player talking to Bowser left the level, the entire fight would be softlocked. A new sync field was created to track which player is currently talking to Bowser, and if that player becomes inactive, Bowser will start attacking.

The minor change to Heave Ho was that he can only throw the player if the player is in ACT_GRABBED, fixing an issue where Heave Ho would launch players that weren't interacting with him (although, this only happens when m->heldByObj is erroneously set to the Heave Ho object, which shouldn't be happening in the first place).

### DIFF
--- a/src/game/behaviors/bowser.inc.c
+++ b/src/game/behaviors/bowser.inc.c
@@ -3,6 +3,7 @@ static u32 networkBowserAnimationIndex = 0;
 static u8 bowserIsDying = FALSE;
 static u8 bowserCutscenePlayed = FALSE;
 static u8 bowserIsCutscenePlayer = FALSE;
+static u8 bowserCutsceneGlobalIndex = UNKNOWN_GLOBAL_INDEX;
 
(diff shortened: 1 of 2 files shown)


sher-b:
WE'RE SO FUCKING BACK

PR #995, opened 2025-11-03
Fix djui hud texture unsupported formats

PeachyPeachSM64:
- `djui_hud_render_texture(_tile)` now accepts properly any `TextureInfo`
- Removed `bitSize`, added `format` and `size` to `TextureInfo`
- `texture_to_lua_table` can now convert any `Texture` into an RGBA table


src/pc/djui/djui_gfx.c
```diff
@@ -91,45 +91,84 @@ static const Vtx vertex_djui_image[] = {
     {{{ 0,  0, 0 }, 0, { 0,       0 }, { 0xff, 0xff, 0xff, 0xff }}},
 };
 
-const Gfx dl_djui_image[] = {
+const Gfx dl_djui_image_begin[] = {
     gsDPPipeSync(),
-    gsSPClearGeometryMode(G_LIGHTING),
+    gsSPClearGeometryMode(G_LIGHTING | G_CULL_BOTH),
```
Isaac0-dev: render both sides? why, it's meant to be 2D right?

src/pc/djui/djui_gfx.c
```diff
@@ -150,31 +189,18 @@ void djui_gfx_render_texture_tile(const Texture* texture, u32 w, u32 h, u32 bitS
 
     // I don't know why adding 1 to all of the UVs seems to fix rendering, but it does...
     // this should be tested carefully. it definitely fixes some stuff, but what does it break?
```
Isaac0-dev, line 141: probably can remove the second line of the comment here since you're re-purposing the function to only be used for fonts. The first line is sufficient.

src/pc/djui/djui_gfx.c
```diff
@@ -150,31 +189,18 @@ void djui_gfx_render_texture_tile(const Texture* texture, u32 w, u32 h, u32 bitS
...
-    gDPSetTextureOverrideDjui(gDisplayListHead++, texture, djui_gfx_power_of_two(w), djui_gfx_power_of_two(h), bitSize);
-	gDPLoadTextureBlockWithoutTexture(gDisplayListHead++, NULL, G_IM_FMT_RGBA, G_IM_SIZ_16b, 64, 64, 0, G_TX_CLAMP, G_TX_CLAMP, 0, 0, 0, 0);
-
-    *(gDisplayListHead++) = (Gfx) gsSPExecuteDjui(G_TEXOVERRIDE_DJUI);
-
+    gDPSetTextureFilter(gDisplayListHead++, G_TF_POINT);
+    gDPSetTextureOverrideDjui(gDisplayListHead++, texture, djui_gfx_power_of_two(w), djui_gfx_power_of_two(h), fmt, siz);
+    gSPDisplayList(gDisplayListHead++, dl_djui_image_begin);
     gSPVertexNonGlobal(gDisplayListHead++, vtx, 4, 0);
-    *(gDisplayListHead++) = (Gfx) gsSPExecuteDjui(G_TEXCLIP_DJUI);
```
Isaac0-dev, line 173: many of these display list changes I'm no expert on, but do we not need texture clipping any more? What if a font renders under another texture or something?

I can't necessarily say for myself, so all I can say is are you certain that it's okay to remove so many graphics settings for both this function and the real `djui_gfx_render_texture_tile`?

Isaac0-dev (changes requested):
the gore/hard mode nether animated effect is broken by your changes to texture tiling. Please try to keep popular mods in tact.

1.3.2:

Image

1.4 - with your changes:

Image


Also these changes have reintroduced a visual bug in mods that use texture tiling. 
You can see in this image some tiles (probably all) are 1 pixel to the right. This is "So Retro!" by EmmiEmily - this is a great mod to test your changes with!
By the way in the past this bug has been due to these lines that you changed:
```C
    f32 offsetX = (font ? -1024.0f / (f32)w : 0) + 1;
    f32 offsetY = (font ? -1024.0f / (f32)h : 0) + 1;
```

Image

The same issue is true in DJUI menus too:

Image

good news is I don't know of any mods that actually use `texture_to_lua_table` anymore.

Here are some comments from me but other than that, good progress!

src/pc/djui/djui_gfx.c
PeachyPeachSM64: negative scales invert triangles, which disappear if culling is enabled

src/pc/djui/djui_gfx.c
PeachyPeachSM64: good catch

src/pc/djui/djui_gfx.c
PeachyPeachSM64: I don't really know what texture clipping does
It was removed (commented) in dl_djui_image, so I just assumed it was leftover
I'll add it back

src/pc/djui/djui_gfx.c
Isaac0-dev: fair enough, i guess that means you're trying to allow negative scales, to render stuff back to front without a rotation

Isaac0-dev (approved):
yeah i guess we'll have to find a solution for that bug at some point. I thought it was a regression for this pr

PR #997, opened 2025-11-08
Add gsDPSetEnvRGB()

Baconator2558:
This command sets the RGB values of a material's environment color without setting the alpha. This is important, as functions like `geo_update_layer_transparency()` use the environment color alpha to control a model's overall opacity. However, if one of the model's materials sets its own environment color alpha, then that will take priority. This command allows these models to still use environment colors in their materials.


AgentXLP:
Handy

include/PR/gbi.h
```diff
@@ -126,6 +126,7 @@
 #ifdef F3DEX_GBI_2E
 /* extended commands */
 #define G_COPYMEM		0xd2
+#define G_SETENVRGB		0xd1
```
Isaac0-dev: I think these changes should be in `include/gbi/PR/gbi_extension.h`, where other custom changes have been put for coop.

Baconator2558: Aren't the player part functions also (at least partially) in gbi.h?

Isaac0-dev: yes that's true, and there's more custom stuff in gbi.h than the example you provided as well. But I guess I'm saying that `gbi_extension.h` _ought_ to be the place for new gbi code and macros.
Technically it doesn't matter, but we do have a separate place for custom stuff, and that's why I suggested it.
No big deal if you disagree though, this pr can probably be merged without changing it.

Baconator2558: I think the best option might be to leave it in gbi.h for now, and then move it and the other custom stuff all to `gbi_extension.h` at once in a separate PR, if that's fine with you.

Baconator2558: I made a separate branch with it in `gbi_extension.h`, so you can tell me which one you prefer, and then I'll update the PR accordingly.

PeachyPeachSM64: I agree with Isaac, please move the custom gfx command to gbi_extension.h
(and remove the #ifdef F3DEX_GBI_2E, it's not needed)

Baconator2558: Alright, did both of those changes.

Isaac0-dev (approved):

PR #1007, opened 2025-11-15
Fix rumble and add PS4/PS5 controller rumble.

gibsonpil:
Previously `thread6_rumble_loop()` was only being called once in `thread5_game_loop()`. This commit removes that call and adds one to `game_loop_one_iteration()`, creating the desired outcome of `thread6_runnable_loop()` running once per frame. This is how it worked prior to cece41f, which appears to be when rumble was broken.

Additionally, PS4 and PS5 rumble support is added. PS4 and PS5 rumble requires a feature called 'extended reporting', which can be enabled using two SDL hints. Since extended reported, once enabled, breaks compatibility with DirectInput until the controller is restarted, an option is also added to the configuration menu to toggle this functionality.


src/pc/configfile.c
```diff
@@ -267,6 +268,7 @@ static const struct ConfigOption options[] = {
     {.name = "rumble_strength",                .type = CONFIG_TYPE_UINT, .uintValue = &configRumbleStrength},
     {.name = "gamepad_number",                 .type = CONFIG_TYPE_UINT, .uintValue = &configGamepadNumber},
     {.name = "background_gamepad",             .type = CONFIG_TYPE_UINT, .boolValue = &configBackgroundGamepad},
+    {.name = "extended_reports",               .type = CONFIG_TYPE_UINT, .boolValue = &configExtendedReports},
```
PeachyPeachSM64: booleans should use CONFIG_TYPE_BOOL, not CONFIG_TYPE_UINT
(configBackgroundGamepad also has the wrong config type)

src/pc/pc_main.c
```diff
@@ -316,6 +316,8 @@ void produce_one_frame(void) {
 
     CTX_EXTENT(CTX_SMLUA, smlua_update);
 
+    thread6_rumble_loop(NULL);
```
PeachyPeachSM64: should be moved to `game_loop_one_iteration` to be profiled

PeachyPeachSM64 (changes requested):

PeachyPeachSM64:
> Additionally, PS4 and PS5 rumble support is added. PS4 and PS5 rumble requires a feature called 'extended reporting', which can be enabled using two SDL hints. Since extended reported, once enabled, breaks compatibility with DirectInput until the controller is restarted, an option is also added to the configuration menu to toggle this functionality.

About this point:
Do extended reports have to be enabled when the game is started? or can it be done when the controller is plugged in?
What do you mean by "until the controller is restarted"? When the physical controller is plugged out, or the whole application (i.e. the game) has to be closed and opened again?

gibsonpil:
> > Additionally, PS4 and PS5 rumble support is added. PS4 and PS5 rumble requires a feature called 'extended reporting', which can be enabled using two SDL hints. Since extended reported, once enabled, breaks compatibility with DirectInput until the controller is restarted, an option is also added to the configuration menu to toggle this functionality.
> 
> About this point: Do extended reports have to be enabled when the game is started? or can it be done when the controller is plugged in? What do you mean by "until the controller is restarted"? When the physical controller is plugged out, or the whole application (i.e. the game) has to be closed and opened again?

The `SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE` and `SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE` hints allow rumble support on PS4/PS5 controllers running over Bluetooth. Once extended reports are enabled, DirectInput doesn't work properly, and extended reports can't be disabled without power cycling the controller.

All that info is from the SDL2 Wiki. 
https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE
https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE

In the patch's current form you have to restart the game for extended reporting to enable, I can change that if you'd like though.

Alex793:
Hi, in V1.4 I don’t have any rumble on my Dualsense (natively or Xbox 360 controller emulated with DS4windows, tested also USB and Bluetooth), even with an Xbox Series controller, no rumble at all.

Is this normal ?

Also tested every older versions, no rumble.

On all other SM64 PC ports, rumble are OK so it’s only with sm64coopdx.

gibsonpil:
@PeachyPeachSM64 @Isaac0-dev should I remake this pull request to merge into dev instead of main? I just realized that's the way things are typically done in this repo.

PeachyPeachSM64:
hi, you're right, you need to target dev branch instead of main
sorry for not catching that earlier

djoslin0:
I'm not sure why rumble strength doesn't seem to work... but I don't think it's related to this PR.
I'd prefer extended reports was auto-detected instead of adding yet another confusing setting.
But I think overall its fine to go in.

EmeraldLoc (approved):
Looks good!

PR #916, opened 2025-08-15
new djui hud tools: viewport, scissor and render line

Isaac0-dev:
added these new djui hud tools for Lua:
- djui_hud_set_viewport
- djui_hud_reset_viewport
- djui_hud_set_scissor
- djui_hud_reset_scissor
- djui_hud_render_line - this code could already be done exactly. this is just to make rendering lines more efficient. 

Also made nametags use a viewport for scaling in the credits.

PR #1008, opened 2025-11-16
Fix djui menu flashing

djoslin0:
Move djui_reset_hud_params() to after the djui display list save point

    Resetting scissor/viewport before the save point caused the djui menu to rapidly
    flicker when interpolating (and an interpolated HUD texture was rendered)

    Thanks to PeachyPeach for finding the better fix

---

This bug was introduced in this commit:
    ac1cf57cf new djui hud tools: viewport, scissor and render line (#916)
   @Isaac0-dev 

---

I used this lua mod to test:
```Lua
t = 0
function on_hud_render()
    t = t + 0.1

    djui_hud_set_viewport(0, 0, 300, 300 + math.sin(t) * 20)
    djui_hud_render_texture_interpolated(gTextures.coin, 0, 0, 10, 10, 0, 0, 10, 10)

    djui_hud_set_resolution(RESOLUTION_N64)
    djui_hud_set_scissor(0, 0, 400, 50 + math.sin(t) * 20)
    djui_hud_render_texture_interpolated(gTextures.coin, 50, 0, 10, 10, 50, 0, 10, 10)
end

hook_event(HOOK_ON_HUD_RENDER, on_hud_render)
```

---

This must be merged (or a similar fix) before v1.4 releases

### DIFF
--- a/src/pc/djui/djui.c
+++ b/src/pc/djui/djui.c
@@ -176,10 +176,10 @@ void djui_reset_hud_params(void) {
 
 void djui_render(void) {
     if (!sDjuiInited || gDjuiDisabled) { return; }
-    djui_reset_hud_params();
 
     sSavedDisplayListHead = gDisplayListHead;
     gDjuiHudUtilsZ = 0;
+    djui_reset_hud_params();
 
     create_dl_ortho_matrix();
     djui_gfx_displaylist_begin();


PeachyPeachSM64:
While it does fix it, the flickering is actually due to another issue:
When the game is paused and the djui menu is open, `HOOK_ON_RENDER_HUD` hooks get called twice because of: `patch_interpolations` -> `patch_djui_interpolated` -> `djui_render` -> `smlua_call_event_hooks(HOOK_ON_HUD_RENDER, djui_reset_hud_params)`
All Lua logic runs at 30 fps, but this particular hook in this specific instance can run up to 60 fps (in your example mod, the coin "moves"  twice as fast when the djui menu is open, because `t` is incremented each time the hook runs)
Ideally only interpolation should be done without having to run the hook again

PeachyPeachSM64:
> Meaning if the display list generated by djui_render is slightly different than the previous drawing frame, the second function could destroy the display list (possibly overriding the gSPEndDisplayList command and leading to buffer overflow...)

This could also explain the random crashes when navigating in the pause menu.
Any reason why the whole djui display list needs to be recreated during interpolation?

djoslin0:
I can't think of an easy way to solve the issue where the hook is called more frequently when the djui menu is open, might need to save that for a later release

PR #1011, opened 2025-11-16
Fix sequenced background music when set in ON_LEVEL_INIT

djoslin0:
On previous release the following mod would correctly set the background music:
```Lua

hook_event(HOOK_ON_LEVEL_INIT, function ()
    set_background_music(0, SEQ_LEVEL_SLIDE, 0)
end)

hook_event(HOOK_ON_SYNC_VALID, function()
    if gNetworkPlayers[0].currLevelNum ~= LEVEL_CASTLE_COURTYARD then
        warp_to_level(LEVEL_CASTLE_COURTYARD, 1, 0)
    end
end)
```

The following commit caused the regression:
cee5957e1 Fix cap sequences not replaced/playing on dynos warp and level entry; Add `gLevelValues.shellSequence`

---

I'm unsure of the intention of the `sDynosWarpNodeNum == -1` and couldn't find the original PR, so I would need @PeachyPeachSM64 's input on whether this is a valid fix or not.


PeachyPeachSM64:
The `sDynosWarpNodeNum == -1` was meant to fix Cap musics not playing when warping via dynos

I would suggest to move the cap musics out of the condition, but I don't know the potential impacts
image

PeachyPeachSM64:
what's that second file?
developer/network.sh

Isaac0-dev (approved):

Issue #1010, opened 2025-11-16
HOOK_ON_HUD_RENDER is run faster than 30hz when the djui menu is open


djoslin0:
Problem:
`HOOK_ON_HUD_RENDER` is run faster than 30hz when the djui menu is open

Expected:
`HOOK_ON_HUD_RENDER` should always run at 30hz

---
The issue is basically that we rebuild the entire djui display list every interpolated frame when the djui menu is open, and need to call the hook currently. But this is bad.

See the following issue for details that PeachyPeach found:
#1008

PR #1014, opened 2025-11-16
Prevent cursor interpolation from causing crashes

djoslin0:
Previously djui_cursor_interp() called djui_cursor_update_position(), which is really bad, because the cursor can trigger events that would affect the display lists. The number of things in the display list must *not* change during interpolation, or we corrupt the list.

Steps to cause the crash before this change are simple, but maybe only on windows:
1. Open game
2. Open console
3. Wiggle mouse on console

image

### DIFF
--- a/src/pc/djui/djui_cursor.c
+++ b/src/pc/djui/djui_cursor.c
@@ -167,7 +167,6 @@ void djui_cursor_interp_before(void) {
 }
 
 void djui_cursor_interp(void) {
-    djui_cursor_update_position();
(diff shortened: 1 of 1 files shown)


djoslin0:
My intention was to get it to stop crashing for release, we can figure out how to make the cursor smooth again for a future release.

The whole djui interpolation system probably needs a rework, since it has more issues: 
#1010 
#1008

Edit -- I believe it still is interpolating though since prev cursor is updated once per game tick
And the part that isn't captured in this small code change is the 3+ hours it took to track down. Displaylist corruption is nasty

djoslin0:
closing in favor of #1018

PR #1018, opened 2025-11-17
a proper fix to djui cursor dl corruption

Isaac0-dev:
An alternative solution to https://github.com/coop-deluxe/sm64coopdx/pull/1014.
Which keeps smooth cursor rendering, while adding safe guards.
This makes djui interactions always update at 30hz instead of 60hz, to avoid more chances of display list corruption.

The cursor gfx is now placed in it's own static buffer to allow it to take up extra room if needed. If the 20 dl command buffer size is too small, we'll hear about it because it'll throw a system error.
I doubt it'll trigger though, in my testing it never needed more than 8.

### DIFF
--- a/src/pc/djui/djui_cursor.c
+++ b/src/pc/djui/djui_cursor.c
@@ -163,17 +163,21 @@ static void djui_cursor_update_position(void) {
 #endif
 }
 
+static void djui_cursor_render_cursor(void) {
+    gDisplayListHead = sDjuiCursorGfx;
+    djui_base_render(&sMouseCursor->base);
+    gSPEndDisplayList(gDisplayListHead++);
+    if (gDisplayListHead - sDjuiCursorGfx >= CURSOR_GFX_MAX_SIZE) {
+        sys_fatal("CURSOR_GFX_MAX_SIZE is too small! %lu", gDisplayListHead - sDjuiCursorGfx);
+    }
+}
+
 // This isn't actually interpolation, it just updates the cursor at a faster rate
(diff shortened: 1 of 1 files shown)


djoslin0:
I can confirm this no longer crashes. And it skips the djui interactable update loop during interpolation, which is my best guess at what caused the display list corruption.
So I believe this is a better fix than mine.

src/pc/djui/djui_cursor.c
```diff
@@ -161,25 +163,32 @@ static void djui_cursor_update_position(void) {
...
+    Gfx *savedDisplayListHead = gDisplayListHead;
+    gDisplayListHead = sDjuiCursorGfx;
     djui_base_render(&sMouseCursor->base);
-    sInterpCursor = gDisplayListHead != sSavedDisplayListHead; // Check that we actually rendered something
+    gSPEndDisplayList(gDisplayListHead++);
+    if (gDisplayListHead - sDjuiCursorGfx >= CURSOR_GFX_MAX_SIZE) {
+        sys_fatal("CURSOR_GFX_MAX_SIZE is too small! %lu", gDisplayListHead - sDjuiCursorGfx);
+    }
+    gDisplayListHead = savedDisplayListHead;
+    gSPDisplayList(gDisplayListHead++, sDjuiCursorGfx);
```
PeachyPeachSM64, line 190: duplicated code, would be better to put it in a function

PeachyPeachSM64 (changes requested):

PR #1022, opened 2025-11-17
Fix HOOK_ON_HUD_RENDER rate at 30hz

PeachyPeachSM64:
Fixes https://github.com/coop-deluxe/sm64coopdx/issues/1010

Not the best looking fix, but moving the hook djui stuff into its own display list would require to change the whole display list system

### DIFF
--- a/src/pc/djui/djui.c
+++ b/src/pc/djui/djui.c
@@ -18,6 +18,8 @@
 #include "pc/utils/misc.h"
 
 static Gfx* sSavedDisplayListHead = NULL;
+static Gfx* sHookHudRenderGfx = NULL;
+static size_t sHookHudRenderGfxSize = 0;
 
 struct DjuiRoot* gDjuiRoot = NULL;
 struct DjuiText* gDjuiPauseOptions = NULL;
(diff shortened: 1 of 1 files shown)


Isaac0-dev (changes requested):
I'm just not sure about the repeated call to `realloc`. 
You should only need to `realloc` if the existing buffer is too small for the display lists created by the hooks. You already have a `sHookHudRenderGfx` which is perfect for keeping track of the buffer size.

Still need to `memcpy` every tick, but it's better than the constant reallocation.

PeachyPeachSM64:
> I'm just not sure about the repeated call to `realloc`. You should only need to `realloc` if the existing buffer is too small for the display lists created by the hooks. You already have a `sHookHudRenderGfx` which is perfect for keeping track of the buffer size.
> 
> Still need to `memcpy` every tick, but it's better than the constant reallocation.

It avoids keeping an arbitrary large buffer at some point, but I guess it would be preferable to reduce memory allocations

Isaac0-dev (changes requested):
sorry for not thinking of this before, but can you `free(sHookHudRenderGfx);` in `network_shutdown`? Mods can render a huge amount, pushing up the side of our allocated buffer, which we never free.

PR #1026, opened 2025-11-20
Fix - Remove incorrect collision validation check

djoslin0:
A validation check in dynos_bin_col.cpp was wrong, which would block otherwise well formed collision files. Since collisions are parsed as s16, if you reach the bounds of s16 values will begin to overflow and the validation will catch it.

To work around this overflow issue, you need to output groups of triangles and vertices. However the existing check in ValidateColInit() would prevent this workaround. Now there is essentially no hard-coded limit when it comes to vertex and triangle counts of collision structs.


PeachyPeachSM64:
I see know, thanks

PeachyPeachSM64 (approved):

Isaac0-dev:
should have documented this! now it's implicit behavior.

PR #1035, opened 2025-11-24
Mod Audio overhaul and fixes

Cooliokid956:
THIS IS THE MOD AUDIO OVERHAUL WOOOOO

## NEW autogen feature: `OVERLOAD(name)`
allows you to combine multiple functions under a single name

# Mod audio changes:
- New function for [get/set] parity: `audio_stream_get_loop_points`
- Arguably better method for managing mod master/stream/sample volumes (sound groups, tie in very nicely with the post-established volume groups)
- Made Lua volume control functions use u8 instead of f32 (that's what they are intended to be)
- check out my new modaudio api 😁 😆
image

This is my test mod that showcases some features in the new API! :) :)
[testy.zip](https://github.com/user-attachments/files/28535157/testy.zip)


src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -19,6 +19,7 @@
 #include "pc/pc_main.h"
 #include "pc/fs/fmem.h"
 
+#pragma region m64
```
PeachyPeachSM64: `#pragma region` is a Visual Studio exclusive preprocessor code
please remove it

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -374,12 +385,7 @@ void audio_stream_destroy(struct ModAudio* audio) {
 void audio_stream_play(struct ModAudio* audio, bool restart, f32 volume) {
     if (!audio_sanity_check(audio, true, "play")) { return; }
     
-    if (configMuteFocusLoss && !WAPI.has_focus()) {
-        ma_sound_set_volume(&audio->sound, 0);
-    } else {
-        f32 musicVolume = (f32)configMusicVolume / 127.0f * (f32)gLuaVolumeLevel / 127.0f;
-        ma_sound_set_volume(&audio->sound, gMasterVolume * musicVolume * volume);
-    }
+    ma_sound_set_volume(&audio->sound, volume);
```
PeachyPeachSM64, line 524: so, if I understand correctly, final volume is `engine->volume * group->volume * sound->volume`?

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -423,13 +429,30 @@ void audio_stream_set_looping(struct ModAudio* audio, bool looping) {
...
+    ma_data_source_get_loop_point_in_pcm_frames(&audio->decoder, &loopStart, &loopEnd);
+
+    lua_newtable(L);
+    lua_pushinteger(L, loopStart);
+    lua_rawseti(L, -2, 1);
+    lua_pushinteger(L, loopEnd);
+    lua_rawseti(L, -2, 2);
+
+    return smlua_to_lua_table(L, -1);
+}
```
PeachyPeachSM64: returning a table for two values is a bit overkill
Lua functions can return multiple values, see `gfx_get_from_name` as example

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -423,13 +429,30 @@ void audio_stream_set_looping(struct ModAudio* audio, bool looping) {
...
+
+void audio_stream_set_loop_points(struct ModAudio* audio, s64 loopStart, OPTIONAL s64 loopEnd) {
     if (!audio_sanity_check(audio, true, "set stream loop points for")) { return; }
     
     u64 length; ma_data_source_get_length_in_pcm_frames(&audio->decoder, &length);
     if (loopStart < 0) loopStart += length;
     if (loopEnd <= 0) loopEnd += length;
```
PeachyPeachSM64: OPTIONAL doesn't do anything if not in .h
the `if`s are strange: it's supposed to put loopStart and loopEnd back in range, but never work for values negative enough?
I don't know how negative loop points are usually dealt with, but I think it should be something like "if loopStart is negative, set it to 0; if loopEnd is zero or negative, set it to length"

src/pc/lua/utils/smlua_collision_utils.c
```diff
@@ -209,7 +209,7 @@ Collision *smlua_collision_util_get_level_collision(u32 level, u16 area) {
     return dynos_level_get_collision(level, area);
 }
 
-void smlua_collision_util_find_surface_types(Collision* data) {
+LuaTable smlua_collision_util_find_surface_types(Collision* data) {
     lua_State* L = gLuaState;
```
PeachyPeachSM64, line 215: should NULL check L

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -605,33 +623,32 @@ void audio_sample_play(struct ModAudio* audio, Vec3f position, f32 volume) {
...
+}
+
+void audio_custom_update_mute(bool mute) {
+    if (!sModAudioPool) { return; }
+
+    if (sModAudioMute != mute) {
+        sModAudioMute = mute;
+        ma_engine_set_volume(&sModAudioEngine, mute ? 0 : gMasterVolume);
     }
 }
```
PeachyPeachSM64: don't use stinky statics when it's not needed
update the volumes when they change
```c
bool shouldMute = (configMuteFocusLoss && !WAPI.has_focus());

// Update master volume
f32 masterVolume = shouldMute ? 0 : ((f32)configMasterVolume / 127.0f * (f32)gLuaVolumeMaster / 127.0f);
gMasterVolume = masterVolume;
if (!sModAudioPool) { return; }
if (ma_engine_get_volume(&sModAudioEngine) != masterVolume) {
    ma_engine_set_volume(&sModAudioEngine, masterVolume);
}

// Update music volume
f32 musicVolume = (f32)configMusicVolume / 127.0f * (f32)gLuaVolumeLevel / 127.0f;
if (ma_sound_group_get_volume(&sModAudioStreamGroup) != musicVolume) {
    ma_sound_group_set_volume(&sModAudioStreamGroup, musicVolume);
}

// Update sound volume
f32 sfxVolume = (f32)configSfxVolume / 127.0f * (f32)gLuaVolumeSfx / 127.0f;
if (ma_sound_group_get_volume(&sModAudioSampleGroup) != sfxVolume) {
    ma_sound_group_set_volume(&sModAudioSampleGroup, sfxVolume);
}
```
should do the trick.
Then you just have to call `audio_custom_update_volume` in `buffer_audio` to update it properly every frame

PeachyPeachSM64 (changes requested):

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: Correct!
engine = master
group =  music/sfx

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: I would've done that but autogen doesn't allow it... maybe with a `RETURN(n)` macro?

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: Ok 😢

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: Well that was quite the exercise :D 
image
image
image
I believe that `gfx_get_from_name` and `vtx_get_from_name` can be relocated to `smlua_gfx_utils.c` now

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: Fixed these

src/pc/lua/utils/smlua_collision_utils.c
Cooliokid956: Ok.

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: I wouldn't want to call more code than needed inside such a frequented function, sounds intensive

Isaac0-dev: C is much faster than Lua. You're not doing expensive checks here. Peachy is suggesting to check the volume before setting it too. If you're so worried about performance, you can introduce your own variables to compare the state of the volume to check if they actually changed without pinging miniaudio's library.

src/pc/lua/utils/smlua_audio_utils.c
PeachyPeachSM64: that's so great

idiotforge:
pragma balls is a good commit name

Isaac0-dev (commented):
in my opinion this is more of a "miscellaneous tweaks" pr in general. I agree with the non audio related adjustments, but they aren't related to audio. can we separate them please?

djoslin0:
I agree with Isaac, split the PR into multiple. Only have one PR per enhancement. Combining stuff makes it harder to review

Isaac0-dev:
merge conflicts!

Cooliokid956:
I have decided that I will put a bit more into this

Cooliokid956:
I think it is ready! @coderabbitai Hit it!

src/pc/djui/djui_hud_utils.c
```diff
@@ -722,6 +722,16 @@ void djui_hud_print_text(const char* message, f32 x, f32 y, f32 scaleX, f32 scal
...
 }
 
+void djui_hud_print_text_uniform(const char* message, f32 x, f32 y, f32 scale) {
+    if (message == NULL) { return; }
+
+    if (djui_hud_text_font_is_legacy()) {
+        scale *= 0.5f;
+    }
+
+    djui_hud_print_text_internal(message, x, y, scale, scale, NULL);
```
PeachyPeachSM64: please do `djui_hud_print_text(message, x, y, scale, scale)` instead

src/pc/djui/djui_hud_utils.c
```diff
@@ -750,6 +760,32 @@ void djui_hud_print_text_interpolated(const char* message, f32 prevX, f32 prevY,
...
+        interp->posY.curr = y;
+        interp->scaleX.prev = prevScale;
+        interp->scaleY.prev = prevScale;
+        interp->scaleX.curr = scale;
+        interp->scaleY.curr = scale;
+        interp->width = font->defaultFontScale;
+        interp->height = font->defaultFontScale;
+    }
+
+    djui_hud_print_text_internal(message, x, y, scale, scale, interp);
```
PeachyPeachSM64: same here
`djui_hud_print_text_interpolated(message, prevX, prevY, prevScale, prevScale, x, y, scale, scale)`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -182,20 +182,171 @@ u8 smlua_audio_utils_allocate_sequence(void) {
 ///////////////
 
 // Optimization: disable spatialization for everything as it's not used
-#define MA_SOUND_STREAM_FLAGS (MA_SOUND_FLAG_NO_SPATIALIZATION | MA_SOUND_FLAG_STREAM)
-#define MA_SOUND_SAMPLE_FLAGS (MA_SOUND_FLAG_NO_SPATIALIZATION | MA_SOUND_FLAG_NO_PITCH | MA_SOUND_FLAG_DECODE) // No pitch, pre-decode audio samples
+#define MA_SOUND_FLAGS (MA_SOUND_FLAG_NO_SPATIALIZATION | MA_SOUND_FLAG_NO_DEFAULT_ATTACHMENT)
 
 static ma_engine sModAudioEngine;
+static const char* sModAudioTypes[] = { "sound", "sample", "stream" };
```
PeachyPeachSM64: `static const char *sModAudioTypes`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -182,20 +182,171 @@ u8 smlua_audio_utils_allocate_sequence(void) {
...
+static ma_sound_group sModAudioChannels[3];
 static struct DynamicPool *sModAudioPool;
 static bool sModAudioShuttingDown = false;
 
+static bool audio_sanity_check(struct ModAudio* audio, u8 type, const char* action) {
```
PeachyPeachSM64: `struct ModAudio *audio`
`const char *action`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -182,20 +182,171 @@ u8 smlua_audio_utils_allocate_sequence(void) {
...
+
+// MA calls the end callback from its audio thread
+// Use mutexes to be sure we don't try to delete the same memory at the same time
+#include <pthread.h>
+static pthread_mutex_t sSoundCopyMutex = PTHREAD_MUTEX_INITIALIZER;
+static struct ModAudio *sSoundCopyFreeTail = NULL;
+
+// Called whenever a sample copy finishes playback (called from the miniaudio thread)
+// removes the copy from its linked list, and adds it to the pending list
+static void audio_destroy_copy(struct ModAudio* copy) {
```
PeachyPeachSM64: `struct ModAudio *copy`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -182,20 +182,171 @@ u8 smlua_audio_utils_allocate_sequence(void) {
...
+            copy->prev = sSoundCopyFreeTail;
+            sSoundCopyFreeTail->next = copy;
+        }
+        sSoundCopyFreeTail = copy;
+    }
+
+    pthread_mutex_unlock(&sSoundCopyMutex);
+}
+
+static void audio_destroy_copy_callback(void* userData, UNUSED ma_sound* sound) {
```
PeachyPeachSM64: well, pointers
I'm not filling the entire review with these
`void *userData`, NOT `void* userData`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -182,20 +182,171 @@ u8 smlua_audio_utils_allocate_sequence(void) {
...
+        copy->next = NULL;
+        copy->prev = NULL;
+        copy->parent = NULL;
+    
+        // add copy to free list
+        if (sSoundCopyFreeTail) {
+            copy->prev = sSoundCopyFreeTail;
+            sSoundCopyFreeTail->next = copy;
+        }
+        sSoundCopyFreeTail = copy;
```
PeachyPeachSM64, line 236: I hate linked lists with a passion just for this kind of non-sense
But, oh well, if it works, it works

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -182,20 +182,171 @@ u8 smlua_audio_utils_allocate_sequence(void) {
...
+
 static void smlua_audio_custom_init(void) {
     sModAudioPool = dynamic_pool_init();
 
     ma_result result = ma_engine_init(NULL, &sModAudioEngine);
     if (result != MA_SUCCESS) {
         LOG_ERROR("failed to init Miniaudio: %d", result);
     }
+
+    for (u8 i = 0; i < 3; i++) {
```
PeachyPeachSM64: ARRAY_COUNT(sModAudioChannels) instead of 3

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -393,399 +518,279 @@ struct ModAudio* audio_load_internal(const char* filename, bool isStream) {
...
-    ma_decoder_uninit(&audio->decoder);
-    if (audio->buffer) {
-        free(audio->buffer);
-        audio->buffer = NULL;
+    f32 dist = 0;
+    if (gCamera) {
+        f32 dX = position[0] - gCamera->pos[0];
+        f32 dY = position[1] - gCamera->pos[1];
+        f32 dZ = position[2] - gCamera->pos[2];
+        dist = sqrtf(dX * dX + dY * dY + dZ * dZ);
```
PeachyPeachSM64: or just `dist = vec3f_dist(position, gCamera->pos)`

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -795,32 +800,28 @@ void audio_custom_shutdown(void) {
...
                 ma_decoder_uninit(&copy->decoder);
                 free(copy);
                 copy = cprev;
             }
-            audio->sampleCopiesTail = NULL;
-            pthread_mutex_unlock(&sSampleCopyMutex);
-
+            audio->copiesTail = NULL;
+            pthread_mutex_unlock(&sSoundCopyMutex);
+
```
PeachyPeachSM64: whitespaces

src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -831,29 +832,30 @@ void audio_custom_shutdown(void) {
...
 void smlua_audio_custom_deinit(void) {
     if (sModAudioPool) {
-        pthread_mutex_lock(&sSampleCopyMutex);
+        pthread_mutex_lock(&sSoundCopyMutex);
         sModAudioShuttingDown = true;
-        pthread_mutex_unlock(&sSampleCopyMutex);
-
+        pthread_mutex_unlock(&sSoundCopyMutex);
+        
+        for (u8 i = 0; i < 3; i++) { ma_sound_group_uninit(&sModAudioChannels[i]); }
```
PeachyPeachSM64: ARRAY_COUNT(sModAudioChannels)

src/pc/utils/miniaudio.h
```diff
@@ -15,13 +15,15 @@ GitHub:        https://github.com/mackron/miniaudio
...
 
+// OUR DEFINES
 #define MA_NO_RESOURCE_MANAGER
 #define MA_NO_GENERATION
 #define MA_NO_ENCODING
 
 // enable Vorbis decoding (provides ogg audio decoding support) for miniaudio
 #define STB_VORBIS_HEADER_ONLY
 #include "stb_vorbis.c"
+// OUR DEFINES
```
PeachyPeachSM64, line 26: what does that mean?

PeachyPeachSM64 (changes requested):

src/pc/utils/miniaudio.h
Cooliokid956: It means that these are defines that WE (I) manually added to the file and must preserve when updating miniaudio
They were defined to not include unused junk in the build

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: I didn't do this

src/pc/lua/utils/smlua_audio_utils.c
Cooliokid956: Ok

PeachyPeachSM64 (approved):

xLuigiGamerx (commented):
Merge conflicts

Cooliokid956:
> Merge conflicts

No

xLuigiGamerx (approved):
Merge resolutions

ManIsCat2:
m,

EmeraldLoc (approved):
Tested and works, nothing looks wrong except possibly using `pthread_mutex_lock/unlock` functions raw rather than creating a `ThreadHandle` and using the `MUTEX_LOCK/UNLOCK` macros, but in this case it looks fine because mini audio is handling the actual thread where the func is called, not coop. Api looks a good bit cleaner so I like this. I did NOT review any code related to autogen changes, I looked at the audio stuff and tested that and that only

PR #1036, opened 2025-11-24
Adds compilation fixes for ARM in headless mode

### LINKED ISSUE #1034: Missing symbol FLT_EPSILON
https://github.com/coop-deluxe/sm64coopdx/blob/983932214adc178a66c9d9970cb0afbcd8429985/src/engine/math_util.c#L830

Hi. Love your work and I've been following for a long time now. I just ran into an issue when compiling release 1.4 using `TARGET_RPI=1` and `HEADLESS=1`, it seems that `math_util.c` is missing a symbol. If I manually edit the constant to point to the global, prefixed and suffixed with underscores, everything compiles well. I had to squelch a warning with `-Wno-stringop-overflow`.  Happy to cut a PR in the morning.

nullset2:
Should fix #1034. It seems sm64coopdx won't compile in `HEADLESS=1` and `TARGET_RPI=1` without these changes. Testing the waters here and I hope to be able to contribute way more in the coming days. Thank you.

### DIFF
--- a/src/engine/math_util.c
+++ b/src/engine/math_util.c
@@ -827,7 +827,7 @@ OPTIMIZE_O3 bool mtxf_inverse_non_affine(OUT Mat4 dest, Mat4 src) {
             if (fabsf(aug[i][k]) > fabsf(aug[piv][k])) { piv = i; }
         }
 
-        if (fabsf(aug[piv][k]) < FLT_EPSILON) { return false; }
+        if (fabsf(aug[piv][k]) < __FLT_EPSILON__) { return false; }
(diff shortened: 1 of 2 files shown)


paradox9955:
I'm curious as to why we still have the headless make flag. It's a cliopt now, so the make flag should just be removed.

paradox9955:
> FWIW I think the flag is still useful because Linux ARM builds are not vended in the releases.

Yeah but that doesn't mean you can't compile without the make flag and then use headless the same way you can in releases

PR #1040, opened 2025-11-27
some fixes for dynos, and a crash fix

Isaac0-dev:
I tested using `ASAN=1` as a build flag and encountered warnings and errors, this resolves these issues plus a null pointer crash fix.

Fixes the following DynOS issues, found through `ASAN=1`:
```
/usr/include/c++/15/bits/basic_string.h:1282:26: runtime error: execution reached an unreachable program point
```
This error actually hinted at a double free, caused within the DynOS `Array` class.
The DynOS `Array` class was doing shallow copies of objects, when it's used in DynOS as if it did deep copies.
It's been changed to perform deep copies, but this is ought to be a temporary fix, since the DynOS `Array` class was designed for speed and now I'm making it slower.
Right now, memory safety is better than performance.
Some places in DynOS were misusing `Array`, when they should have been using a class designed for copying of complex structures. Replaced with `std::vector` in most places.

---
```
data/dynos_level.cpp:57:43: runtime error: load of misaligned address 0x00000349d702 for type 'u64', which requires 8 byte alignment
0x00000349d702: note: pointer points here
00 00 1f 08 01 00 00 00 00 00 80 00 4b 03 00 00 00 00 06 08 00 00 00 00 00 00 c0 cc 49 03 00 00
^
```
This error was fixed by letting `memcpy` handle the unaligned address.

---
```
data/dynos_bin_pointer.cpp:16:13: runtime error: member access within null pointer of type 'struct Lights1'
```
Started handling light pointers lazily by calling their parser function, which already returns itself if it already exists.

### DIFF
--- a/data/dynos.cpp.h
+++ b/data/dynos.cpp.h
@@ -216,30 +216,22 @@ public:
 
     inline Array(const std::initializer_list<T> &aList) : mBuffer(NULL), mCount(0), mCapacity(0) {
         Resize(aList.size());
-        for (s32 i = 0; i < mCount; ++i) {
-            mBuffer[i] = *(aList.begin() + i);
-        }
+        memcpy(mBuffer, aList.begin(), mCount * sizeof(T));
     }
 
     inline Array(const T *aBegin, const T *aEnd) : mBuffer(NULL), mCount(0), mCapacity(0) {
         Resize(aEnd - aBegin);
-        for (s32 i = 0; i < mCount; ++i) {
-            mBuffer[i] = aBegin[i];
-        }
+        memcpy(mBuffer, aBegin, mCount * sizeof(T));
     }
 
     inline Array(const Array &aOther) : mBuffer(NULL), mCount(0), mCapacity(0) {
         Resize(aOther.mCount);
-        for (s32 i = 0; i < mCount; ++i) {
-            mBuffer[i] = aOther.mBuffer[i];
-        }
+        memcpy(mBuffer, aOther.mBuffer, mCount * sizeof(T));
     }
 
     inline void operator=(const Array &aOther) {
         Resize(aOther.mCount);
-        for (s32 i = 0; i < mCount; ++i) {
-            mBuffer[i] = aOther.mBuffer[i];
-        }
+        memcpy(mBuffer, aOther.mBuffer, mCount * sizeof(T));
     }
 
     inline ~Array() {


data/dynos_bin_pointer.cpp
```diff
@@ -13,44 +13,52 @@ typedef Pair<String, u32> PointerData;
 static PointerData GetDataFromPointer(const void* aPtr, GfxData* aGfxData) {
     // Lights
     for (auto& _Node : aGfxData->mLights) {
-        if (&_Node->mData->l[0] == aPtr) { // Light *, not Lights1 *
-            return { _Node->mName, 1 };
+        auto node = DynOS_Lights_Parse(aGfxData, _Node);
+        if (!node->mData) { continue; }
```
PeachyPeachSM64: Parse is not needed, you can just check `if (!_Node->mData) { continue; }`

data/dynos.cpp.h
```diff
@@ -209,22 +210,30 @@ class Array {
 
     inline Array(const std::initializer_list<T> &aList) : mBuffer(NULL), mCount(0), mCapacity(0) {
         Resize(aList.size());
-        memcpy(mBuffer, aList.begin(), mCount * sizeof(T));
+        for (s32 i = 0; i < mCount; ++i) {
+            mBuffer[i] = *(aList.begin() + i);
+        }
```
PeachyPeachSM64: did you make that change before or after changing some Arrays into std::vectors?
Because that defeats the point of the Array class, which is supposed to hold structs without constructors (builtin types, pointers)

data/dynos.cpp.h
```diff
@@ -597,8 +606,8 @@ struct PackData {
     bool mEnabled;
     SysPath mPath;
     String mDisplayName;
-    Array<Pair<const char *, GfxData *>> mGfxData;
-    Array<DataNode<TexData>*> mTextures;
+    std::vector<std::pair<std::string, GfxData *>> mGfxData;
```
PeachyPeachSM64, line 611: isn't that just std::map?

data/dynos_mgr_pack.cpp
```diff
@@ -1,10 +1,11 @@
+#include <deque>
 #include "dynos.cpp.h"
 extern "C" {
 #include "engine/graph_node.h"
 }
 
-static Array<PackData>& DynosPacks() {
-    static Array<PackData> sDynosPacks;
+static std::deque<PackData>& DynosPacks() {
```
PeachyPeachSM64, line 7: why a deque?

PeachyPeachSM64 (changes requested):

data/dynos_bin_pointer.cpp
Isaac0-dev: it's not needed but it's convenient because if it's some how not loaded, it will be.

data/dynos.cpp.h
Isaac0-dev: After, and yes it defeats the point of the `Array` class, but I made that change _just in case_.
We can remove that part of the PR though if I successfully replaced all the misuses of `Array`.

data/dynos.cpp.h
Isaac0-dev: no, `std::map` is different. imagining an `std::map<std::string, GfxData*>` as an example :
- std::map lets you lookup by it's key (in this case by `std::string`).
- std::map sorts elements by their key in the map.
- std::vector is as fast as a regular array when reading, but lookups for std::string would be slower than a map.

So yeah, they're definitely different.

data/dynos_mgr_pack.cpp
Isaac0-dev: Because it doesn't reallocate blocks of memory that are already in the deque (invalidating existing pointers), it just allocates new chunks when needed and ties the chunks together with pointers.
That's useful. We don't want the pointer to a pack moving in memory when the `Array` (or `std::vector`) reallocates the whole block.
Also in comparison to `std::list` which does a similar thing, `std::deque` is apparently O(1) compared to `std::list`'s O(N).

data/dynos_bin_pointer.cpp
PeachyPeachSM64: it doesn't serve any purpose actually
`GetDataFromPointer` is called during the write process by various types
When it is called for the first time by a type that requires it, all mLights, mLight0s, mLightTs, mAmbientTs are already parsed and written to the binary file, they can't be somehow not loaded

Isaac0-dev: i see what you mean, I removed it.

data/dynos.cpp.h
Isaac0-dev: I removed it.

PR #1042, opened 2025-11-29
Normalize paths in smlua_find_mod_file()

djoslin0:
Without normalizing the relative path, the comparison fails when a mod using require is hosted/joined cross-platform.

Without this fix, any mod using require is effectively not cross platform. This is critical.


src/pc/lua/smlua_require.c
```diff
@@ -68,6 +68,7 @@ void smlua_cache_module_result(lua_State* L, struct Mod* mod, struct ModFile* fi
 static struct ModFile* smlua_find_mod_file(const char* moduleName) {
     char basePath[SYS_MAX_PATH] = "";
     char absolutePath[SYS_MAX_PATH] = "";
+    char normalized_relative[SYS_MAX_PATH] = "";
```
PeachyPeachSM64: just a nitpick, call it `normalizedRelative` instead

djoslin0: done

PeachyPeachSM64 (approved):

PR #1048, opened 2025-12-06
Implement `gsDPSetTexturePersp` (to disable texture perspective correction)

Cooliokid956:
OpenGL: bumped from 2.1 to 3.0 for new keyword required
Direct3D 11: bumped from 9.x shader profile to newer profile containing a required keyword, only if available (feature will not work otherwise but there's nothing to do at that point)

Lets you do things like

https://github.com/user-attachments/assets/5e8fd7cd-1fe9-461b-a582-0cebad63f6b4

and PlayStation-style texture warping


idiotforge:
ok but like actually why was coopdx on the dx9 shader profile

PeachyPeachSM64:
conflicts

Cooliokid956:
Actually @EmeraldLoc closed it but was too lazy to do it himself 1.1

PR #1051, opened 2025-12-10
[WIP] BungeeCord64 Prototype

iZePlayzYT:
### Background

BungeeCord revolutionized Minecraft multiplayer by enabling seamless server-to-server switching. It became the backbone of every major Minecraft network (Hypixel, Mineplex, etc.), allowing players to hop between mini-games, lobbies, and worlds without disconnecting. This prototype brings that concept to SM64CoopDX.

### Features

- `/switch <port>` - Switch to another server seamlessly
- `/setserverfallback <port>` - Configure where players go if this server crashes
- `/bungeecord64` - Show connection status
- **Auto-fallback**: If a server crashes, players automatically reconnect to the fallback server

### Status
⚠️ **Prototype** - Work in progress, testing welcome!

### DIFF
--- a/autogen/lua_definitions/functions.lua
+++ b/autogen/lua_definitions/functions.lua
@@ -9509,6 +9509,61 @@ function get_os_name()
     -- ...
 end
 
+--- @param port integer
+--- @return boolean
+--- [BungeeCord64] Switches to a different server on localhost with the specified port. Returns true if switch was initiated successfully. Only works when connected as client, not as server host.
+function network_switch_to_server(port)
+    -- ...
+end
+
+--- @return integer
+--- [BungeeCord64] Gets the current server port the client is connected to or hosting on. Returns 0 if not connected.
+function network_get_current_port()
+    -- ...
+end
+
+--- @return string
+--- [BungeeCord64] Gets the current connection IP address as a string. Returns empty string if not connected.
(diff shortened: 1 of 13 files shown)


EmeraldLoc:
Putting in my 2 cents, is each server treated like a unique server with its own player count? Because if so that could be cool, as it could allow for queueing systems. For instance, if you were running instances of arena, mariohunt, etc, instead of having one server where there’s people trying to squeeze into a single server, you could have it automatically distribute the number of players, allowing for everyone to have a good experience. It could also make it significantly easier to have multiple gamemode mods switched between if that’s another direction you want to take it. 

> Mods can handle having separate mini games without needing to connect to a whole other server and arguably that would just be a far worse solution that doesn't retain the state of the mod as everything is reloaded

Unless there’s already functionality I’m not aware of, if I wanted to switch from arena to drench games, you would need to first of all have the client download both of the mods even if only one is being used, and modify each mod to make sure they don’t run if they aren’t on. If you don’t want to do that, you can have a script that changes the mod time based, but you couldn’t have a voting system so players could pick. So while it is possible, if I’m understanding this pr correctly, it could make it a good bit easier.

When I say all this I’m assuming the pr allows switching to servers with a different mod list, with their own unique player count.

PR #1053, opened 2025-12-14
Allow action hook to return -1 to continue execution

PeachyPeachSM64:
### Problem:

Vanilla actions can be overridden with `hook_mario_action`, but one needs to implement partially or fully the action again to retain the vanilla behavior.

### Solution:

Allow action hooks to return -1 to tell the game to continue the execution and call the vanilla action code.

Also works with gravity hooks.

### DIFF
--- a/src/pc/lua/smlua_hooks.c
+++ b/src/pc/lua/smlua_hooks.c
@@ -343,12 +343,13 @@ bool smlua_call_action_hook(enum LuaActionHookType hookType, struct MarioState* 
 
             // call the callback
             if (0 != smlua_call_hook(L, 1, 1, 0, hook->mod, hook->modFile)) {
-                LOG_LUA("Failed to call the action callback: %u", m->action);
+                LOG_LUA("Failed to call the action callback: '%08X'", m->action);
                 continue;
             }
 
             // output the return value
-            // returning a negative value allows to continue the execution, useful when overriding vanilla actions
+            // special return values:
+            // - returning -1 allows to continue the execution, useful when overriding vanilla actions
             bool stopActionHook = true;
             *cancel = FALSE;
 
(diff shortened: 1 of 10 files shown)


src/pc/lua/smlua_hooks.c
```diff
@@ -348,13 +348,31 @@ bool smlua_call_action_hook(enum LuaActionHookType hookType, struct MarioState*
...
+                } break;
+
+                case LUA_TNUMBER: {
+                    s32 returnValue = (s32) smlua_to_integer(L, -1);
+                    if (returnValue > 0) {
+                        *cancel = TRUE;
+                    } else if (returnValue == 0) {
+                        *cancel = FALSE;
+                    } else {
+                        stopActionHook = false;
```
Isaac0-dev, line 368: I think you should require it to be exactly -1, rather than any negative value.
It forces mods to be consistent, so that we can add more options later if we need.

src/pc/lua/smlua_hooks.c
```diff
@@ -348,13 +348,31 @@ bool smlua_call_action_hook(enum LuaActionHookType hookType, struct MarioState*
...
-            *returnValue = false;
-            if (lua_type(L, -1) == LUA_TBOOLEAN || lua_type(L, -1) == LUA_TNUMBER) {
-                *returnValue = smlua_to_integer(L, -1);
+            // returning a negative value allows to continue the execution, useful when overriding vanilla actions
+            bool stopActionHook = true;
+            *cancel = FALSE;
+
+            switch (lua_type(L, -1)) {
+                case LUA_TBOOLEAN: {
+                    *cancel = smlua_to_boolean(L, -1) ? TRUE : FALSE;
```
Isaac0-dev, line 358: is this for clarity? because C will convert a boolean to an integer automatically, so you don't need the ` ? TRUE : FALSE`

Isaac0-dev (changes requested):

src/pc/lua/smlua_hooks.c
PeachyPeachSM64: yes, it is redundant, but it's for clarity

src/pc/lua/smlua_hooks.c
PeachyPeachSM64: good point

PeachyPeachSM64: should we raise an error message if the return value is out of range (i.e negative but not -1)? or just do nothing and return silently?

Isaac0-dev: tbh I'm kinda sick of debugging coop's silent errors so I vote that you put an error message.

Isaac0-dev (approved):

PR #1057, opened 2025-12-16
Update djui_chat_message.h with MAX_CHAT_PACKET_LENGTH

occamsaturn:
This, in combination with a change to djui_chat_message.c, will also assist with the chat character limit increase by allowing for larger strings to be sent over chat packets without loss.

This could probably be refined down to exactly the amount of extra data chat packets can contain, but 600 works, so that's what I went with.

### DIFF
--- a/src/pc/djui/djui_chat_message.h
+++ b/src/pc/djui/djui_chat_message.h
@@ -2,6 +2,7 @@
 #include "djui.h"
 
 #define MAX_CHAT_MSG_LENGTH 500
+#definte MAX_CHAT_PACKET_LENGTH 600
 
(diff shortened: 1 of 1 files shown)


src/pc/djui/djui_chat_message.h
```diff
@@ -2,6 +2,7 @@
 #include "djui.h"
 
 #define MAX_CHAT_MSG_LENGTH 500
+#definte MAX_CHAT_PACKET_LENGTH 600
```
AgentXLP, line 5: Your `#define` has a rogue T in it

PR #1058, opened 2025-12-16
Update djui_chat_message.c to replace two instances of MAX_CHAT_MSG_LENGTH with MAX_CHAT_PACKET_LENGTH

occamsaturn:
If we replace MAX_CHAT_MSG_LENGTH in these two spots, it fixes the previous issue of having the chat text entry spill over the character limit (resulting in characters at the end of messages that are lost) while raising the character limit to the intended limit of 500. Or 499 more specifically.


AgentXLP:
For future references base PRs on the dev branch

PR #1059, opened 2025-12-16
Update packet_chat.c, djui_chat_message.h, djui_chat_message.c to fix chat character limit

occamsaturn:
Increase `remoteMessage` to 500 and `remoteMessageLength` to 499, respectively, to accommodate the previous (and tragically incomplete) chat character limit increase.

Then define `MAX_CHAT_PACKET_LENGTH` and replace `MAX_CHAT_MSG_LENGTH` in a couple of areas to allow for larger strings to be sent over chat packets without loss.

`MAX_CHAT_PACKET_LENGTH` could probably be refined down to exactly the amount of extra data chat packets can contain, but 600 works, so that's what I went with.

All of this together makes for a character limit of 499, up from 255, without the text entry field letting you type more than that into the chat (and having those extra characters be lost).

### DIFF
--- a/src/pc/network/packets/packet_chat.c
+++ b/src/pc/network/packets/packet_chat.c
@@ -78,12 +78,12 @@ void network_send_chat(char* message, u8 globalIndex) {
 
 void network_receive_chat(struct Packet* p) {
     u16 remoteMessageLength = 0;
-    char remoteMessage[500] = { 0 };
+    char remoteMessage[MAX_CHAT_MSG_LENGTH] = { 0 };
     u8 globalIndex;
 
     packet_read(p, &globalIndex, sizeof(u8));
     packet_read(p, &remoteMessageLength, sizeof(u16));
-    if (remoteMessageLength >= 499) { remoteMessageLength = 499; }
+    if (remoteMessageLength >= MAX_CHAT_MSG_LENGTH - 1) { remoteMessageLength = MAX_CHAT_MSG_LENGTH - 1; }
     packet_read(p, &remoteMessage, remoteMessageLength * sizeof(u8));
 
     // anti spoof


src/pc/djui/djui_chat_message.h
```diff
@@ -2,6 +2,7 @@
 #include "djui.h"
 
 #define MAX_CHAT_MSG_LENGTH 500
+#define MAX_CHAT_PACKET_LENGTH 600
```
PeachyPeachSM64: please use a precise value instead of an approximation
chat msg consists of:
- `\#RRGGBB\` (9 chars)
- `np->name` (max (MAX_CONFIG_STRING - 1) chars)
- `\#dcdcdc\` (9 chars)
- `: ` (2 chars)
- `message` (max (MAX_CHAT_MSG_LENGTH - 1) chars)
- `\0` (null term)

so the value of MAX_CHAT_PACKET_LENGTH would be:
`(MAX_CONFIG_STRING + MAX_CHAT_PACKET_LENGTH + 19)`

src/pc/network/packets/packet_chat.c
```diff
@@ -78,12 +78,12 @@ void network_send_chat(char* message, u8 globalIndex) {
...
 void network_receive_chat(struct Packet* p) {
     u16 remoteMessageLength = 0;
-    char remoteMessage[256] = { 0 };
+    char remoteMessage[500] = { 0 };
     u8 globalIndex;
 
     packet_read(p, &globalIndex, sizeof(u8));
     packet_read(p, &remoteMessageLength, sizeof(u16));
-    if (remoteMessageLength >= 255) { remoteMessageLength = 255; }
+    if (remoteMessageLength >= 499) { remoteMessageLength = 499; }
```
PeachyPeachSM64: please use named values instead of hardcoded
500 -> MAX_CHAT_MSG_LENGTH
499 -> MAX_CHAT_MSG_LENGTH  - 1

PR #1064, opened 2025-12-26
Add various interpolation skipping functions + `get_time_stop_flags()`

Cooliokid956:
## `void geo_skip_interpolation(struct GraphNode *node, struct GraphNodeObject *obj)`
Skips graph node interpolation for a frame
feed it a graph node and its corresponding object

## `void obj_skip_interpolation(struct Object *o)`
Skips object interpolation for a frame
takes care of only the object transform, otherwise you need to use

## `void obj_anim_skip_interpolation(struct Object *o)`
Skips animation interpolation for a frame
useful for choppy animation

------------
## `u32 get_time_stop_flags(void)`
Gets the current time stop state

### DIFF
--- a/src/pc/lua/utils/smlua_misc_utils.c
+++ b/src/pc/lua/utils/smlua_misc_utils.c
@@ -659,7 +659,7 @@ struct GraphNodeHeldObject* geo_get_current_held_object(void) {
 
 void geo_skip_interpolation(struct GraphNode *node, struct GraphNodeObject *obj) {
     struct GraphNodeInterpData *interp = geo_get_interp_data(node, obj);
-    interp->timestamp = 0;
+    if (interp) { interp->timestamp = 0; }
 }
 
 ///
--- a/src/pc/lua/utils/smlua_obj_utils.c
+++ b/src/pc/lua/utils/smlua_obj_utils.c
@@ -523,11 +523,11 @@ void set_whirlpools(f32 x, f32 y, f32 z, s16 strength, s16 area, s32 index) {
 }
 
 void obj_skip_interpolation(struct Object *o) {
-    o->header.gfx.skipInterpolationTimestamp = gGlobalTimer + 1;
+    if (o) { o->header.gfx.skipInterpolationTimestamp = gGlobalTimer + 1; }
 }
 
 void obj_anim_skip_interpolation(struct Object *o) {
-    o->header.gfx.animInfo.prevAnimFrameTimestamp = 0;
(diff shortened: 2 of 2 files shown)


src/pc/lua/utils/smlua_misc_utils.c
```diff
@@ -651,6 +657,13 @@ struct GraphNodeHeldObject* geo_get_current_held_object(void) {
     return gCurGraphNodeHeldObject;
 }
 
+void geo_skip_interpolation(struct GraphNode *node, struct GraphNodeObject *obj) {
+    struct GraphNodeInterpData *interp = geo_get_interp_data(node, obj);
```
Isaac0-dev, line 661: what if this returns null?

src/pc/lua/utils/smlua_obj_utils.c
```diff
@@ -522,6 +522,14 @@ void set_whirlpools(f32 x, f32 y, f32 z, s16 strength, s16 area, s32 index) {
     gAreas[area].whirlpools[index]->strength = strength;
 }
 
+void obj_skip_interpolation(struct Object *o) {
+    o->header.gfx.skipInterpolationTimestamp = gGlobalTimer + 1;
```
Isaac0-dev: check if `o` is null.

src/pc/lua/utils/smlua_obj_utils.c
```diff
@@ -522,6 +522,14 @@ void set_whirlpools(f32 x, f32 y, f32 z, s16 strength, s16 area, s32 index) {
...
+}
+
+void obj_anim_skip_interpolation(struct Object *o) {
+    o->header.gfx.animInfo.prevAnimFrameTimestamp = 0;
```
Isaac0-dev: same here

src/pc/lua/utils/smlua_misc_utils.c
Cooliokid956: we don't access it

src/pc/lua/utils/smlua_obj_utils.c
Cooliokid956: can I check if it's not null

src/pc/lua/utils/smlua_obj_utils.c
Cooliokid956: me too

Isaac0-dev:
👍

Isaac0-dev:
👍

PR #1069, opened 2026-01-03
Add version info (Windows) + Smart update checker

Cooliokid956:
# Add version info (Windows)
Added version info to the resource file to display the information as shown below:

image

# Smart update checker
Version strings are now parsed and compared properly- development versions will no longer see false positives if their version is later than the latest:

#### Nothing on v41.1, as expected
image

#### Nothing wrong with a hypothetical later version!
image

#### I should probably update
image


src/pc/network/version.c
```diff
@@ -4,22 +4,34 @@
...
-    snprintf(sVersionString, MAX_VERSION_LENGTH, "%s %s", SM64COOPDX_VERSION, VERSION_REGION);
+        "%s %s", SM64COOPDX_VERSION, VERSION_REGION
 #endif // VERSION_US
+    );
     return sVersionString;
 }
 
+const char* get_version(void) {
+    get_version_online();
+    extern void exify_version_str(char* str);
```
djoslin0: should use header file declarations when possible instead of extern
I know the codebase is guilty of this elsewhere, but it should be reduced

src/pc/update_checker.c
```diff
@@ -51,21 +64,57 @@ size_t write_callback(char *ptr, size_t size, size_t nmemb, void *userdata) {
...
 #endif
 
+void strtov(const char *str, struct Version *ver) {
+    char* end;
+    ver->maj = strtol(str+1, &end, 10);
+    if (end) ver->min = strtol(end+1, &end, 10);
+    if (end) ver->fix = strtol(end+1, &end, 10);
+}
+
+void vtostr(struct Version ver, char* str) {
```
djoslin0: vtostr is an unclear name
Should accept a size parameter instead of assuming 8 bytes `void vtostr(struct Version ver, char* str, size_t size)`

src/pc/update_checker.c
```diff
@@ -51,21 +64,57 @@ size_t write_callback(char *ptr, size_t size, size_t nmemb, void *userdata) {
 }
 #endif
 
+void strtov(const char *str, struct Version *ver) {
+    char* end;
```
djoslin0, line 68: If parsing fails at any step, the remaining fields are not initialized. Should initialize the struct first:
```C

ver->maj = 0;
ver->min = 0;
ver->fix = 0;

```

src/pc/djui/djui_panel_main.c
```diff
@@ -24,8 +29,56 @@ static void djui_panel_main_quit(struct DjuiBase* caller) {
...
+static void djui_panel_main_egg_end(struct DjuiBase *base, UNUSED bool *noRender) {
+    panel->base.x.value -= gGlobalTimer - sEggLastFired + 60;
+    base->color.a = MAX(255.f - ((gGlobalTimer - sEggLastFired + 60) << 3), 0);
+    if (gGlobalTimer == sEggLastFired) { game_exit(); }
+}
+
+static void djui_panel_main_egg(struct DjuiBase *base, UNUSED bool *noRender) {
+    if (sEggCounter && sEggLastY == base->clip.y) {
+        bool hovering = djui_cursor_inside_base(base);
+        bool clicking = (gInteractablePad.button & PAD_BUTTON_A || mouse_buttons & L_MOUSE_BUTTON);
```
djoslin0: The bitwise `&` has lower precedence than `||`
always wrap `(var & bitmask)` in parenthesis

src/pc/djui/djui_panel_main.c
```diff
@@ -6,12 +6,17 @@
...
 #include "pc/pc_main.h"
 #include "pc/update_checker.h"
+#include "sounds.h"
+#include "audio/external.h"
+#include "game/area.h"
 
 extern ALIGNED8 u8 texture_coopdx_logo[];
 
 bool gDjuiPanelMainCreated = false;
+struct DjuiThreePanel* panel = NULL;
```
djoslin0: there should never be a global called `panel`
`gPanelMain` if global or `sPanelMain` if static

src/pc/update_checker.c
```diff
@@ -51,21 +64,57 @@ size_t write_callback(char *ptr, size_t size, size_t nmemb, void *userdata) {
 }
 #endif
 
+void strtov(const char *str, struct Version *ver) {
```
djoslin0: unclear name

src/pc/network/version.h
```diff
@@ -8,6 +8,8 @@
 #define VERSION_NUMBER 41
 #define MINOR_VERSION_NUMBER 1
 
+#define VERSION_OFFSET 37
```
djoslin0: should leave a comment explaining why this offset exists

djoslin0 (changes requested):
I don't really like such an elaborate _egg_ personally, but that part isn't up to me. My comments focus on the code itself

src/pc/update_checker.c
Cooliokid956: my thinking at the time said it's just as clear as string to long (strtol), maybe even stronger given the prevalent subject and context of the file it's in
but if it's really that bad, do you suggest expanding the v to ver?

djoslin0: I'd prefer `string_to_version` personally. The old C naming conventions are unnecessarily terse in modern times and aren't a good example to follow imo

Isaac0-dev:
Please let's not keep those theme changes. It's too much for a theme.

src/pc/network/version.c
Cooliokid956: the related code was nuked C:

src/pc/update_checker.c
Cooliokid956: I hope this is acceptable
```c
void string_to_version(const char *str, struct Version *ver) {
    char* end;
    ver->maj = strtol(str+1, &end, 10);
    ver->min = 0; ver->fix = 0;
    if (end) ver->min = strtol(end+1, &end, 10);
    if (end) ver->fix = strtol(end+1, &end, 10);
}
```

src/pc/djui/djui_panel_main.c
Cooliokid956: Done

Cooliokid956: and nuked

src/pc/djui/djui_panel_main.c
Cooliokid956: also nuked

src/pc/update_checker.c
Cooliokid956: done

src/pc/network/version.h
Cooliokid956:
```c
#define VERSION_OFFSET 37 // difference from old versioning system
```

PR #1071, opened 2026-01-06
Shift S and T support for Texture Tiles

ManIsCat2:
Adds Shift S and Shift T support for the PC Renderer, now people can use this in their models which they couldn't do before

### DIFF
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -1166,6 +1166,24 @@ static void OPTIMIZE_O3 gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t 
         if (use_texture) {
             float u = (v_arr[i]->u - rdp.texture_tile.uls * 8) / 32.0f;
             float v = (v_arr[i]->v - rdp.texture_tile.ult * 8) / 32.0f;
+
+            int shifts = rdp.texture_tile.shifts;
+            int shiftt = rdp.texture_tile.shiftt;
+            if (shifts != 0) {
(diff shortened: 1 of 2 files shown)


src/pc/gfx/gfx_pc.c
```diff
@@ -1166,6 +1166,24 @@ static void OPTIMIZE_O3 gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t
...
+                    u *= 1 << (16 - shifts);
+                }
+            }
+            if (shiftt != 0) {
+                if (shiftt <= 10) {
+                    v /= 1 << shiftt;
+                } else {
+                    v *= 1 << (16 - shiftt);
+                }
+            }
```
PeachyPeachSM64, line 1185: I'm not sure if it's the right reference or not, but it seems that the formula for shift > 10 is multiply by 2^(shift - 10): https://hack64.net/wiki/doku.php?id=super_mario_64:fast3d_display_list_commands#f5g_settile

src/pc/gfx/gfx_pc.c
```diff
@@ -1404,6 +1422,8 @@ static void gfx_dp_set_tile(uint8_t fmt, uint32_t siz, uint32_t line, uint32_t t
         rdp.texture_tile.siz = siz;
         rdp.texture_tile.cms = cms;
         rdp.texture_tile.cmt = cmt;
+        rdp.texture_tile.shifts = shifts;
+        rdp.texture_tile.shiftt = shiftt;
```
PeachyPeachSM64, line 1426: removed UNUSED from function signature

PR #1077, opened 2026-01-11
Improved deadzone implementation

abcpea:
Reworked analog stick inputs so that deadzones normalize the axes instead of simply masking them.
Makes sneaking possible even with high deadzone values.


src/pc/controller/controller_sdl2.c
```diff
@@ -176,6 +176,30 @@ static inline void update_button(const int i, const bool new) {
     }
 }
 
+static inline void update_analog_stick(s8 *stick_x, s8 *stick_y,
+                                        int16_t input_x, int16_t input_y) {
+    static const float INPUT_ABS_MAX = 0x8000;
```
EmeraldLoc: This should either just be used directly since it's only used once or defined in a macro above. My recommendation would be going to line 36 and adding your macro there

```c
#define INPUT_ABS_MAX 0x8000
```

EmeraldLoc (changes requested):
Tested and this works as advertised, just needs a minor changes

src/pc/controller/controller_sdl2.c
abcpea: I don't feel that it makes sense as a macro so I have just changed it to the value.

EmeraldLoc (approved):

PR #1081, opened 2026-01-12
Extra Characters Plus - Update

Funky90sLion:
Changelog:

Coins will now add seconds to the oxygen timer. Yellow coins add 5, red coins add 10, and blue coins 25. Rings do not add any oxygen.
Prevented the drowning code from running if Sonic is in a cutscene action, like the star dance action.
Stars reset oxygen timer.
Fixed an oversight in the homing attack code for breaking boxes.
Fixed doors not opening when you're just under a bit of water.
Fixed Sonic's ground pound not properly killing some enemies.
Fix script error in homing attack action
Fix drowning code that didn't care about cutscene actions
Fix scaling issue on homing attack cursor and make it move from Sonic instead of a random pos
Make water rings and bubbles fully restore oxygen underwater


EmeraldLoc:
Arena changes should probably be in a separate pr, although heads up I don't really like the arena player color palette changes.

PR #1094, opened 2026-01-28
Merry-Go-Round Fixes (includes fixes for rooms and remote object-spawned objects)

yoshiweegee:
In the original SM64 code, `gMarioOnMerryGoRound` is tied with the music checks. This involves checking if Mario is on the merry-go-round platform, or if the floor type is `SURFACE_MGR_MUSIC`. Since music is client-side, SM64CDX needs the check for a Mario on the merry-go-round to be separated.

It currently does this by checking if any player character is on the platform, but not if the floor type is `SURFACE_MGR_MUSIC`. This has a side effect of the merry-go-round Boos not activating until a player character is on the merry-go-round platform. Critically, this means **the entrances to the merry-go-round do not activate the merry-go-round Boos**.

This can be observed in v1.4.1 if the player stands their character in the entrance of the merry-go-round closest to the painting where the Boos exit, and looks at said painting while close enough, <ins>but not on the platform</ins>. The Boos will spawn, but will not move, since they don't activate.

The fix is simple: also check if any player character's floor type is `SURFACE_MGR_MUSIC`. I considered removing the object platform check, since it's redundant in SM64, but I chose to keep it for consistency with expected behavior for mods.

---

It's very common for the merry-go-round Boos to desync if more than one player is in the same act of BBH. Sometimes it's possible to hit the Boos in the middle of the action transition from exiting the painting to roaming, but sometimes they stop spawning at all.

<ins>From minimal testing</ins>, it seems this is caused by a check on a Boo's action to determine if it should continue to be synced. It checks if `oAction` is less than 3. This seems to make sense at first glance, because action 3 is for dying, and action 4 is for displaying the text box for a "Go on a Ghost Hunt" Boo. However, there's also action 5, which is for a merry-go-round Boo exiting the painting. So this check is **causing the sync object to be forgotten specifically for merry-go-round Boos as they exit the painting**.

Adjusting the check to make sure action 5 is also included for syncing ~~seems to fix the issue.~~ However, I have my doubts, as since Boos sometimes don't spawn at all, and the spawning action is action 0, there's a chance that there's a different issue causing that. For that reason, in addition to the volatile nature of desyncs, **testing with various server circumstances would be prudent to see if this fix is sufficient**.

---

In addition to the first section, various parts of the code were preventing the game from properly detecting if a remote player was on the merry-go-round. Included in this was the rotating platform, since **an object's dynamic collision isn't loaded unless the rooms match**. The code has been modified to account for this. **This means other objects might be affected**.

---

Merry-go-round Boos aren't supposed to chase Mario if he goes too far away from their home. For some reason, Coop DX made it so merry-go-round Boos can **chase a player character regardless of distance**. This check has been added back.

---

Objects that are spawned after level load by a remote player are flagged as having been sent over the network. This affects the `SET_HOME` behavior command's function, `bhv_cmd_set_home`, where it **doesn't set the home if the object was sent over the network**. This is a problem for merry-go-round Boos since they use their behavior to set their home on spawn. To account for this, objects sent over the network can now set their home via behavior if the home hasn't already been set by behavior (e.g., if `Object.setHome` is `FALSE`) and the home is still the default initialization of (`0.0f`, `0.0f`, `0.0f`).  **This change potentially affects many objects, and obligates testing in an uncontrolled environment to prevent nasty oversights from affecting nonexperimental players.**

### DIFF
--- a/src/engine/behavior_script.c
+++ b/src/engine/behavior_script.c
@@ -797,7 +797,22 @@ static s32 bhv_cmd_load_collision_data(void) {
 // Command 0x2D: Sets the home position of the object to its current position.
 // Usage: SET_HOME()
 static s32 bhv_cmd_set_home(void) {
-    if (!(gCurrentObject->coopFlags & (COOP_OBJ_FLAG_LUA | COOP_OBJ_FLAG_NETWORK))) {
+    // COOP: only set home via behavior for the following cases
+    if (
+        // if the object wasn't created via Lua
+        !(gCurrentObject->coopFlags & COOP_OBJ_FLAG_LUA)
+        // if the object wasn't created via network
+        // OR
+        // the object has never had its home set via behavior AND its home is default (e.g. (0, 0, 0))
+        // (this case handles an object that needs its home set via behavior after being spawned by another player)
+        && (
+            !(gCurrentObject->coopFlags & COOP_OBJ_FLAG_NETWORK)
+            || (
+                !gCurrentObject->setHome
+                && gCurrentObject->oHomeX == 0.0f && gCurrentObject->oHomeY == 0.0f && gCurrentObject->oHomeZ == 0.0f
+            )
+        )
+    ) {
         gCurrentObject->oHomeX = gCurrentObject->oPosX;
         gCurrentObject->oHomeY = gCurrentObject->oPosY;
         gCurrentObject->oHomeZ = gCurrentObject->oPosZ;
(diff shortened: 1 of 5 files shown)


sher-b:
doing what no one else could

sher-b:
there's this very consistent issue that happens when [at least one person is in the mgr act and not in the basement](https://youtu.be/BeMyozPaI4E)

Isaac0-dev (approved):

PR #1095, opened 2026-01-28
Fix Texture tiles & Add shift support

ManIsCat2:
Fixes RDP tiles in gfx_pc, now each texture tile has its own UV when rendering it. Also adds Shift S and T support to the tiles.

Currently only OpenGL works with this PR, other renderers will be supported soon:tm:


AgentXLP:
I think this is probably good to merge for now

EmeraldLoc (approved):
To the best of my ability, this looks fine

PR #1097, opened 2026-01-29
Add methods to common structs

Sunketchupm:
I thought of this a while back and figured it shouldn't be too hard to do. It wasn't, thanks to PeachyPeach's efforts for making the ModFS methods work.
Functions that have their first parameter with a struct will be added as a method onto that struct. For example, `set_mario_action` will be added onto the `MarioState` struct and, after a bit of name trimming, can be called by instead doing `m:set_action`.

Most likely there exists some way to make this work well without being this manual, but it's the best I can do with my knowledge.

Test mod:
```lua
---@param m MarioState

    local obj = obj_get_nearest_object_with_behavior_id(m.marioObj, id_bhvGoomba)
    if obj then
        if obj:check_hitbox_overlap(m.marioObj) then
            m:set_action(ACT_JUMP_KICK, 0)
        end
    end
end

```
And it works exactly as would with normal function calls


EmeraldLoc:
conflicts

PR #1098, opened 2026-01-31
Lighting Engine refactor (up to 10x perf increase)

AgentXLP:
Ported the lighting engine code to C++ and I'm using vectors instead of iterating through a large array for every single vertex to check for lights that potentially aren't even enabled. Big performance increase while still working basically the same.

### DIFF
--- a/src/engine/lighting_engine.cpp
+++ b/src/engine/lighting_engine.cpp
@@ -1,6 +1,7 @@
 #include "lighting_engine.h"
 extern "C" {
 #include "math_util.h"
+#include "pc/lua/smlua.h"
 }
 #undef clamp
 #undef min
@@ -13,17 +14,11 @@ extern "C" {
 
 #define C_FIELD extern "C"
 
-#define LE_MAX_LIGHTS_PER_VERTEX 4
-
 struct LELight
 {
     s16 id;
-    f32 posX;
-    f32 posY;
-    f32 posZ;
-    u8 colorR;
-    u8 colorG;
-    u8 colorB;
+    Vec3f pos;
+    Color color;
     f32 radius;
     f32 intensity;
     bool useSurfaceNormals;
@@ -36,6 +31,7 @@ static s16 sLightID = -1;
 static enum LEMode sMode = LE_MODE_AFFECT_ALL_SHADED_AND_COLORED;
 static enum LEToneMapping sToneMapping = LE_TONE_MAPPING_WEIGHTED;
 static bool sEnabled = false;
+static u8 sMaxLightsPerVertex = 4;
 
 static inline f32 rsqrt(f32 value) {
 #ifdef __SSE__
@@ -57,9 +53,8 @@ static inline void color_copy(Color dest, Color src) {
     dest[2] = src[2];
 }
 
-
-static inline u8 to_u8(float value) {
-    int v = (int)value;
+static inline u8 clamp_u8(f32 value) {
+    s32 v = (s32)value;
     v = v < 0 ? 0 : v;
     v = v > 255 ? 255 : v;
     return (u8)v;
@@ -93,22 +88,26 @@ C_FIELD void le_set_ambient_color(u8 r, u8 g, u8 b) {
     sEnabled = true;
 }
 
-static inline void le_tone_map_total_weighted(Color out, Color inAmbient, Vec3f inColor, float weight) {
-    out[0] = to_u8((inAmbient[0] + inColor[0]) / weight);
-    out[1] = to_u8((inAmbient[1] + inColor[1]) / weight);
-    out[2] = to_u8((inAmbient[2] + inColor[2]) / weight);
+C_FIELD void le_set_max_lights_per_vertex(u8 count) {
+    sMaxLightsPerVertex = count;
 }
 
-static inline void le_tone_map_weighted(Color out, Color inAmbient, Vec3f inColor, float weight) {
-    out[0] = to_u8(inAmbient[0] + (inColor[0] / weight));
-    out[1] = to_u8(inAmbient[1] + (inColor[1] / weight));
-    out[2] = to_u8(inAmbient[2] + (inColor[2] / weight));
+static inline void le_tone_map_total_weighted(Color out, Color inAmbient, Vec3f inColor, f32 weight) {
+    out[0] = clamp_u8((inAmbient[0] + inColor[0]) / weight);
+    out[1] = clamp_u8((inAmbient[1] + inColor[1]) / weight);
+    out[2] = clamp_u8((inAmbient[2] + inColor[2]) / weight);
+}
+
+static inline void le_tone_map_weighted(Color out, Color inAmbient, Vec3f inColor, f32 weight) {
+    out[0] = clamp_u8(inAmbient[0] + (inColor[0] / weight));
+    out[1] = clamp_u8(inAmbient[1] + (inColor[1] / weight));
+    out[2] = clamp_u8(inAmbient[2] + (inColor[2] / weight));
 }
 
 static inline void le_tone_map_clamp(Color out, Color inAmbient, Vec3f inColor) {
-    out[0] = to_u8(inAmbient[0] + inColor[0]);
-    out[1] = to_u8(inAmbient[1] + inColor[1]);
-    out[2] = to_u8(inAmbient[2] + inColor[2]);
+    out[0] = clamp_u8(inAmbient[0] + inColor[0]);
+    out[1] = clamp_u8(inAmbient[1] + inColor[1]);
+    out[2] = clamp_u8(inAmbient[2] + inColor[2]);
 }
 
 static inline void le_tone_map_reinhard(Color out, Color inAmbient, Vec3f inColor) {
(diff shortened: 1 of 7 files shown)


src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+#undef max
+#include <vector>
+#include <algorithm>
+#ifdef __SSE__
+#include <xmmintrin.h>
+#endif
+
+#define C_FIELD extern "C"
+
+#define LE_MAX_LIGHTS_PER_VERTEX 4
```
PeachyPeachSM64: I feel like this should be a setting rather than a hardcoded limit
that would allow modders to control the "accuracy" of their lighting

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+struct LELight
+{
+    s16 id;
+    f32 posX;
+    f32 posY;
+    f32 posZ;
+    u8 colorR;
+    u8 colorG;
+    u8 colorB;
```
PeachyPeachSM64: any reason why they're not Vec3f and Color?

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+}
+
+static inline void color_copy(Color dest, Color src) {
+    dest[0] = src[0];
+    dest[1] = src[1];
+    dest[2] = src[2];
+}
+
+
+static inline u8 to_u8(float value) {
```
PeachyPeachSM64: extra newline
change `float` to `f32`
I think `clamp_u8` would be a better name, it's more than a simple cast

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    sEnabled = true;
+}
+
+static inline void le_tone_map_total_weighted(Color out, Color inAmbient, Vec3f inColor, float weight) {
+    out[0] = to_u8((inAmbient[0] + inColor[0]) / weight);
+    out[1] = to_u8((inAmbient[1] + inColor[1]) / weight);
+    out[2] = to_u8((inAmbient[2] + inColor[2]) / weight);
+}
+
+static inline void le_tone_map_weighted(Color out, Color inAmbient, Vec3f inColor, float weight) {
```
PeachyPeachSM64: change `float` to `f32`

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    inColor[0] += inAmbient[0];
+    inColor[1] += inAmbient[1];
+    inColor[2] += inAmbient[2];
+
+    out[0] = to_u8((inColor[0] / (inColor[0] + 255.0f)) * 255.0f);
+    out[1] = to_u8((inColor[1] / (inColor[1] + 255.0f)) * 255.0f);
+    out[2] = to_u8((inColor[2] / (inColor[2] + 255.0f)) * 255.0f);
+}
+
+static inline void le_tone_map(Color out, Color inAmbient, Vec3f inColor, float weight) {
```
PeachyPeachSM64: `float` to `f32`
and don't inline this function

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+    // accumulate
+    out_color[0] += light.colorR * brightness;
+    out_color[1] += light.colorG * brightness;
+    out_color[2] += light.colorB * brightness;
+    weight += brightness;
+    contribution++;
+}
+
+C_FIELD OPTIMIZE_O3 void le_calculate_vertex_lighting(Vtx_t* v, Vec3f pos, Color out) {
```
PeachyPeachSM64: since you're not editing it, `v` should be a `const Vtx_t *`
that way, you could remove the cast in gfx_pc.c, line 932: `le_calculate_vertex_lighting((Vtx_t*)v, vpos, color);`

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+static inline void color_copy(Color dest, Color src) {
+    dest[0] = src[0];
+    dest[1] = src[1];
+    dest[2] = src[2];
+}
+
+
+static inline u8 to_u8(float value) {
+    int v = (int)value;
```
PeachyPeachSM64: `int` to `s32`

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    for (LELight* light : sActiveLights) {
+        le_calculate_light_contribution(*light, pos, normal, lightIntensityScalar, color, weight, contribution);
+        if (contribution == LE_MAX_LIGHTS_PER_VERTEX) { break; }
+    }
+
+    // tone map and output
+    le_tone_map(out, gLEAmbientColor, color, weight);
+}
+
+C_FIELD void le_calculate_lighting_dir(Vec3f pos, Vec3f out) {
```
PeachyPeachSM64: these functions miss the VEC_OUT hint in their definition:
le_calculate_vertex_lighting
le_calculate_lighting_color
le_calculate_lighting_color_with_normal
le_calculate_lighting_dir

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    out[0] = lightingDir[0] / (f32)(count);
+    out[1] = lightingDir[1] / (f32)(count);
+    out[2] = lightingDir[2] / (f32)(count);
+    vec3f_normalize(out);
+}
+
+C_FIELD s16 le_add_light(f32 x, f32 y, f32 z, u8 r, u8 g, u8 b, f32 radius, f32 intensity) {
+    if (sLightPool.size() >= LE_MAX_LIGHTS) { return -1; }
```
PeachyPeachSM64: I think you should add an error message indicating that you reached the max number of lights, like how the hooks and behaviors do

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    if (id < 0) { return; }
+
+    sLightPool.erase(
+        std::remove_if(sLightPool.begin(), sLightPool.end(),
+            [id](const LELight& light) {
+                return light.id == id;
+            }
+        ),
+        sLightPool.end()
+    );
```
PeachyPeachSM64: I don't understand the logic here
First, you're removing the light that matches the provided id, then you remove all lights from whatever is returned by `std::remove_if` until the end?

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    if (id < 0) { return; }
+
+    auto it = std::find_if(sLightPool.begin(), sLightPool.end(),
+        [id](const LELight& light) {
+            return light.id == id;
+        }
+    );
+
+    if (it != sLightPool.end()) {
+        LELight& light = *it;
```
PeachyPeachSM64: ngl, that's a lot of duplicated code
it would be a good idea to move it to a `le_get_light_by_id` function

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+    }
+}
+
+void le_clear(void) {
+    sLightPool.clear();
+    sLightID = 0;
+
+    gLEAmbientColor[0] = 127;
+    gLEAmbientColor[1] = 127;
+    gLEAmbientColor[2] = 127;
```
PeachyPeachSM64: why isn't it using `color_set`?

src/engine/lighting_engine.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+    if (it != sLightPool.end()) {
+        LELight& light = *it;
+        light.useSurfaceNormals = useSurfaceNormals;
+    }
+}
+
+void le_clear(void) {
+    sLightPool.clear();
+    sLightID = 0;
```
PeachyPeachSM64: shouldn't `sLightID` be reset to -1 instead of 0?

PeachyPeachSM64 (changes requested):

src/engine/lighting_engine.cpp
AgentXLP: Good catch, I see now that's why my lights are acting weird sometimes and having the wrong IDs.

src/engine/lighting_engine.cpp
AgentXLP: Will fix

src/engine/lighting_engine.cpp
AgentXLP: I remember hearing its slightly faster for the computer to access individual variables instead of elements of an array but it's probably so negligible I could change it.

src/engine/lighting_engine.cpp
AgentXLP: Alright

src/engine/lighting_engine.cpp
AgentXLP: Alright

src/engine/lighting_engine.cpp
AgentXLP: Alright

src/engine/lighting_engine.cpp
AgentXLP: Alright

src/engine/lighting_engine.cpp
AgentXLP: Alright

src/engine/lighting_engine.cpp
AgentXLP: Alright

src/engine/lighting_engine.cpp
AgentXLP: Will fix

src/engine/lighting_engine.cpp
AgentXLP: Good idea

src/engine/lighting_engine.cpp
AgentXLP: Will fix

src/engine/lighting_engine.cpp
AgentXLP: Good idea

PeachyPeachSM64:
sounds good to me too

PR #1117, opened 2026-02-16
Fix oob read/write caused by sync_object_init_field

PeachyPeachSM64:
Fix out-of-bounds read/write caused by `sync_object_init_field`, that assumed every pointer passed to it points to a 32-bit object.


src/pc/network/sync_object.c
```diff
@@ -182,37 +182,20 @@ struct SyncObject* sync_object_init(struct Object *o, float maxSyncDistance) {
...
-    so->extraFields[index] = field;
-    so->extraFieldsSize[index] = 32;
-}
-
-void sync_object_init_field_with_size(struct Object *o, void* field, u8 size) {
-    if (o->coopFlags & COOP_OBJ_FLAG_NON_SYNC) { return; }
-    if (o->oSyncID == 0) { return; }
-
-    SOFT_ASSERT(size == 8 || size == 16 || size == 32 || size == 64);
+    SOFT_ASSERT(sizeBytes > 0);
```
Isaac0-dev, line 198: the byte size should never be larger than 8, right? While a buffer could technically be used here, we should at least check it on the macro.

PeachyPeachSM64: nothing enforces size past that point, the packets can accept any size
we just need to make sure it's not 0

Isaac0-dev (approved):

PR #1119, opened 2026-02-18
Fix lives

Blockyyy:
- Fix lives going below 0 right before a game over (also fixes a crash where if you exited with -1 lives and went into a subarea you would crash)
- Restore how lives get removed upon death exit (like in vanilla)
- Prevent bubbling from happening if you don't have enough lives (this fixes the annoying bug where you would get put back in a bubble and get popped immediately after over and over again)


src/game/mario.c
```diff
@@ -419,12 +419,13 @@ bool mario_is_ground_pound_landing(struct MarioState *m) {
         (!(m->action & ACT_FLAG_AIR) && (determine_interaction(m, m->marioObj) & INT_GROUND_POUND));
 }
 
-bool mario_can_bubble(struct MarioState* m) {
+bool mario_can_bubble(struct MarioState* m, OPTIONAL u8 unallow) {
```
EmeraldLoc: `unallow` is not a great variable name, probably because it shouldn't exist. This function should just return false if Mario's health is less than or equal to zero. Not one place is this not the same, it's always ran as `mario_can_bubble(m, m->numLives <= 0)`. This argument should not exist imo

src/game/mario.c
```diff
@@ -444,12 +445,12 @@ void mario_set_bubbled(struct MarioState* m) {
     if (m->playerIndex != 0) { return; }
     if (m->action == ACT_BUBBLED) { return; }
 
+    m->numLives--;
+
     gLocalBubbleCounter = 20;
 
     drop_and_set_mario_action(m, ACT_BUBBLED, 0);
-    if (m->numLives > -1) {
```
EmeraldLoc, line 450: Why did you remove this check? `mario_set_bubbled` is exposed to the Lua API, and lives shouldn't go negative there. The check is wrong though, it should be `if (m->numLives > 0) { m->numLives-- }`

EmeraldLoc (changes requested):

src/game/mario.c
Blockyyy: > unallow is not a great variable name, probably because it shouldn't exist.

only reason i have made that argument a thing is to maintain mod compatibility since mods like better bubbling exist, if i were to always return false if mario's health was zero then you wouldn't be able to bubble with that mod on while at zero lives

src/game/mario.c
Blockyyy: oops, i was experimenting and forgot to revert that

autogen/lua_definitions/functions.lua
```diff
@@ -6007,6 +6007,11 @@ function launch_mario_until_land(m, endAction, animation, forwardVel)
     -- ...
 end
 
+--- @param m MarioState
+function lose_life_after_death_exit(m)
+    -- ...
+end
```
PeachyPeachSM64: why is this exposed? I don't think it should

Blockyyy: it was done automatically

autogen/convert_functions.py
```diff
@@ -109,7 +109,7 @@
     "src/engine/surface_collision.h":           [ " debug_", "f32_find_wall_collision" ],
     "src/game/mario_actions_airborne.c":        [ "^[us]32 act_.*" ],
     "src/game/mario_actions_automatic.c":       [ "^[us]32 act_.*" ],
-    "src/game/mario_actions_cutscene.c":        [ "^[us]32 act_.*", " geo_", "spawn_obj", "print_displaying_credits_entry" ],
+    "src/game/mario_actions_cutscene.c":        [ "^[us]32 act_.*", " geo_", "spawn_obj", "print_displaying_credits_entry", "lose_life_after_death_exit" ],
```
Isaac0-dev:
```suggestion
    "src/game/mario_actions_cutscene.c":        [ "^[us]32 act_.*", " geo_", "spawn_obj", "print_displaying_credits_entry" ],
```

src/game/mario_actions_cutscene.c
```diff
@@ -1437,6 +1436,12 @@ s32 act_exit_land_save_dialog(struct MarioState *m) {
     return FALSE;
 }
 
+void lose_life_after_death_exit(struct MarioState *m) {
```
Isaac0-dev:
```suggestion
static void lose_life_after_death_exit(struct MarioState *m) {
```

Isaac0-dev (commented):
looks good, maybe just make that a static function since it's not used elsewhere

PR #1124, opened 2026-02-24
Move HOOK_ON_ADD_SURFACE to before cell calculation

djoslin0:
Moving it before the cell calculation should allow more possibilities
in runtime collision modification


xLuigiGamerx:
wouldn't it be better if you just made a before hook and allowed returning a new surface table

AgentXLP:
I think this hook is so new and barely used that we can feasibly get away with changing the spot its called at without having to make a new hook for it

PR #1126, opened 2026-02-24
fixed error models in sm74 jrb area 2 and mad toads transparency

ColbyRayz:
rotating block with fire and sinking rocks
image
image


ColbyRayz:
image

PR #1131, opened 2026-03-01
DJUI Text improvements and bug fixes

PeachyPeachSM64:
- Drastically improve and optimize display list usage for text:

https://github.com/user-attachments/assets/f0f45c1c-54a5-4f69-a2ec-062c17d09cd4

- Add text alignment with and without interpolation, and color codes, new lines and tabs are now handled properly:

https://github.com/user-attachments/assets/7c1d56f7-ca9b-4b73-8478-ad8d8128c9ca

- Restored alpha color code (`#RGBA` or `#RRGGBBAA`) for `djui_print_text` functions (regular DJUI text allows alpha color codes, but ignores the alpha component)
- Add constants for common text alignment and rotation pivot values
- Fix interpolation issues with all `djui_hud` elements
- A few autogen fixes (missing `number` type for constants, missing return values for some functions)
- Fix recolorable hud font offsets on `e`, `i` and `o` letters


src/pc/djui/djui_gfx.c
```diff
@@ -137,14 +138,10 @@ void djui_gfx_render_texture_tile(const Texture* texture, u32 w, u32 h, u8 fmt,
 
     f32 aspect = tileH ? ((f32)tileW / (f32)tileH) : 1;
 
-    // I don't know why adding 1 to all of the UVs seems to fix rendering, but it does...
-    // this should be tested carefully. it definitely fixes some stuff, but what does it break?
-    f32 offsetX = (font ? -1024.0f / (f32)w : 0) + 1;
```
Isaac0-dev: are you completely certain that it's safe to make this change? 
Do you understand why the condition existed, and have you confirmed that's no longer necessary?
What have you done to test that it's safe? (including what mods have you tested)

src/pc/djui/djui_hud_utils.c
```diff
@@ -67,8 +97,16 @@ struct GlobalTextures gGlobalTextures = {
     .wario_head   = { .texture = texture_hud_char_wario_head,   "texture_hud_char_wario_head",   .width = 16, .height = 16, .format = G_IM_FMT_RGBA, .size = G_IM_SIZ_16b }
 };
 
+static inline const struct DjuiFont *djui_hud_get_text_font() {
+    return gDjuiFonts[sHudUtilsState.font < 0 ? FONT_NORMAL : sHudUtilsState.font];
```
Isaac0-dev, line 103: if `font > FONT_COUNT` oob array access.
Likely not a problem right now since `djui_hud_set_font` has the safety check.

src/pc/djui/djui_hud_utils.h
Isaac0-dev: I think we should have `FONT_LEGACY = -1,` here to make the existence the value more explicit.

src/pc/djui/djui_hud_utils.c
```diff
@@ -364,142 +478,207 @@ void djui_hud_reset_scissor(void) {
...
 
         c = djui_unicode_next_char(c);
     }
+    return max(width, maxWidth) * font->defaultFontScale;
+}
 
-    // pop
-    gSPPopMatrix(gDisplayListHead++, G_MTX_MODELVIEW);
+static Mtx *allocate_dl_translation_matrix() {
+    Mtx *matrix = (Mtx *) alloc_display_list(sizeof(Mtx));
```
Isaac0-dev, line 526: just thought i'd mention that `alloc_display_list` can be very slow if used many times per frame. It seems that if the dl pool has many entries it starts reallocating memory for every new entry instead of reusing it. The bug is in the GrowingPool module as a whole though.

src/pc/djui/djui_hud_utils.c
```diff
@@ -24,13 +24,43 @@
...
+    .font = FONT_NORMAL,
+    .color = { 255, 255, 255, 255 },
+    .rotation = {
+        .degrees = { 0, 0 },
+        .pivotX = { ROTATION_PIVOT_X_LEFT, ROTATION_PIVOT_X_LEFT },
+        .pivotY = { ROTATION_PIVOT_Y_TOP, ROTATION_PIVOT_Y_TOP },
+    },
+    .textAlignment = {
+        .h = { TEXT_HALIGN_LEFT, TEXT_HALIGN_LEFT },
+        .v = { TEXT_VALIGN_TOP, TEXT_VALIGN_TOP },
```
Isaac0-dev: could use a macro so you don't have to repeat stuff here: `#define INTERP(v) {v, v}`

src/pc/mods/mods.c
```diff
@@ -188,8 +188,8 @@ static void mods_sort(struct Mods* mods) {
         struct Mod* mod = mods->entries[i];
         for (s32 j = 0; j < i; ++j) {
             struct Mod* mod2 = mods->entries[j];
-            char* name = str_remove_color_codes(mod->name);
-            char* name2 = str_remove_color_codes(mod2->name);
+            char* name = djui_text_get_uncolored_string(mod->name);
```
Isaac0-dev: I know this was an issue before too, but i think `djui_text_get_uncolored_string` should accept another parameter which can allow the caller to pass in a buffer for `djui_text_get_uncolored_string` to use.
That way you don't have to `strdup` every time, especially in a loop like this case. 
For this case, you can declare static buffers with `MOD_NAME_MAX_LENGTH` before the loop and pass them inside the loop.

src/pc/djui/djui_hud_utils.c
```diff
@@ -132,140 +179,207 @@ void patch_djui_hud_before(void) {
...
     gDjuiHudUtilsZ = savedZ;
 }
 
+struct InterpHud *djui_hud_create_interp() {
+    if (sInterpHudCount >= MAX_INTERP_HUD) { return NULL; }
+
+    struct InterpHud *interp = &sInterpHuds[sInterpHudCount++];
+    interp->z = gDjuiHudUtilsZ;
+    interp->state = sHudUtilsState;
+    interp->gfx = growing_array_init(interp->gfx, 8, malloc, free);
```
Isaac0-dev: memory leak? you shouldn't need to realloc the growing array every frame (`patch_djui_hud_before` sees to that). just clear it

src/pc/djui/djui_hud_utils.c
```diff
@@ -132,140 +179,207 @@ void patch_djui_hud_before(void) {
...
+        sHudUtilsState.color.b = 255;
+        sHudUtilsState.color.a = 255;
         sColorAltered = FALSE;
         gDPSetEnvColor(gDisplayListHead++, 255, 255, 255, 255);
     }
 }
 
-struct HudUtilsRotation* djui_hud_get_rotation(void) {
-    return &sRotation;
+void djui_hud_get_rotation(RET s16 *rotation, RET f32 *pivotX, RET f32 *pivotY) {
```
Isaac0-dev, line 346: since you changed the return type, mods that used it will break

Isaac0-dev (changes requested):
This looks like it'll significantly improve performance, good idea!
What mods have you tested with? I have tested the hard mode nether portal effect and "so retro".
~~In my testing these changes cause no issues.~~ - Edit: Actually I have been watching my system monitor over the last 30 minutes and it is clear that there is a memory leak. Coop has slowly but progressively consumed ~35mb more than when I started watching it.

src/pc/djui/djui_hud_utils.h
PeachyPeachSM64: good idea

src/pc/djui/djui_hud_utils.c
PeachyPeachSM64: yes, the safety check in djui_hud_set_font ensures that sHudUtilsState.font stays within bounds

src/pc/djui/djui_hud_utils.c
PeachyPeachSM64: will do

src/pc/mods/mods.c
PeachyPeachSM64: there is `djui_text_remove_colors`
but you're right, there is no function that takes src and dest, copies src to dest and then removes color codes from dest

src/pc/djui/djui_hud_utils.c
PeachyPeachSM64: growing_array_init frees the previous array
but that's right, I should just clear the buffer instead of reallocating it

src/pc/djui/djui_hud_utils.c
PeachyPeachSM64: no existing mod uses djui_hud_get_rotation
at least not the mods extracted from the 2020-2025 archive

src/pc/djui/djui_hud_utils.h
PeachyPeachSM64: done

src/pc/djui/djui_hud_utils.c
PeachyPeachSM64: done

src/pc/mods/mods.c
PeachyPeachSM64: done

src/pc/djui/djui_hud_utils.c
PeachyPeachSM64: done

Isaac0-dev (approved):
looks good!

PR #1129, opened 2026-03-01
Add searchbox, add it to mod panel and dynos panel

EmeraldLoc:
Pretty much title, here's a pic of what it looks like
Screenshot 2026-02-28 at 9 35 58 PM
Something to note though is I had to redo some of the input box stuff relating to theming. This means the chatbox now looks like this
Screenshot 2026-02-28 at 9 36 43 PM
Screenshot 2026-02-28 at 9 37 11 PM
If need be, and depending on feedback, I can revert the theming changes for the chatbox specifically.

Screenshot 2026-02-28 at 9 38 14 PM

Other input boxes related to theming have been changed aswell by side effect, but these look nice so I don't see a reason to revert it. It's proper now.

PR #1132, opened 2026-03-02
Add support for more save files

EmeraldLoc:
Mainly title. Instead of all save data being stored in a single file, each individual save is stored in multiple files in the `saves` folder (can easily be customized if necessary). Your current save will automatically be converted to the new one. Menu saves were removed, as they aren't needed, and hi score coins are per save instead of the entire game (which was never working well anyways). `touch_coin_score_age` has been removed, as currently it appears no mods use it, and is no longer necessary. The maximum number of save files is 64.

Here is what the ui looks like:

https://github.com/user-attachments/assets/f9892d7c-5807-4f5c-a7a5-bf841605bd88

PR is marked as draft due to syncing needing to be tested, these needing be translated:

```ini
CREATE_SAVE = "Create Save"
CREATE = "Create"
CREATE_TITLE = "CREATE"
CREATE_NAME = "Set Save File @'s Name:"
```

and I also want https://github.com/coop-deluxe/sm64coopdx/pull/1129 to be merged first so I can implement it into this pr.


DM-kun:
German
```ini
IMPORT_SAVE_SUCCESS = "\\#a0ffa0\\Spielstand importiert:\n\\#dcdcdc\\'@'"
```

mingokrb:
here (Portuguese)
```ini
IMPORT_SAVE_SUCCESS = "\\#a0ffa0\\Arquivo importado\n\\#dcdcdc\\'@'"
```

EmeraldLoc:
This should be ready for a more thorough review

wbmarioo:
The only issue I have noticed is names of the save file. On a normal coopdx window, I made a save file titled "Triple Jump." If I open this PR's coop, it will not transfer. It will also reset the name on a normal window back to SM64. Here's a more detailed explanation:
1. In a normal sm64coopdx window, name a save file A-D "Test"
2. Launch the PR coopdx window. Notice that the save file names A-D are SM64 (or a different name, but only if you already renamed them on this version). Then, close the window.
3. Launch a normal sm64coopdx window. The save file name will be reset to SM64.
Extra Note: The name given on the PR version will never be affected by this. If I name a save file on this PR 'Triple Jump', then launch a normal coop window, the save file on the NORMAL window will be SM64, but relaunching the PR still has the file named 'Triple Jump', or whatever you named it.

EmeraldLoc:
Issue above has been resolved. Note that going back to an old version will still not have your old names preserved, as I don't want dead space in the configfile, but switching to any version will not nuke the PR's names. Now when migrating from the old system to the new, your save file names are transferred over.

Isaac0-dev:
conflicts

EmeraldLoc:
Lots of things to address, will reopen when ready :)

src/pc/mods/mod_import.c
```diff
@@ -252,10 +254,48 @@ static bool mod_import_zip(char* path, bool* isLua, bool* isDynos) {
...
+    if (status == 0) {
+        // old data is a go, write eeprom data
+        if (save_file_get_amount_of_available_indexes() < 4) {
+            LOG_ERROR("Ran out of save files slots");
+            return false;
+        }
+        for (int i = 0; i < 4; i++) {
+            int file = save_file_get_first_available_index();
+            write_eeprom_data(file, legacySaveBuffer.files[i], sizeof(legacySaveBuffer.files[i]), 0);
+        }
```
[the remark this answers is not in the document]
EmeraldLoc: What do you mean? `write_eeprom_data` calls `osEepromLongWrite` which writes to a file

```c
s32 osEepromLongWrite(UNUSED OSMesgQueue *mq, u8 fileIndex, u8 address, u8 *buffer, int nbytes) {
    if (fileIndex >= NUM_SAVE_FILES) { return -1; }
    if (gOverrideEeprom[fileIndex] != NULL) {
        memcpy(gOverrideEeprom[fileIndex] + address * 8, buffer, nbytes);
        return 0;
    }

    u8 content[EEPROM_SIZE] = { 0 };
    if (address != 0 || nbytes != EEPROM_SIZE) {
        osEepromLongReadFile(mq, fileIndex, 0, content, EEPROM_SIZE);
    }
    memcpy(content + address * 8, buffer, nbytes);

    if (!fs_sys_dir_exists(fs_get_write_path(SAVE_DIRECTORY))) {
        fs_sys_mkdir(fs_get_write_path(SAVE_DIRECTORY));
    }

    char filePath[SYS_MAX_PATH];
    save_file_get_dir(fileIndex, filePath, SYS_MAX_PATH, NULL);
    FILE *fp = fopen(fs_get_write_path(filePath), "wb");
    if (fp == NULL) {
        return -1;
    }
    s32 ret = fwrite(content, 1, EEPROM_SIZE, fp) == EEPROM_SIZE ? 0 : -1;
    fclose(fp);

    return ret;
}
```

```
s32 write_eeprom_data(u8 file, void *buffer, s32 size, const uintptr_t baseofs) {
    s32 status = 1;

    if (gEepromProbe != 0) {
        s32 triesLeft = 4;
        u32 offset = (u32)baseofs >> 3;

        do {
            block_until_rumble_pak_free();
            triesLeft--;
            status = osEepromLongWrite(&gSIEventMesgQueue, file, offset, buffer, size);
            release_rumble_pak_control();
        } while (triesLeft > 0 && status != 0);
    }

    return status;
}
```

src/pc/mods/mod_import.c
```diff
@@ -252,10 +254,48 @@ static bool mod_import_zip(char* path, bool* isLua, bool* isDynos) {
...
+    if (status == 0) {
+        // data is a go, write to eeprom data
+        write_eeprom_data(firstIndex, gSaveBuffer.files[firstIndex], sizeof(gSaveBuffer.files[firstIndex]), 0);
+        LOG_INFO("Imported save: '%s'", src);
+        save_file_load_all(TRUE);
+        return true;
+    }
+
+    return false;
+}
```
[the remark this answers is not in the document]
EmeraldLoc: If it returns nonzero it does not write any garbage data, see function implementation

```c
s32 osEepromLongRead(UNUSED OSMesgQueue *mq, u8 address, u8 *buffer, int nbytes, char *path, size_t size) {
    u8 content[size];
    s32 ret = -1;

    FILE *fp = fopen(path, "rb");
    if (!fp) {
        return -1;
    }
    if (fread(content, 1, size, fp) == size) {
        memcpy(buffer, content + address * 8, nbytes);
        ret = 0;
    }
    fclose(fp);

    return ret;
}

s32 osEepromLongReadFile(UNUSED OSMesgQueue *mq, u8 fileIndex, u8 address, u8 *buffer, int nbytes) {
    if (fileIndex >= NUM_SAVE_FILES) { return -1; }
    if (gOverrideEeprom[fileIndex] != NULL) {
        memcpy(buffer, gOverrideEeprom[fileIndex] + address * 8, nbytes);
        return 0;
    }

    u8 content[EEPROM_SIZE];
    s32 ret = -1;

    char filePath[SYS_MAX_PATH];
    save_file_get_dir(fileIndex, filePath, SYS_MAX_PATH, NULL);
    fs_file_t *fp = fs_open(filePath);
    if (fp == NULL) {
        return -1;
    }
    if (fs_read(fp, content, EEPROM_SIZE) == EEPROM_SIZE) {
        memcpy(buffer, content + address * 8, nbytes);
        ret = 0;
    }
    fs_close(fp);

    return ret;
}
```

docs/lua/functions-6.md
```diff
@@ -4795,26 +4795,22 @@ Gets the level number's corresponding course number
...
 ### Lua Example
-`touch_coin_score_age(fileIndex, courseIndex)`
+`save_file_get_filename_at_index(fileIndex, outFilename[MAX_SAVE_NAME_STRING])`
 
 ### Parameters
 | Field | Type |
 | ----- | ---- |
 | fileIndex | `integer` |
-| courseIndex | `integer` |
+| outFilename[MAX_SAVE_NAME_STRING] | `integer` |
```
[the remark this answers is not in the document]
EmeraldLoc: Autogenerated by autogen, should not be reviewed

docs/lua/functions-6.md
```diff
@@ -4795,26 +4795,22 @@ Gets the level number's corresponding course number
...
```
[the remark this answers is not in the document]
EmeraldLoc: Autogenerated by autogen, should not be reviewed

EmeraldLoc:
Some issues with headers I'd like to fix, marking as draft

EmeraldLoc:
There's still a problem but unsure on the best way to fix that wouldn't feel out of the scope. Will let a reviewer analyze, but here's the warning:

```
In file included from src/goddard/renderer.c:25:
In file included from include/types.h:599:
In file included from src/game/characters.h:5:
In file included from src/pc/configfile.h:8:
In file included from src/game/area.h:7:
src/game/camera.h:51:9: warning: 'ABS' macro redefined [-Wmacro-redefined]
   51 | #define ABS(x) ((x) > 0.f ? (x) : -(x))
      |         ^
src/goddard/gd_macros.h:13:9: note: previous definition is here
   13 | #define ABS(val) (((val) < 0 ? (-(val)) : (val)))
      |         ^
```

The issue is that gd_macros.h defines ABS, and camera.h defines ABS. The compiler is actually pretty clear on the issue. I'd think the best approach is either renaming or removing the camera's abs macro and either using standard abs and fabsf functions or including gd_macros. The latter feels wrong.

xLuigiGamerx (commented):
Merge conflicts

include/PR/os_eeprom.h
```diff
@@ -94,8 +94,9 @@ extern "C" {
 extern s32 osEepromProbe(OSMesgQueue *);
 extern s32 osEepromRead(OSMesgQueue *, u8, u8 *);
 extern s32 osEepromWrite(OSMesgQueue *, u8, u8 *);
-extern s32 osEepromLongRead(OSMesgQueue *, u8, u8 *, int);
-extern s32 osEepromLongWrite(OSMesgQueue *, u8, u8 *, int);
+extern s32 osEepromLongRead(OSMesgQueue *, u8 , u8 *, int, char *, size_t);
+extern s32 osEepromLongReadFile(OSMesgQueue *, u8, u8, u8 *, int);
+extern s32 osEepromLongWrite(OSMesgQueue *, u8, u8, u8 *, int);
```
PeachyPeachSM64: I'm not a big fan of modifying the N64 SDK... Not only the function names are not the most explicit ones, but these are just supposed to be a compatibility layer for the PC port.
Since these are also only used in save_file.c, they should be defined as static in save_file.c, not split into 3 different files.
If you introduce a new way to read and write save files, define it directly in the corresponding C file and do not stick with the N64 SDK style, which no longer makes sense.

src/pc/djui/djui_panel_host.c
```diff
@@ -175,7 +175,9 @@ void djui_panel_host_create(struct DjuiBase* caller) {
             djui_text_set_drop_shadow(text1, 64, 64, 64, 100);
 
             char starString[64] = { 0 };
-            snprintf(starString, 64, "%c x%d - %s", '~' + 1, save_file_get_total_star_count(configHostSaveSlot - 1, 0, 24), configSaveNames[configHostSaveSlot - 1]);
+            char saveName[MAX_SAVE_NAME_STRING] = { 0 };
+            save_file_get_filename_at_index(configHostSaveSlot - 1, saveName);
+            snprintf(starString, 64, "%c x%d - %s", '~' + 1, save_file_get_total_star_count(configHostSaveSlot - 1, 0, 24), saveName);
```
PeachyPeachSM64, line 180: just 0x7F instead of '~' + 1, or define a macro for that specific symbol
(I know it was there before, but it's ugly)

src/pc/djui/djui_panel_host_save.c
```diff
@@ -6,39 +6,100 @@
...
+static bool is_save_name_input_valid() {
+    if (sSaveName[0] == '\0'
+        || strstr(sSaveName, ".")
+        || strstr(sSaveName, "/")
+        || strstr(sSaveName, "\\")) {
+        return false;
+    }
+
+    return true;
+}
```
PeachyPeachSM64, line 44: I will come back to this later, but dumping the save name in the filename is not a good idea

src/pc/djui/djui_panel_host_save.c
```diff
@@ -6,39 +6,100 @@
...
+    char filePath[SYS_MAX_PATH];
+    save_file_get_dir(sButtonTag, filePath, SYS_MAX_PATH, sSaveName);
+    if (fs_sys_file_exists(fs_get_write_path(filePath))) { return; }
+    u8 content[EEPROM_SIZE] = { 0 };
+    FILE *fp = fopen(fs_get_write_path(filePath), "wb");
+    if (fp == NULL) { return; }
+    fwrite(content, 1, EEPROM_SIZE, fp);
+    fclose(fp);
+    djui_panel_host_reload_saves();
+    djui_panel_menu_back(caller);
```
PeachyPeachSM64, line 79: please put some blank newlines, in particular after early returns
there is like 5 different actions performed in this function packed into a single blob

src/pc/djui/djui_panel_host_save.c
```diff
@@ -6,39 +6,100 @@
...
+        || strstr(sSaveName, ".")
+        || strstr(sSaveName, "/")
+        || strstr(sSaveName, "\\")) {
+        return false;
+    }
+
+    return true;
+}
 
 static void djui_panel_host_save_save_name_change(UNUSED struct DjuiBase* caller) {
```
PeachyPeachSM64, line 46: weird function name, when `djui_panel_host_save_update_save_name` is right above

src/pc/djui/djui_panel_host_save.c
```diff
@@ -6,39 +6,100 @@
...
+        djui_inputbox_set_text_color(sSaveNameInputBox, 255, 0, 0, 255);
+    } else {
+        djui_inputbox_set_text_color(sSaveNameInputBox, textColor->r, textColor->g, textColor->b, textColor->a);
     }
 }
 
-static bool djui_panel_edit_back(UNUSED struct DjuiBase* caller) {
-    if (configSaveNames[sButtonTag][0] == '\0') {
-        snprintf(configSaveNames[sButtonTag], MAX_SAVE_NAME_STRING, "SM64");
+static void djui_panel_create_create(struct DjuiBase* caller) {
```
PeachyPeachSM64, line 62: create create?

src/pc/djui/djui_panel_host_save.c
```diff
@@ -6,39 +6,100 @@
...
+    u8 content[EEPROM_SIZE] = { 0 };
+    FILE *fp = fopen(fs_get_write_path(filePath), "wb");
+    if (fp == NULL) { return; }
+    fwrite(content, 1, EEPROM_SIZE, fp);
+    fclose(fp);
+    djui_panel_host_reload_saves();
+    djui_panel_menu_back(caller);
+}
+
+static void djui_panel_edit_save(UNUSED struct DjuiBase* caller) {
```
PeachyPeachSM64, line 82: Well, there is no fixed convention for this stuff?
Sometimes it's prefixed by `djui_panel_host_save`, sometimes it's just `djui_panel_save` and sometimes it's not?
What's the rule?
Please stick to a naming convention for the whole file.

src/pc/djui/djui_panel_host_save.c
```diff
@@ -6,39 +6,100 @@
...
+    struct DjuiThreePanel* panel = djui_panel_menu_create(sEditing ? DLANG(HOST_SAVE, EDIT_TITLE) : DLANG(HOST_SAVE, CREATE_TITLE), false);
     struct DjuiBase* body = djui_three_panel_get_body(panel);
     {
         struct DjuiRect* rect1 = djui_rect_container_create(body, 32);
         {
             char buffer[64] = { 0 };
-            djui_language_replace(DLANG(HOST_SAVE, EDIT_NAME), buffer, 64, '@', sSaveLetters[sButtonTag]);
+            char slotBuffer[4];
+            snprintf(slotBuffer, sizeof(slotBuffer), "%d", sButtonTag + 1);
+            djui_language_replace(sEditing ? DLANG(HOST_SAVE, EDIT_NAME) : DLANG(HOST_SAVE, CREATE_NAME), buffer, 64, '@', slotBuffer);
```
PeachyPeachSM64, line 102: sizeof(buffer) instead of hardcoded 64 in case someone changes it later

src/pc/djui/djui_panel_host_save.c
```diff
@@ -50,23 +111,54 @@ static void djui_panel_edit_create(struct DjuiBase* caller) {
...
+    djui_panel_edit_create(caller);
+}
+
+static char *get_save_name(int slot) {
+    static char sStarString[64] = { 0 };
+    char name[MAX_SAVE_NAME_STRING] = { 0 };
+    save_file_get_filename_at_index(slot, name);
+    snprintf(sStarString, 64, "%c x%d - %s", '~' + 1, save_file_get_total_star_count(slot, 0, 24), name);
+    return sStarString;
+}
```
PeachyPeachSM64, line 152: Duplicate code with djui_panel_host.c, which should use this function instead
Should take a char* buffer and max size as inputs instead of relying on a static string

src/pc/configfile.c
```diff
@@ -581,10 +582,11 @@ static void save_name_read(char** tokens, int numTokens) {
...
 }
 
-static void save_name_write(FILE* file) {
-    for (int i = 0; i < NUM_SAVE_FILES; i++) {
+static void save_name_write(UNUSED FILE* file) {
+    // don't write save name data anymore, this only exists for compat
+    /*for (int i = 0; i < 4; i++) {
         fprintf(file, "%s %d %s\n", "save-name:", i, configSaveNames[i]);
-    }
+    }*/
```
PeachyPeachSM64, line 589: delete this, the comment is enough

src/pc/configfile.h
```diff
@@ -4,6 +4,9 @@
 #include <stdbool.h>
 #include <PR/ultratypes.h>
 #include "game/player_palette.h"
+#include "pc/lua/smlua_autogen.h"
+#include "game/area.h"
+#include "game/save_file.h"
```
PeachyPeachSM64, line 9: Why do you need this in the header? If it's only used in configfile.c, put theme here instead

src/game/save_file.c
```diff
@@ -144,19 +125,19 @@ static inline void bswap_savefile(struct SaveFile *data) {
  * Read from EEPROM to a given address.
  * The EEPROM address is computed using the offset of the destination address from gSaveBuffer.
  * Try at most 4 times, and return 0 on success. On failure, return the status returned from
- * osEepromLongRead. It also returns 0 if EEPROM isn't loaded correctly in the system.
+ * osEepromLongReadFile. It also returns 0 if EEPROM isn't loaded correctly in the system.
  */
-static s32 read_eeprom_data(void *buffer, s32 size) {
+s32 read_eeprom_data(u8 file, void *buffer, s32 size) {
```
PeachyPeachSM64: Why u8 and not a larger type? There isn't much sense to use a small data type and would only cause issues in the future

src/game/save_file.c
```diff
@@ -170,7 +151,7 @@ static s32 read_eeprom_data(void *buffer, s32 size) {
  * Try at most 4 times, and return 0 on success. On failure, return the status returned from
  * osEepromLongWrite. Unlike read_eeprom_data, return 1 if EEPROM isn't loaded.
  */
-static s32 write_eeprom_data(void *buffer, s32 size, const uintptr_t baseofs) {
+s32 write_eeprom_data(u8 file, void *buffer, s32 size, const uintptr_t baseofs) {
```
PeachyPeachSM64: same here

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+        write_eeprom_data(i, saveBuffer.files[i], sizeof(saveBuffer.files[i]), 0);
+        save_file_rename_file(i, configSaveNames[i]);
+    }
+}
+
+/**
+ * Gets all save file names
+*/
+void save_file_get_all_filenames(char filenames[NUM_SAVE_FILES][MAX_SAVE_NAME_STRING]) {
+    char* directory = (char*)fs_get_write_path(SAVE_DIRECTORY);
```
PeachyPeachSM64, line 323: `const char *directory`
please use `char *` only when necessary

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+    DIR* d = opendir(directory);
+    if (!d) { return; }
+
+    // iterate
+    char path[SYS_MAX_PATH] = { 0 };
+    while ((dir = readdir(d)) != NULL) {
+        // sanity check
+        if (!directory_sanity_check(dir, directory, path)) { continue; }
+        snprintf(path, SYS_MAX_PATH, "%s", dir->d_name);
+        if (strlen(path) == 0 || strlen(path) >= 256) { continue; }
```
PeachyPeachSM64, line 338: why?

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+
+        // verify filename follows format (index)_(name)(SAVE_EXTENSION)
+        int index = 0;
+        char name[MAX_SAVE_NAME_STRING];
+        char extension[12];
+        if (sscanf(path, "%d_%31[^.]%11s", &index, name, extension) == 3) {
```
PeachyPeachSM64, line 344: Hardcoded string length in a very hidden spot will make maintainability very painful... Same for the extension.
You could use something like this: https://stackoverflow.com/a/5873453

However, I disagree on embedding the save name in the filename. It disallows the use of some characters and you take the risk of having very weird filenames (control chars, whitespaces of any kind...).

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+
+    closedir(d);
+}
+
+/**
+ * Gets save file name for index. If it does not exist, use SM64
+*/
+void save_file_get_filename_at_index(int fileIndex, char outFilename[MAX_SAVE_NAME_STRING]) {
+    char filenames[NUM_SAVE_FILES][MAX_SAVE_NAME_STRING] = { 0 };
+    save_file_get_all_filenames(filenames);
```
PeachyPeachSM64, line 360: Not very efficient... You're reading the entire user's save folder everytime you call this?
Also if any filename change happens between two calls or if a file becomes unavailable for some reason, you'll end up in a bad state...
Same thing for the following functions, which seem to all call this one...

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+        snprintf(outFilename, MAX_SAVE_NAME_STRING, "%s", filenames[fileIndex]);
+    } else {
+        snprintf(outFilename, MAX_SAVE_NAME_STRING, "%s", "SM64");
+    }
+}
+
+/**
+ * Get directory for a save file at index
+*/
+void save_file_get_dir(int fileIndex, char* outPath, size_t size, char* overrideName) {
```
PeachyPeachSM64, line 371: `const char *overrideName`
If you're not modifying a string, declare it const please

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+}
+
+/**
+ * Renames a specific save file. Returns false on failure
+*/
+void save_file_rename_file(s32 fileIndex, char* name) {
+    if (!fs_sys_dir_exists(fs_get_write_path(SAVE_DIRECTORY))) { return; }
+    if (strstr(name, ".")) { return; }
+    if (strstr(name, "/")) { return; }
+    if (strstr(name, "\\")) { return; }
```
PeachyPeachSM64, line 428: This is an issue I mentioned in one of my previous messages and explain in the review message.

src/game/save_file.c
```diff
@@ -428,20 +453,27 @@ void save_file_do_save(s32 fileIndex, s8 forceSave) {
...
     if (INVALID_FILE_INDEX(fileIndex)) { return; }
 
-    touch_high_score_ages(fileIndex);
     bzero(&gSaveBuffer.files[fileIndex][0], sizeof(gSaveBuffer.files[fileIndex][0]));
     bzero(&gSaveBuffer.files[fileIndex][1], sizeof(gSaveBuffer.files[fileIndex][1]));
 
-    gSaveFileModified = TRUE;
-    save_file_do_save(fileIndex, TRUE);
+    if (!fs_sys_dir_exists(fs_get_write_path(SAVE_DIRECTORY))) return;
+    char filepath[256];
```
PeachyPeachSM64, line 497: Not the first time I see this, but why all these paths are not SYS_MAX_PATH? Might screw users with very long or deep folder hierarchy

src/game/save_file.c
```diff
@@ -396,14 +267,168 @@ static u8 save_file_need_bswap(const struct SaveBuffer *buf) {
...
+    closedir(d);
+}
+
+/**
+ * Gets save file name for index. If it does not exist, use SM64
+*/
+void save_file_get_filename_at_index(int fileIndex, char outFilename[MAX_SAVE_NAME_STRING]) {
+    char filenames[NUM_SAVE_FILES][MAX_SAVE_NAME_STRING] = { 0 };
+    save_file_get_all_filenames(filenames);
+    if (filenames[fileIndex] && filenames[fileIndex][0] != '\0') {
```
PeachyPeachSM64, line 361: filenames[fileIndex] always evaluates to true

PeachyPeachSM64 (changes requested):
This is only a partial review. But I noticed some points I disagree with:
- It is a pretty good start, but feels a bit incomplete... Why are we still limited to 64 save files? It's not like memory management is hard, we already have all the tools for this.
- The main purpose of increasing the save files was to have an easier time with managing the saves with rom-hacks. However, I don't see any feature that points in that direction... We increased the number of files, and that's it?
  - Why not, for example, adding a simple feature that automatically create or switch to a rom-hack save file when the rom-hack is detected in the selected mods? The savefile could also easily use the rom-hack name.
  - There is some space in the file select screen that could be used for a button like "Create save file for rom-hack" or something like this.
- Speaking of savefile names, I don't think it's a good idea to dump it directly in the filename. It restricts the set of characters you can use, and you have to worry about weird or problematic filenames. Why not storing it in the save data instead? Since you're already using a custom save file format, nothing stops you from adding the savefile name.

Other details, but not as important:
- Most of languages are also missing.
- Please change all instances of `int`, `float` into `s32`, `f32`, etc...

EmeraldLoc:
That’s one hell of a review. I’ll end up taking a look at this in more detail later, but I’ll probably do a partial rewrite of it since I wrote this a really long time ago. Don’t want to know what my other prs look like lol

- For static allocation, yes I’ll change it over to use a growing array
- as for the romhack thing, I don’t really plan on implementing that as a feature. That would be quite hard! It depends though, I can’t remember the last time I looked at this code so I don’t know how metadata is stored (probably scuffed but i can hope)
- absolutely on the save file name
- I hate languages
- Ok, that one (along with everything else) came from a lack of experience, will fix!

thanks for the detailed review. Will address these and probably look at tidying up some other code

I should probably self review all my old code since a lot of this is stuff I can catch easily nowadays, nonetheless

PeachyPeachSM64:
> That’s one hell of a review. I’ll end up taking a look at this in more detail later, but I’ll probably do a partial rewrite of it since I wrote this a really long time ago. Don’t want to know what my other prs look like lol
> 
> * For static allocation, yes I’ll change it over to use a growing array
> * as for the romhack thing, I don’t really plan on implementing that as a feature. That would be quite hard! It depends though, I can’t remember the last time I looked at this code so I don’t know how metadata is stored (probably scuffed but i can hope)
> * absolutely on the save file name
> * I hate languages
> * Ok, that one (along with everything else) came from a lack of experience, will fix!
> 
> thanks for the detailed review. Will address these and probably look at tidying up some other code
> 
> I should probably self review all my old code since a lot of this is stuff I can catch easily nowadays, nonetheless

Great! Glad you actually agree on most of these points. This feature is rather old and definitely needs some rewrite.

PR #1140, opened 2026-03-12
Make languages default to english with missing data

EmeraldLoc:
Really just title. Tested by nuking half the Spanish language, it indeed works. Nothing that complicated here, just gonna add a bunch of stuff here ig...

This pr is a massive quality of life change and a huge codebase change. Lots was done to make this pull request happen, This pr does a complete language refactor, allowing for multiple new languages, new systems, and more! Changing a language means all words in the game get changed, every single one! With the old system only some words were translated, now they all are! I know this may not seem that great, but it is truly something special. This pr is something that every bilingual person should enjoy, I don't even speak any other languages and I love this change! This refactor will be an extremely life changing pr, that includes not only language changes, but word changes, and backend changes, and code changes, safety changes, management changes, and more! This is great, a wonderful showcase of what human engineering can do, a showcase of the ability to gain motivation to code truly great things. This is easily the biggest pr I've done. It's like it halo 2 and animal crossing were combined. This pr also allows the language to fall back. This is huge. Have you ever wanted something to fall back on? Languages have that privilege, a privilege indeed, caused by true creativity. This pr truly is one of the greatest. It can even load 2 languages in parallel, that's how you know it's good! Please merge like right now because this pr is the undeniable GOAT. Thank you for reading this, this pr means so much to me, and it would mean the world if it got merged right now. Thank you, have a wonderful, amazing, great, incredible, super night. Keep playing Mario 64, keep coding, keep gaming, and for my signature, my amazing signature. I will give it, as it's such an amazing thing. The way the words are inserted onto the page is unbelievable, here it is, it's coming, get ready, here it comes!

- EmeraldLockdown

### DIFF
--- a/src/pc/djui/djui_language.c
+++ b/src/pc/djui/djui_language.c
@@ -33,14 +33,12 @@ bool djui_language_init(char* lang) {
 }
 
 static char* djui_language_get_english(const char *section, const char *key) {
-    if (!sEnglishLang) { return (char*)key; }
     char* value = (char*)ini_get(sEnglishLang, section, key);
     if (!value) { return (char*)key; }
     return value;
 }
 
 char* djui_language_get(const char *section, const char *key) {
-    if (!sLang) { return djui_language_get_english(section, key); }
     char* value = (char*)ini_get(sLang, section, key);
     if (!value) { return djui_language_get_english(section, key); }
     return value;
--- a/src/pc/djui/djui_language.c
+++ b/src/pc/djui/djui_language.c
@@ -22,7 +22,7 @@ bool djui_language_init(char* lang) {
     if (!lang || lang[0] == '\0') { lang = "English"; }
     snprintf(path, SYS_MAX_PATH, "%s/lang/%s.ini", sys_resource_path(), lang);
 
-    // load normal lang
+    // load user lang
     sLang = ini_load(path);
 
     // load english lang
--- a/src/pc/djui/djui_language.c
+++ b/src/pc/djui/djui_language.c
@@ -7,30 +7,42 @@
 #include "pc/djui/djui_language.h"
 #include "pc/djui/djui_popup.h"
 
+static ini_t* sEnglishLang = NULL;
 static ini_t* sLang = NULL;
 
 bool djui_language_init(char* lang) {
     // free old ini
-    if (sLang != NULL) {
-        ini_free(sLang);
-        sLang = NULL;
-    }
+    ini_free(sLang);
+    sLang = NULL;
+    ini_free(sEnglishLang);
+    sEnglishLang = NULL;
 
     // construct path
     char path[SYS_MAX_PATH] = "";
     if (!lang || lang[0] == '\0') { lang = "English"; }
     snprintf(path, SYS_MAX_PATH, "%s/lang/%s.ini", sys_resource_path(), lang);
 
-    // load
+    // load normal lang
     sLang = ini_load(path);
+
+    // load english lang
+    snprintf(path, SYS_MAX_PATH, "%s/lang/English.ini", sys_resource_path());
+    sEnglishLang = ini_load(path);
 
     return sLang != NULL;
 }
 
-char* djui_language_get(const char *section, const char *key) {
-    if (!sLang) { return (char*)key; }
-    char* value = (char*)ini_get(sLang, section, key);
+static char* djui_language_get_english(const char *section, const char *key) {
+    if (!sEnglishLang) { return (char*)key; }
+    char* value = (char*)ini_get(sEnglishLang, section, key);
     if (!value) { return (char*)key; }
+    return value;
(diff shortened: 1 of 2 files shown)


src/pc/djui/djui_language.c
```diff
@@ -7,30 +7,42 @@
...
+}
+
 char* djui_language_get(const char *section, const char *key) {
-    if (!sLang) { return (char*)key; }
+    if (!sLang) { return djui_language_get_english(section, key); }
     char* value = (char*)ini_get(sLang, section, key);
-    if (!value) { return (char*)key; }
+    if (!value) { return djui_language_get_english(section, key); }
     return value;
 }
```
PeachyPeachSM64, line 45: So much `char *` casts...
`djui_language_get` is supposed to be const, but it is used in some instances of `djui_language_replace`, which takes a `char *` as `src` param...
`src` should also be const, but `djui_unicode_next_char` only accepts `char *`...
`djui_unicode_next_char` should be const as well, but the pointer returned has the same source as the input, which may or may not be const...
This is a mess and I hate it, but it would inevitably need a lot of rewriting to fix...

PeachyPeachSM64: Btw, why not adding a NULL check in `ini_get` directly to avoid duplicating the default return in both `djui_language_get_english` and `djui_language_get`?

EmeraldLoc: > Btw, why not adding a NULL check in ini_get directly to avoid duplicating the default return in both djui_language_get_english and djui_language_get?

By this I assume making `ini_get` return the `key` if the value is NULL, so value is the key? That would make this system require a strcmp, or something of the sort. Because if the value is null, we want to try using the English language first, but if the value doesn't return null, but just the key value, you'd have to compare both of those strings. This is less than ideal. Sure the English func would be cleaner, but the normal func would be very ugly imo.

> So much char * casts...
djui_language_get is supposed to be const, but it is used in some instances of djui_language_replace, which takes a char * as src param...
src should also be const, but djui_unicode_next_char only accepts char *...
djui_unicode_next_char should be const as well, but the pointer returned has the same source as the input, which may or may not be const...

You *could* just make all of it take in a char* instead of const char*. That may not be proper (as the strings being implemented should not be modified, thus, const char*), but it's an option and removes a lot of the casting. Probably not the ideal solution though.

PeachyPeachSM64: > By this I assume making ini_get return the key if the value is NULL, so value is the key? That would make this system require a strcmp, or something of the sort. Because if the value is null, we want to try using the English language first, but if the value doesn't return null, but just the key value, you'd have to compare both of those strings. This is less than ideal. Sure the English func would be cleaner, but the normal func would be very ugly imo.

No, I mean return NULL in ini_get if ini is NULL
then you would just have to
```c
    char* value = (char*)ini_get(sEnglishLang, section, key);
    if (!value) { return (char*)key; }
    return value;
```
removing the extra NULL check for the language

EmeraldLoc: Ah gotcha, yea that makes perfect sense now lol. Did it

AgentXLP:
> Really just title. Tested by nuking half the Spanish language, it indeed works. Nothing that complicated here, just gonna add a bunch of stuff here ig...
> 
> This pr is a massive quality of life change and a huge codebase change. Lots was done to make this pull request happen, This pr does a complete language refactor, allowing for multiple new languages, new systems, and more! Changing a language means all words in the game get changed, every single one! With the old system only some words were translated, now they all are! I know this may not seem that great, but it is truly something special. This pr is something that every bilingual person should enjoy, I don't even speak any other languages and I love this change! This refactor will be an extremely life changing pr, that includes not only language changes, but word changes, and backend changes, and code changes, safety changes, management changes, and more! This is great, a wonderful showcase of what human engineering can do, a showcase of the ability to gain motivation to code truly great things. This is easily the biggest pr I've done. It's like it halo 2 and animal crossing were combined. This pr also allows the language to fall back. This is huge. Have you ever wanted something to fall back on? Languages have that privilege, a privilege indeed, caused by true creativity. This pr truly is one of the greatest. It can even load 2 languages in parallel, that's how you know it's good! Please merge like right now because this pr is the undeniable GOAT. Thank you for reading this, this pr means so much to me, and it would mean the world if it got merged right now. Thank you, have a wonderful, amazing, great, incredible, super night. Keep playing Mario 64, keep coding, keep gaming, and for my signature, my amazing signature. I will give it, as it's such an amazing thing. The way the words are inserted onto the page is unbelievable, here it is, it's coming, get ready, here it comes!
> 
>     * EmeraldLockdown

Wtf lmao

AgentXLP:
What happens if English.ini is missing?

EmeraldLoc:
It'll fallback to the key system like it normally does

PR #1142, opened 2026-03-13
Add dynamic object limit

Sunketchupm:
This was something I have been wanting to do for a while and finally managed to. Thanks to Isaac and FrozenAlex for their help.

The object limit doesn't become truly infinite though. Aside from the obvious limiting factor of computer power, there appears to be other forms of the object limit. Sync objects have their own limit determined by `SYNC_ID_BLOCK_SIZE` (which I don't think I should increase), and the game seems to give up on processing object graphics at a little below 5000 objects (but they are still definitely processed since in testing. I was able to interact with the invisible coins, which were my main test object). The game isn't very stable past that point, so the object limit still exists, but has been increased to 4096 (which isn't very much).

There's no doubt some form of memory leak here since this is my first time using the heap. This, combined with the issues mentioned last paragraph, and uncertainty of how well this will work across the network, is why this is currently a draft. Any help would be appreciated.

Testing mod:
```lua
hook_event(HOOK_MARIO_UPDATE, function (m)

    if m.controller.buttonDown & L_TRIG ~= 0 then
        for i = 1, 100, 1 do
            spawn_sync_object(id_bhvYellowCoin, E_MODEL_YELLOW_COIN, m.pos.x, m.pos.y + 500, m.pos.z, nil)
        end
    end
end)

hook_event(HOOK_UPDATE, function ()
    djui_chat_message_create(tostring(obj_get_count()))
end)
```


Sunketchupm:
Just now attempting this suggestion and I don't actually see a point in using a growing array. It would be useful if I rewrote the object allocation system, but not only do I have a skill issue, I would also like to preserve cloning which I don't know how to transfer over to growing arrays. From what I can tell, cloning can work and doesn't crash because the entire object pool is preallocated, so holding an unloaded object is still holding an object and not a null. It's why my approach was the way it was, so that the vanilla behavior for allocating (not necessarily deallocating though) was used and allows for cloning. However I also think how I did it is a bit dumb and it is a bit unstable, so instead I'm going to rethink that and see if a preallocated array is really needed at all, or even an object pool.

Isaac0-dev:
growing arrays will preserve cloning.
This is because in it's current state, `growing_array_alloc` copies the entire array capacity during array expansions, deleted objects and all. So it doesn't matter that the array won't be preallocated (it's reallocated every expansion).
I'd say growing arrays are a good idea here. They're fast and easy.
They also work well with the existing sm64 object pool allocation system, only very minor tweaks necessary I imagine.

Sunketchupm:
Sorry about the merge mess.

My justification for still not using the growing array really comes down to a skill issue, but also I put a bit more thought on how this should work.
No matter what there will be an object limit, either from syncing or from whatever graphics issue happens at 5000 objects. Even if those were fixed, having so many objects anyways would be pretty susceptible to crashing, either from sheer lag or some instability I don't know about.
Using a regular array that's much bigger but having a soft cap is not only much easier for me to do, but also solves a pretty big reason why we shouldn't just increase the object limit. Just increasing the limit means going through the entire array for every loop that uses the object pool. With a pool that's 3x bigger, that's 3x less performance (Raw numbers haven't been calculated). Using a soft cap no longer has this performance cut for when the pool isn't as large. Compared to using the proper growing array however, this does come at a tradeoff of the pool using 3x more memory though (3kb to 10kb).

Still WIP due to unexplainable crashes/infinite loops that happened with this test mod:
```lua
hook_event(HOOK_MARIO_UPDATE, function (m)

    if m.controller.buttonDown & L_TRIG ~= 0 then
        local id = math.random(id_bhv_max_count)
        local model = math.random(E_MODEL_MAX)
        spawn_sync_object(id, model, m.pos.x, m.pos.y, m.pos.z, nil)
    end
end
```
(Release version of coop doesn't crash with this mod)

Blockyyy:
are you sure the crashes and infinite loops aren't because you got unlucky with some of the behavior rolls

Sunketchupm:
I'm quite sure since the release version only crashed once after stress testing with this a lot and that seemed to be due to syncing something, but this version always loops/crashes (most likely loops, since there is no crash screen and the game has trouble closing).

I am able to avoid the crash by having the game link the entire object pool on level init, but then that defeats the purpose of having this dynamic limit since there aren't that many places that loop through the entire object pool, at least directly.

Sunketchupm:
Closing this since the commit history is going to be very chaotic. Will reopen once I have this fully working in the way I expect it to.

PR #1147, opened 2026-03-14
Mark all unused variables as unused, resolve other warnings

EmeraldLoc:
Was getting annoyed at the obnoxious amount of "Warning: Unused variable" garbage I would get during compilation. So I marked them all as unused. For the most part these changes should be fine, but I did end up altering an `sprintf` in `stb_image_write.h`, as it was marked as unsafe by the compiler. It was an easy fix, but because it's a lib like that, I thought I'd mention it. (Also, I don't think I'm actually meant to be getting unused variable warnings lol, but on macOS the flag isn't being set ¯\_(ツ)_/¯, well, at least it let me do this).


AgentXLP:
If I were you I would just deleted the unused variables and other data but this is probably a better approach than nuking all of it

EmeraldLoc:
I don't mind doing so, I was cautious on deleting anything that seemed like it could ever come back to life, but I could probably do another pass and delete more.

src/pc/apparition.inc.c
```diff
@@ -1,14 +1,14 @@
 #include "types.h"
 
-static ALIGNED8 const Texture apparition_texture_1[] = {
+UNUSED static ALIGNED8 const Texture apparition_texture_1[] = {
```
AgentXLP: These textures are referenced in `rom_asset_load`

src/pc/apparition.inc.c
```diff
@@ -1,14 +1,14 @@
...
 };
-static ALIGNED8 const Texture apparition_texture_2[] = {
+UNUSED static ALIGNED8 const Texture apparition_texture_2[] = {
```
AgentXLP: Same as above

src/pc/apparition.inc.c
```diff
@@ -1,14 +1,14 @@
...
 };
-static ALIGNED8 const Texture apparition_texture_3[] = {
+UNUSED static ALIGNED8 const Texture apparition_texture_3[] = {
```
AgentXLP: Same as above

src/pc/apparition.inc.c
```diff
@@ -1,14 +1,14 @@
...
 };
-static ALIGNED8 const Texture apparition_texture_4[] = {
+UNUSED static ALIGNED8 const Texture apparition_texture_4[] = {
```
AgentXLP: Same as above

src/pc/apparition.inc.c
EmeraldLoc: Screenshot 2026-04-21 at 9 16 40 PM
While that's true, my compiler doesn't seem to agree, which is why it's there. Don't know why this would be happening.

AgentXLP: Weird, just leave it not marked as UNUSED since it is used even if the compiler disagrees

PR #1145, opened 2026-03-13
Make temporary objects actually unique and temporary

PeachyPeachSM64:
- Fixes the `temp` functions always returning the same CObject, that was making it impossible to work with multiple instances (like calling `resolve_and_return_wall_collisions_data` with two different `WallCollisionData`).
- Makes these objects temporary. They're deleted as soon as the GC picks up the underlying Lua object. The GC behavior had to be changed a little to be able to identify and collect unused userdata.
- Removes now obsolete `get_temp_s32_pointer` and `deref_s32_pointer`.

PR #1149, opened 2026-03-14
Fix nametags color and rendering order

PeachyPeachSM64:
**Based on: https://github.com/coop-deluxe/sm64coopdx/pull/1145**

- Fixed nametags color and rendering order:

nametags-order

Nametags now support color codes and are rendered based on their distance to the camera.

- Added `djui_hud_get/set/reset_text_color`. Text color refers to the default color the text uses outside of color codes. Color codes are not affected by this color, but are still affected by the HUD color (`djui_hud_get/set/reset_color`).

color-vs-textcolor

Line 1 is **HUD color** set to magenta. Notice that it also affects the word "color", which is normally yellow.
Line 2 is **HUD alpha** set to 50%. It affects all words, including the ones with color codes.
Line 3 is **Text color** set to magenta. None of the color codes are affected.
Line 4 is **Text alpha** set to 50%. None of the color codes are affected, they still are full opaque.

- Made HUD interpolations unlimited using a Growing Array.


src/pc/lua/smlua.c
```diff
@@ -407,8 +407,10 @@ void smlua_update(void) {
     // doing incremental garbage collection.
     // The real fix would be to make smlua produce less
     // garbage.
-    // lua_gc(L, LUA_GCSTOP, 0);
-    // lua_gc(L, LUA_GCCOLLECT, 0);
+    // EDIT²: smlua is so lazy during its GC that userdata is
+    // never garbage collected. Perform a GC cycle at the end
+    // of each frame to make sure it doesn't "forget" about it.
+    lua_gc(L, LUA_GCSTEP, 0);
```
djoslin0, line 413: Considering how many times we've been back and forth on GC stuff.. **_please,_** profile complicated mods before and after this change. Make sure that any tweak to the garbage collector doesn't tank performance

PeachyPeachSM64: this part is from https://github.com/coop-deluxe/sm64coopdx/pull/1145 and is out of the scope of this PR
but I understand your concern

PeachyPeachSM64:
will be rebased on dev without the temp pointer fix changes

PR #1150, opened 2026-03-15
Fix nametags color and rendering order

PeachyPeachSM64:
- Fixed nametags color and rendering order:

image

Nametags now support color codes and are rendered based on their distance to the camera.

- Added `djui_hud_get/set/reset_text_color`. Text color refers to the default color the text uses outside of color codes. Color codes are not affected by this color, but are still affected by the HUD color (`djui_hud_get/set/reset_color`).

image

Line 1 is **HUD color** set to magenta. Notice that it also affects the word "color", which is normally yellow.
Line 2 is **HUD alpha** set to 50%. It affects all words, including the ones with color codes.
Line 3 is **Text color** set to magenta. None of the color codes are affected.
Line 4 is **Text alpha** set to 50%. None of the color codes are affected, they still are full opaque.

- Made HUD interpolations unlimited using a Growing Array.
- Fixed a bug in regular text rendering not using the correct alpha value for color codes.

## Summary by CodeRabbit

* **New Features**
  * Added text color control for HUD elements with dedicated get/set/reset functions.

* **Bug Fixes**
  * Improved text rendering in HUD and crash handler with corrected color state handling.
  * Enhanced nametag rendering for better visual consistency.

* **Documentation**
  * Updated HUD color API documentation for clarity.
  * Removed obsolete function documentation.


PeachyPeachSM64:
@Isaac0-dev 
- moved `growing_array_init` in `djui_hud_clear_interp_data`
- `growing_array_init` can reuse array and buffer if already allocated

src/game/memory.c
```diff
@@ -185,10 +185,41 @@ void growing_pool_free_pool(struct GrowingPool *pool) {
...
-    array = calloc(1, sizeof(struct GrowingArray));
-    array->buffer = calloc(capacity, sizeof(void *));
+    growing_array_free_elements(array);
+
+    // reuse buffer if array was already allocated
+    if (array) {
+        if (!array->buffer || array->capacity != capacity) {
+            void **buffer = realloc(array->buffer, sizeof(void *) * capacity);
+
+            // if realloc fails, destroy the array and create a new one
```
Isaac0-dev, line 210: If the new capacity is smaller, there should almost never be a situation where the realloc fails but the standard alloc succeeds. (doesn't require any change, just observing)

src/game/memory.c
```diff
@@ -185,10 +185,41 @@ void growing_pool_free_pool(struct GrowingPool *pool) {
...
+                growing_array_free(&array);
+                return growing_array_init(NULL, capacity, alloc, free);
+            }
+
+            array->buffer = buffer;
+            array->capacity = capacity;
+            memset(array->buffer, 0, sizeof(void *) * array->capacity);
+        }
+    } else {
+        array = calloc(1, sizeof(struct GrowingArray));
```
Isaac0-dev: since every field of the growing array structure is set in this function, for performance this can safely be changed to:

```suggestion
        array = malloc(sizeof(struct GrowingArray));
```

src/game/memory.c
```diff
@@ -185,10 +185,41 @@ void growing_pool_free_pool(struct GrowingPool *pool) {
...
+
+            // if realloc fails, destroy the array and create a new one
+            if (!buffer) {
+                growing_array_free(&array);
+                return growing_array_init(NULL, capacity, alloc, free);
+            }
+
+            array->buffer = buffer;
+            array->capacity = capacity;
+            memset(array->buffer, 0, sizeof(void *) * array->capacity);
```
Isaac0-dev: `array->capacity` is set below too, so
```suggestion
            memset(buffer, 0, sizeof(void *) * capacity);
            array->buffer = buffer;
```

src/game/memory.c
```diff
@@ -185,10 +185,41 @@ void growing_pool_free_pool(struct GrowingPool *pool) {
...
+    if (array) {
+        for (u32 i = 0; i != array->capacity; ++i) {
+            if (array->buffer[i]) {
+                array->free(array->buffer[i]);
+                array->buffer[i] = NULL;
+            }
+        }
+        array->count = 0;
+    }
+}
```
Isaac0-dev, line 200: It's good practice that you've added this but, but both uses of this function (in `growing_array_init` and `growing_array_free`) the lines `array->buffer[i] = NULL;` and `array->count = 0;` go unused. Because `growing_array_init` clears the buffer, and `growing_array_free` frees it.
I guess at this point we're asking how fast we need the growing array to be.

Isaac0-dev (changes requested):
not a full review, just looked at the growing array stuff so far. I'll come back when I've got more time.

EmeraldLoc (approved):

EmeraldLoc:
I was not meant to approve this pr as I have not reviewed it, but GitHub is stupid and does not let me undo, rip

src/game/memory.c
PeachyPeachSM64: there is a case where array->buffer isn't cleared outside of this function, it's when the capacity doesn't change
besides, if the function is reused elsewhere, it avoids the caller having to manually clearing the buffer or setting the count to 0

src/pc/djui/djui_hud_utils.h
```diff
@@ -80,6 +82,12 @@ struct DjuiColor* djui_hud_get_color(void);
 void djui_hud_set_color(u8 r, u8 g, u8 b, u8 a);
 /* |description|Resets the current DJUI HUD color|descriptionEnd| */
 void djui_hud_reset_color(void);
+/* |description|Gets the current DJUI HUD text default color|descriptionEnd| */
+struct DjuiColor* djui_hud_get_text_color(void);
+/* |description|Sets the current DJUI HUD text default color|descriptionEnd| */
+void djui_hud_set_text_color(u8 r, u8 g, u8 b, u8 a);
+/* |description|Resets the current DJUI HUD text default color|descriptionEnd| */
+void djui_hud_reset_text_color(void);
```
Isaac0-dev: I think these should briefly explain what text colour actually means here. Something like "will be overridden by HUD color and color codes" should suffice.

Isaac0-dev (commented):
Did you check the ex-coop mod archive for any usages of `cur_obj_check_frame_prior_current_frame`?

PeachyPeachSM64:
> Did you check the ex-coop mod archive for any usages of `cur_obj_check_frame_prior_current_frame`?

yes, not a single use

src/pc/djui/djui_hud_utils.h
PeachyPeachSM64: I agree

Cooliokid956 (approved):
Please I need this

src/pc/nametags.c
```diff
@@ -96,61 +100,85 @@ void nametags_render(void) {
...
-
-            struct StateExtras* e = &sStateExtras[i];
-            if (!e->inited) {
-                vec3f_copy(e->prevPos, out);
-                e->prevScale = scale;
-                e->inited = true;
+            s32 j = 0;
+            for (; j < numNametags; ++j) {
+                if (scale < nametags[j].scale) {
+                    memmove(nametags + j + 1, nametags + j, sizeof(struct NametagInfo) * (numNametags - j));
```
Isaac0-dev: You don't need to do a `memmove` here, since you're copying a whole (empty) string during this.
You could do other things instead:
- An index map, which can be an array which you sort just the indices. Then, later on you just do `struct NametagInfo *nametag = &nametags[indices[k]];`. 
- Or, a similar solution where you have a pointer to every element in the array, and then you're just sorting the pointers instead of the whole memory block.
- `qsort`

Either way, I think it's a good idea to avoid the full `memmove`.

PeachyPeachSM64: done

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
     extern bool gDjuiHudToWorldCalcViewport;
     gDjuiHudToWorldCalcViewport = false;
 
-    // sort nametags by their distance to the camera
```
Isaac0-dev: this comment is necessary! this comment explains what exactly the following code is supposed to do. 
It's important to explain what the code is doing because without it someone may change the code without understanding it and that could lead to buffer overflows, read out of bounds, use after free, write out of bounds, memory leaks, double free, buffer underflow, simple off by one errors, etc.
So as you can see it's very important to include this comment in the code because it directly impacts the functionality of the code below.

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-    // insertion sort is quick enough for such small array
```
Isaac0-dev: yeah, good call. The englisH in this comment is completely outrageous. First of all, they don't even start with a capital `E`. Also the comment doesn't even say anything relevant about the surrounding code. The array in _this_ code is not small. The array is ~672 bits!

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-    for (s32 i = gNametagsSettings.showSelfTag ? 0 : 1; i < MAX_PLAYERS; i++) {
```
Isaac0-dev: probably not a bad idea too, considering it was blatantly programmed to do an off by one error every time `gNametagsSettings.showSelfTag` is true.

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-            case ACT_STOP_CROUCHING:
-            case ACT_START_CRAWLING:
-            case ACT_CRAWLING:
-            case ACT_STOP_CRAWLING:
-            case ACT_IN_CANNON:
-            case ACT_DISAPPEARED:
-            continue;
-        }
-
-        if (m->marioBodyState->mirrorMario || m->marioBodyState->updateHeadPosTime != gGlobalTimer) { continue; }
```
Isaac0-dev, line 80: yeah this code obviously didn't even do anything anyway, it just continued on to the next line so that's a pretty good optimisation if you ask me 👌

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-
-        Vec3f pos;
-        Vec3f out;
-        vec3f_copy(pos, m->marioBodyState->headPos);
```
Isaac0-dev, line 84: this one does nothing too because it doesn't read the return value. good change 👍

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-        pos[1] += 100;
```
Isaac0-dev, line 85: this just adds extra offset for no reason, so that's actually a write out of bounds error you just solved here just by optimising it

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-
```
Isaac0-dev, line 86: empty line, wasted space

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-        if ((i != 0 || (i == 0 && m->action != ACT_FIRST_PERSON)) &&
```
Isaac0-dev, line 89: actually this line is a good idea! it doesn't compile if you don't include it, so might want to add it back...

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-            djui_hud_world_pos_to_screen_pos(pos, out)) {
-
-            char name[MAX_CONFIG_STRING];
```
Isaac0-dev, line 92: who even wrote this? obviously you need to add an integer in between the braces.

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-            if (!djui_hud_world_pos_to_screen_pos(pos, out)) {
-                continue;
-            }
-
-            f32 scale = -300 / out[2] * djui_hud_get_fov_coeff();
-
-            s32 j = 0;
-            for (; j < numNametags; ++j) {
-                if (scale < nametags[j].scale) {
-                    memmove(nametags + j + 1, nametags + j, sizeof(struct NametagInfo) * (numNametags - j));
```
Isaac0-dev: oh good idea, this sorting method was inefficient, removing sorting altogether would be better.

src/pc/nametags.c
```diff
@@ -63,71 +63,26 @@ void nametags_render(void) {
...
-            }
-
-            nametags[j].playerIndex = i;
-            vec3f_copy(nametags[j].pos, out);
-            nametags[j].scale = scale;
-            memcpy(nametags[j].name, name, sizeof(name));
-            numNametags++;
-        }
-    }
+    // Ybbx ng gur pbzzvg qngr :)
```
Isaac0-dev: i got you! the commit date is 1st of April. anything specific you wanted to know that for?
Just saying, most people can't read rot13 natively, so it might come off as an obfuscated comment. it's just a little counter intuitive.

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+        char name[MAX_CONFIG_STRING];
+    };
+    struct NametagInfo nametags[MAX_PLAYERS] = {0};
+    s32 numNametags = 0;
+
+    extern bool gDjuiHudToWorldCalcViewport;
+    gDjuiHudToWorldCalcViewport = false;
+
+    // Ybbx ng gur pbzzvg qngr :)
+    #define A struct MarioState
```
Isaac0-dev: i appreciate your efficiency! this was already pretty obvious anyway so this will save us time later.

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define B struct NetworkPlayer
```
Isaac0-dev: same here, and your naming style is very elegant!

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define C struct NametagInfo
+    #define D Vec3f
+    #define f for (s32 i=1-gNametagsSettings.showSelfTag;i<MAX_PLAYERS;i++)
```
Isaac0-dev: i see your optimisation right there, very clean! I think you fixed the off by one error by one!

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define c continue
```
Isaac0-dev: perfection, gotta appreciate the attention to detail, with the macro name being the first letter of the word!

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define a case ACT_START_CROUCHING:case ACT_CROUCHING:case ACT_STOP_CROUCHING:case ACT_START_CRAWLING:case ACT_CRAWLING:case ACT_STOP_CRAWLING:case ACT_IN_CANNON:case ACT_DISAPPEARED
```
Isaac0-dev: that's a lot of optimisation!

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define b m->marioBodyState
```
Isaac0-dev: wow, you're really taking this seriously!

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define v(x) vec3f_##x
```
Isaac0-dev: don't you think it's a little bit excessive at this point?

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define dhwptsp djui_hud_world_pos_to_screen_pos
```
Isaac0-dev: i see what you did there 🥇

src/pc/nametags.c
```diff
@@ -59,98 +51,89 @@ void nametags_render(void) {
...
+    #define dhgfc djui_hud_get_fov_coeff
+    #define in i[nametags]
+    s32 n[MAX_PLAYERS]={0};f{A*m=&i[gMarioStates];B*q=&i[gNetworkPlayers];if(!is_player_active(m)+!q->currAreaSyncValid)c;switch(m->action){a:c;}if(b->mirrorMario||b->updateHeadPosTime-gGlobalTimer)c;D p,o;v(add)(v(set)(p,0,'d',0),b->headPos);if((i||m->action-ACT_FIRST_PERSON)*dhwptsp(p,o)){char g[MAX_CONFIG_STRING];const char*z=NULL;smlua_call_event_hooks(HOOK_ON_NAMETAGS_RENDER,i,p,&z);snprintf(g,MAX_CONFIG_STRING,"%s",z?z:q->name);if (!dhwptsp(p, o))c;f32 y=-'x'*'x'/'<'/'<'*'K'/2[o]*dhgfc();s32 j=0;for(;j<numNametags;++j){C*u=&(j[n])[nametags];if(y<u->scale){memmove(n+j+1,j+n,sizeof(s32)*(numNametags-j));break;}};j[n]=i;in.playerIndex=i;v(copy)(in.pos,o);in.scale=y;memcpy(in.name,g,sizeof(g));numNametags++;}}
```
Isaac0-dev: okay well... this could probably use a new line after each semi colon. you introduced a separate index array `n` to track ordering instead of directly sorting nametags. but why? the original one was obviously faster. it might have been moving an empty string along with the rest of the block, but was it really necessary? also, your backwards array accesses will work but aren't that intuitive to the average guy, just saying. How could one know that `i` is actually not an array, but `gMarioStates` is the array? it's kinda backwards. for most folks, it's instinctive to assume that the one on the left side of the braces is the array, it's not just an addition operation, _obviously_. also you missed a new line there. might be a good idea to put some spaces in between those operators, just on second thoughts. the `+` instead of `||` is a good way to show us you know your stuff in coding but is it really necessary? I mean, again most people would expect to see the `or` instead of a plus sign. you forgot a new line again. Also why did you bring back `if (m->marioBodyState->mirrorMario || m->marioBodyState->updateHeadPosTime != gGlobalTimer) { continue; }`? like I said, it doesn't do anything because `continue` just goes to the next piece of code. i think subtraction might not be as clean as just removing it tbh. that's a lot of circles for code! I'm thinking the character literal actually is a good idea but the +100 is actually a bug, so maybe just remove that and you're good 👌 . checking truthiness instead of `!= 0` is probably the cleanest idea here. but the math operators instead of `and` might be a far stretch. by the way putting text inside that braces won't compile. i think the character literals may be a little unreadable... the average guy has to lookup the character's decimal value. my estimation is that this is more just obfuscation rather than optimisation - just a personal opinion. by the way you missed a new line there too. and you also accidentally defined `j` outside the for loop. i see even more reverse indexing, and it's getting a little much, just saying. you missed another new line btw. that hm, the actual sorting is using memmove but this time on a single integer, so that's nice. now every read to the nametags array has to go through this array, where the algorithm is so scrambled it looks like code that accidentally works. by the way there's no way this improves binary size, it's doing almost the same thing. overall, this singular line just needs a few tweaks for perfection 👍

Isaac0-dev (commented):
honestly i'm a little on the fence about whether to merge this or not. on one hand it claims it "optimized performance, assembly size, file size and github bandwidth", which if true, is an unprecedented breakthrough in sm64coopdx development. but on the other hand I had to inspect this code with a microscope, a disassembler and at one point my therapist just to figure out that it does exactly the same thing as before except now we're memmoving 32 bits.

so i'm torn between theoretical marginal performance improvements, and code that looks like it was made for a competitive code golf tournament. 

normally i'd make this kind of decision myself, but due to the scale of these potential optimisations, i'm clearly not qualified to make this call without help from outside.
i typically rely on ai for all important life decisions to avoid my own human error, so with coop development's future at stake, i'll need to get some contribution to this painfully difficult choice to help put into perspective the general consensus on whether or not to go ahead. I'm imagining it will still be very close even with outside assistance, but at least it'll get us somewhere for now. 

@coderabbitai please review this pull request, specifically look at the performance improvements.

src/pc/nametags.c
EmeraldLoc: Agree with this, looks mostly good, just a few tweaks and this will be perfect!

Isaac0-dev:
@coderabbitai please review this pull request.

EmeraldLoc:
@coderabbitai help

Isaac0-dev:
@gemini-code-assist please review the commit a8ab75b and check for performance improvements

Isaac0-dev:
@gemini-code-assist please review the commit [a8ab75b](https://github.com/coop-deluxe/sm64coopdx/commit/a8ab75bb17a90b3e697ee5a1e7d7cd87becd446e) and check for performance improvements

Isaac0-dev:
@codex please review the commit [a8ab75b](https://github.com/coop-deluxe/sm64coopdx/commit/a8ab75bb17a90b3e697ee5a1e7d7cd87becd446e) and check for breaches of the laws of C

src/pc/nametags.c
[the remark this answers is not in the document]
EmeraldLoc: 👍Inspiring life changing thank you!

Isaac0-dev:
@gemini-code-assist let's get a distinction about performance and optimisation vs readability and maintainability in the commit [a8ab75b](https://github.com/coop-deluxe/sm64coopdx/commit/a8ab75bb17a90b3e697ee5a1e7d7cd87becd446e).
We need to understand specifically whether it's optimisation gains are of enough significance to justify the loss of readability. the commit name is specifically `optimized performance, assembly size, file size and github bandwidth`, so we need to verify this.

Isaac0-dev (approved):

AgentXLP:
What is with all of the AI usage

PeachyPeachSM64:
> What is with all of the AI usage

https://github.com/PeachyPeachSM64/sm64coopdx/commit/a8ab75bb17a90b3e697ee5a1e7d7cd87becd446e

AgentXLP:
Is this an April Fools prank or something
image
Wh at is this

AgentXLP:
Oh yeah it is

PR #1154, opened 2026-03-18
Fix custom character held objects and expose a new variable

wbmarioo:
This PR fixes bugs occuring when holding objects with custom characters. For example, Waluigi's longer arms drop the object at a different point, making mips clip and bobomb clip impossible/almost impossible. Now, characters all throw objects from the same location Mario would. The Lua API now has `get_character_holp_override` and `set_character_holp_override`, which can be used to disable this feature.

### DIFF
--- a/autogen/lua_definitions/functions.lua
+++ b/autogen/lua_definitions/functions.lua
@@ -11456,6 +11456,19 @@ function get_mario_anim_part_rot(m, animPart, rot)
     -- ...
 end
 
+--- @return boolean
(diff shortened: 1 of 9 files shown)


wbmarioo:
I also want to thank EmeraldLockdown and PeachyPeach for helping me fix this up and test a bit.

wbmarioo:
I have learned this is not accurate to vanilla (and there is no way to make it accurate) so i'll close this for now.

PR #1143, opened 2026-03-13
Refactor collision surface systems and add the ability to add/remove/move surfaces

djoslin0:
Refactor collision surface systems and add the ability to add/remove/move surfaces

- Add `smlua_collision_add_surface`, `_move_surface`, `_delete_surface` Lua APIs
- Add `remove_static_object_collision` to fully free a SOC and its surfaces
- Split surface pools into separate static, SOC, and dynamic pools
- Replace index-based SOC tracking with unique ID counter (`sSOCIdCounter`)
- Invalidate Lua CObjects for recycled surfaces on pool clear
- Expose `SURFACE_POOL_STATIC/DYNAMIC/SOC` constants to Lua
- Add `growing_array_swap_and_pop` and `growing_array_swap_and_pop_index` to memory utils

PR #1157, opened 2026-03-21
memory.c cleanup and optimisation

Isaac0-dev:
## Low hanging fruit optimisations for the memory data structures in `memory.c`
Mostly generic performance improvements:
- Swapping `calloc` calls for `malloc` calls where it's safe
- Combined allocations into one large allocation for `node + buffer`
- Using `realloc` instead of full `calloc` + `memcpy` + `free`.


PeachyPeachSM64:
> Revised growing_array_move

Oh, come on... Why did you have to bring that up again?
This has to be a joke...

> Currently without this PR, it uses the logic where to is relative to the array after collapsing the removed block.
This PR switches to the logic that to should be relative to the array before any operation is performed.
Because it's more intuitive. It's what a developer should be able to expect from an API perspective.
Developers shouldn't have to mentally simulate where to will be during the internal operation of the function.

That is absolutely WRONG. You reversed the logic AGAIN. Not once, not twice, but THREE times.
I even wrote an entire comment explaining why you're wrong (https://github.com/coop-deluxe/sm64coopdx/pull/924#discussion_r2495659839), and visibly you still haven't get it right.
It really baffles me that while you were right on the very first implementation of the function that the logic was not intuitive (which assumed `to` was the destination on the modified array) (see comment: https://github.com/coop-deluxe/sm64coopdx/pull/924#discussion_r2305487663), you claimed that the logic wasn't still right even after I modified it (commit: https://github.com/coop-deluxe/sm64coopdx/commit/9768e049f82ce0c4d36ffe08e2ee0676a6e25224, which happened **after** your comment), while proving you multiple times that the new logic used was clearly "`to` should be relative to the array before any operation is performed".

So, because it seems it's not clear enough, here we go again.

Let's take the array from your "Test logic": `[A, B, C, D, E, F]`
Let's use your "Move to End" example: `from = 0, to = 4, count = 2`
You claim that: **`to` should be relative to the array before any operation is performed.**

```
Let's index each element:

[ A, B, C, D, E, F ]
  ^  ^  ^  ^  ^  ^
  0  1  2  3  4  5

'to' has the value: 4. So, where is 'to'?
It should refer to the slot before any operation is performed: it's the slot that points to E.

[ A, B, C, D, E, F ]
              ^   
              4   

It means the moved chunck will be inserted right before E.
Which gives the following result:

[ A, B, C, D, E, F ]
  \___/      ^
    |________|

=>

[ C, D, A, B, E, F ]

Now, let's compare with your "expected" result: [ C, D, E, F, A, B ]
It's clearly not the same.
```
**`to` should be relative to the array before any operation is performed**, but your code doesn't seem to follow that logic, while the existing code **does**, and always did since this commit: https://github.com/coop-deluxe/sm64coopdx/commit/9768e049f82ce0c4d36ffe08e2ee0676a6e25224.

> This version also has a new optimisation which uses a stack buffer for smaller, faster operations but uses dynamic memory when dealing with larger blocks.

This benchmark is not completely relevant. You gain a 80% speed boost **only** when the count is lower than the stack allocated buffer size. For all the other cases, it's actually slightly worse, because of the branch you introduce with the `if` statement.

Currently, this function is only used for SOC surfaces and surface nodes. Which has barely any usage. So this optimization trick doesn't even matter.
Even worse, with https://github.com/coop-deluxe/sm64coopdx/pull/1143, this function would not be used at all and would likely be removed.

---

I'm tired of arguing over the most pointless stuff and explaining it again and again.
Once again, I spent so much time writing this, hoping you will finally understand.
Please know that I don't and won't approve this. You can decide to ignore this message and merge it anyway, I don't care.
I won't care anymore.

djoslin0:
Peachy is correct here, and the step-by-step process she showed makes the issue very clear. This shouldn't be merged.

Isaac0-dev:
I apologise for causing so much frustration here, that wasn't my intention.
I now understand that the main confusion was the semantics of `to`. I thought `to` should be a destination index, while the current implementation treats it more like an "anchor" to existing data. Since the "anchor" logic is already being used in the codebase, I agree that it shouldn't be changed.

I'll remove my `growing_array_move` changes from this PR, and keep the memory optimisations.

Again, sorry for dragging this out so much, I didn't mean to waste people's time.

sdkxhyp49q-coder (approved):

EmeraldLoc (approved):

src/game/memory.c
```diff
@@ -200,11 +213,11 @@ void *growing_array_alloc(struct GrowingArray *array, u32 size) {
     if (array && array->buffer) {
 
         // Increase capacity if needed
-        while (array->count >= array->capacity) {
+        if (array->count >= array->capacity) {
```
PeachyPeachSM64: I don't fully agree with this
while it's normally unachievable, if you have `array->count` > `array->capacity * 2`, it will cause a buffer overflow
I admit the initial implementation was lazy, but I think the newCapacity should take the array->count into account

PeachyPeachSM64 (changes requested):

src/game/memory.c
Isaac0-dev: 🤷  i guess it can't hurt


```suggestion
        while (array->count >= array->capacity) {
```

PeachyPeachSM64 (approved):

PR #1165, opened 2026-03-27
Add "do not show again" checkbox to host message panel

mingokrb:
Adds a checkbox to `djui_panel_host_message` to skip the firewall/port forward warning and jump straight to the room, regardless of its max number of players, everytime you host using Direct Connection

hide_socket_warning

Does not apply until you click **Host** while it's checked

Creates a new boolean entry in `sm64config.txt` called `hide_socket_warning` to store its value, which is the only way of reverting it. Please lmk if I should add an option to toggle it back off and in which category (prob 'Misc')

Needs translation for all languages but Portuguese (and English ofc):
```ini
WARN_SOCKET_HIDE = "OK, do not show this again"
```


frog8412:
```ini
WARN_SOCKET_HIDE = "ОК, больше не показывать"
```

Manolommm:
Spanish:
```ini
WARN_SOCKET_HIDE = "OK, no volver a mostrar esto"
```

EmeraldLoc (approved):
Looks good! I think it could use a bit less spacing, but that's not your fault, that value is hardcoded right now.

PR #1172, opened 2026-04-01
Fix issue with get mod files

EmeraldLoc:
<https://github.com/coop-deluxe/sm64coopdx/commit/8735cf96adf5c876f953cb911793403afdeaa4d7> caused an issue for windows. Since it had a hardcoded check for the `/` character, it cause windows to not work. This was fixed by making it change depending on which platform you are on.

Tested on windows and macOS


EmeraldLoc:
https://github.com/coop-deluxe/sm64coopdx/commit/ba54cbd1d0b8ec9474f5cfd3bebae758f69c2e2b fixes it, thanks peachy!

PR #1180, opened 2026-04-06
Improve readability of behavior files.

Bloxxel64:
Improve the readability of many of the vanilla behaviors, in order to ease the creation of custom/modified ones.

Changes include:

- Giving variables proper names (e.g. `distanceFromHome` instead of `sp20`).
- Giving act functions proper names (e.g. `whomp_act_7_patrol_pivot_to_home` instead of just `whomp_act_7`)
- Documenting extra things such as the functionality of `oBehParams2ndByte`.
- Generally adding comments to files, to insure they all have similar levels of documentation.


PeachyPeachSM64:
Hi, while I agree we should get rid of these UnkXX object fields, you can't remove them completely, mods that use them will break

AgentXLP:
Stale and Emerald is already basically doing this as well as fixing syncing bugs too

Bloxxel64:
That's fair, I figured this would get knocked out once Emerald started his work, if there's any glaring issues I find when emerald is done I can always make a PR then

PR #1181, opened 2026-04-06
Fix `get_texture_info` and add `HOOK_ON_DYNOS_PACKS_TOGGLED`

kermeow:
Previously `get_texture_info` did not respect DynOS overrides as getting the info of built-in textures has special behaviour. Now overrides are checked, however the `TextureInfo` returned will still include the original texture pointer as that is what's expected for overridden textures by the rest of the engine.

Additionally, a new hook `HOOK_ON_DYNOS_CHANGED` is added to allow mod developers to detect when a pack is enabled/disabled and reload any texture info accordingly.


Isaac0-dev:
dynos is a very broad name, I suggest making the hook name `HOOK_ON_DYNOS_PACKS_CHANGED`

kermeow:
I went with `HOOK_ON_DYNOS_PACKS_TOGGLED` just to be a little more specific

AgentXLP (approved):
Looks good to me, but I'm going to wait until Peachy takes a look in case of any suggestions she might have.

data/dynos_mgr_tex.cpp
```diff
@@ -501,6 +501,18 @@ bool DynOS_Tex_Get(const char* aTexName, struct TextureInfo* aOutTexInfo) {
             }
         }
         return false;
+    } else {
+        // check override
+        auto &_DynosOverrideTextures = DynosOverrideTextures();
+        const Texture *_BuiltinTex = info->texture;
+        if (_DynosOverrideTextures[_BuiltinTex]) {
+            auto *_Node = _DynosOverrideTextures[_BuiltinTex]->node;
```
Isaac0-dev: that's a double lookup, do it once:

```suggestion
        auto it = _DynosOverrideTextures.find(_BuiltinTex);
        if (it != _DynosOverrideTextures.end() && it->second) {
            auto *_Node = it->second->node;
```

src/pc/lua/smlua_hook_events.inl
```diff
@@ -64,3 +64,4 @@ SMLUA_EVENT_HOOK(HOOK_ON_FIND_FLOOR, _, f32 posX, f32 posY, f32 posZ, struct Sur
 SMLUA_EVENT_HOOK(HOOK_ON_FIND_WATER_LEVEL, _, f32 x, f32 z, f32 *waterLevel) // Manually defined hook
 SMLUA_EVENT_HOOK(HOOK_ON_FIND_POISON_GAS_LEVEL, _, f32 x, f32 z, f32 *gasLevel) // Manually defined hook
 SMLUA_EVENT_HOOK(HOOK_ON_FIND_SURFACE_ON_RAY, _, Vec3f orig, Vec3f dir, struct Surface **hit_surface, Vec3f hit_pos) // Manually defined hook
+SMLUA_EVENT_HOOK(HOOK_ON_DYNOS_PACK_TOGGLED, HOOK_RETURN_NEVER)
```
Isaac0-dev: fix the new line at the end:
```suggestion
SMLUA_EVENT_HOOK(HOOK_ON_DYNOS_PACK_TOGGLED, HOOK_RETURN_NEVER)

```

Isaac0-dev (changes requested):
these are nitpicks.
they should be easy to merge using github's suggestion feature

data/dynos_mgr_tex.cpp
```diff
@@ -501,6 +501,19 @@ bool DynOS_Tex_Get(const char* aTexName, struct TextureInfo* aOutTexInfo) {
...
+        auto it = _DynosOverrideTextures.find(_BuiltinTex);
+        if (it != _DynosOverrideTextures.end() && it->second) {
+            auto *_Node = it->second->node;
+            auto &_Data = _Node->mData;
+            CONVERT_TEXINFO(aTexName);
+            // rendering errors if we dont link back to the original because it cant find the node
+            aOutTexInfo->texture = info->texture;
+            return true;
+        }
     }
```
Isaac0-dev: sorry for more nitpicks but I think these are kinda important.
I looked into why those rendering errors you mentioned happen.
Also don't need the else block.

```suggestion
    }

    // check override
    auto &_DynosOverrideTextures = DynosOverrideTextures();
    auto it = _DynosOverrideTextures.find(info->texture);
    if (it != _DynosOverrideTextures.end() && it->second) {
        auto *_Node = it->second->node;
        auto &_Data = _Node->mData;
        CONVERT_TEXINFO(aTexName);

        // must link to the vanilla built in texture
        // so that the texture is correctly overridden
        // when DynOS_Tex_RetrieveNode is called.
        aOutTexInfo->texture = info->texture;
        return true;
    }

```

Isaac0-dev:
given the need to link to the original texture, I think it's a considerable limitation since Lua mods can usually use `texture_to_lua_table`, but this will just give them an old texture

kermeow:
> Lua mods can usually use `texture_to_lua_table`, but this will just give them an old texture

It looks like `texture_to_lua_table` ultimately calls `DynOS_Tex_RetrieveNode` (via `dynos_texture_get_from_data` -> `DynOS_Tex_GetFromData`) so the override applies and the new texture should be returned

Isaac0-dev:
ah, good point you're right!

Isaac0-dev (approved):

PeachyPeachSM64:
wouldn't it be possible to pass the pack name and state (enabled/disabled) as hook parameters?

kermeow:
i don't see why not
i'll get on that in a min

PR #1182, opened 2026-04-06
Add independent X and Y scaling to text rendering functions

Squishy6094:
This PR modifies both `djui_hud_print_text` and `djui_hud_print_text_interpolated` to take inputs for both X and Y scale rather than just scale. This was intentionally made to be fully backwards compatible with the existing method of using these functions, and thus have been moved to their respective manual documentation files.

Here is an Example Video showing this in action:

https://github.com/user-attachments/assets/db753d67-9506-459a-85d6-1d7ed8657271

And the code rendering it:
```lua
local function hud_render()
    djui_hud_set_resolution(RESOLUTION_N64)
    local scaleX = math.sin((get_global_timer()) * 0.1)/math.pi + 0.5
    local scaleY = math.sin((get_global_timer() + 5) * 0.1)/math.pi + 0.5
    local prevScaleX = math.sin((get_global_timer() - 1) * 0.1)/math.pi + 0.5
    local prevScaleY = math.sin((get_global_timer() + 4) * 0.1)/math.pi + 0.5
    
    -- Original Function Format
    djui_hud_print_text("LEGACY TEST", 50, 50, scaleX)
    -- Added Y Scale Format
    djui_hud_print_text("NEW TEST", 50, 100, scaleX, scaleY)

    -- Original Interp Function Format
    djui_hud_print_text_interpolated("LEGACY INTERP", 150, 50, prevScaleX, 150, 50, scaleX)
    -- Added Y Scale Interp Format
    djui_hud_print_text_interpolated("NEW INTERP", 150, 100, prevScaleX, prevScaleY, 150, 100, scaleX, scaleY)
end

hook_event(HOOK_ON_HUD_RENDER, hud_render)
```

### DIFF
--- a/src/pc/lua/smlua_functions.c
+++ b/src/pc/lua/smlua_functions.c
@@ -1098,6 +1098,7 @@ void smlua_bind_functions(void) {
     smlua_bind_function(L, "collision_find_surface_on_ray", smlua_func_collision_find_surface_on_ray);
     smlua_bind_function(L, "cast_graph_node", smlua_func_cast_graph_node);
     smlua_bind_function(L, "get_uncolored_string", smlua_func_get_uncolored_string);
+    smlua_bind_function(L, "gfx_set_command", smlua_func_gfx_set_command);
     smlua_bind_function(L, "djui_hud_print_text", smlua_func_djui_hud_print_text);
     smlua_bind_function(L, "djui_hud_print_text_interpolated", smlua_func_djui_hud_print_text_interpolated);
 }
--- a/src/pc/lua/smlua_functions.c
+++ b/src/pc/lua/smlua_functions.c
@@ -1005,6 +1005,69 @@ int smlua_func_gfx_set_command(lua_State* L) {
     return 1;
 }
 
+  /////////
+ // hud //
+/////////
+
+
+int smlua_func_djui_hud_print_text(lua_State* L) {
+    if (L == NULL) { return 0; }
+
+    int top = lua_gettop(L);
+    if (top != 4 && top != 5) {
+        LOG_LUA_LINE("Improper param count for '%s': Expected %u, Received %u", "djui_hud_print_text", 4, top);
+        return 0;
+    }
+
+    const char* message = smlua_to_string(L, 1);
+    if (!gSmLuaConvertSuccess) { LOG_LUA("Failed to convert parameter %u for function '%s'", 1, "djui_hud_print_text"); return 0; }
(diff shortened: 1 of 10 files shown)


xLuigiGamerx:
We kinda went backwards with this one, djui hud functions used to be in manual.lua then we moved it to functions.lua using autogen, it should stay there, you should modify autogen instead to add a keyword to treat fields as optional instead of moving it back to manual.lua

Squishy6094:
I went with this method mainly because it makes the code way less messy then trying to fit backwards compatibility in-function, If it is super important that it stays autogen'd then I can rewrite things but otherwise I feel this works just fine.

Cooliokid956:
the true solution is to add overloading to autogen

PeachyPeachSM64:
I like the idea of overloads
- it makes easier to add features to existing functions without the need to create an entirely new function or risking to break compatibility
- modders wouldn't have to deal with the same function existing multiple times but with different names depending on the number of arguments

AgentXLP:
Yeah if you can make this use overloads its a pretty good addition that can be merged (on topic comment)

Squishy6094:
Implementing overloads into autogen is outside the scope of this PR. This PR achieves the goal that it aims to, and can be later updated in the possible overloads PR. There are other functions that already use the method I've implemented, thus this is the best fitting solution for now.

src/pc/lua/smlua_functions.c
```diff
@@ -1035,5 +1098,6 @@ void smlua_bind_functions(void) {
     smlua_bind_function(L, "collision_find_surface_on_ray", smlua_func_collision_find_surface_on_ray);
     smlua_bind_function(L, "cast_graph_node", smlua_func_cast_graph_node);
     smlua_bind_function(L, "get_uncolored_string", smlua_func_get_uncolored_string);
-    smlua_bind_function(L, "gfx_set_command", smlua_func_gfx_set_command);
```
Isaac0-dev, line 1038: was this an accident?

Squishy6094: Yeah this was a mistake, I'm new to PRs and especially new to autogen, is this the only thing that need to be restored or is there anything else?

AgentXLP:
> Implementing overloads into autogen is outside the scope of this PR. This PR achieves the goal that it aims to, and can be later updated in the possible overloads PR. There are other functions that already use the method I've implemented, thus this is the best fitting solution for now.

Fair enough, I decided I'll merge this as is since it is useful either way

PR #1183, opened 2026-04-07
Add color interpolation and combiner editing to DJUI 

Cooliokid956:
## `djui_hud_set_color_interpolated`
Takes twice as many arguments as `djui_hud_set_color`
u8 prevR, u8 prevG, u8 prevB, u8 prevA, u8 r, u8 g, u8 b, u8 a

## `djui_hud_set_combiner_cycles`
Sets the amount of cycles used in rendering

## `djui_hud_set_combiner`
```
Sets the current DJUI HUD combiner.
Each part uses the following equation: `P = (A - B) * C + D`.
Cycle 2 may be used to extend the equation, with the result of the previous cycle accessible through CS_COMBINED

void djui_hud_set_combiner(u8 cycle, bool alpha, OPTIONAL enum CombinerSource a, OPTIONAL enum CombinerSource b, OPTIONAL enum CombinerSource c, OPTIONAL enum CombinerSource d);
```
## `djui_hud_reset_combiner`
Resets the above, reverting combine modes back to the defaults for each element


AgentXLP:
Fascinating

Blockyyy:
i'm assuming this has the same combiner issues as everything else

AgentXLP:
It would inherit them yes

Cooliokid956:
I will close this because the transforms PR is very dependent on it, so I'll just ship all three features together as originally intended

PR #1188, opened 2026-04-09
Separate fixed smoke from cafe pipe and added star overwrite dynos

SprSn64:
collect my shine sprites
star dynos is a better dynos example than a pipe thats only seen in the intro and a single level


AgentXLP:
collect my 8 pronouns

Bloxxel64:
I'd say it would be better to put any model replacement stuff intended only for examples in a documentation folder rather than it being part of DX for no real good reason

Cooliokid956:
you should also probably include the blend file to provide as a template, like the one I released with my Carl's Jr Chicken Stars dynos pack which I designed to export to a geolayout matching that of the original star
maybe make a new repo that acts like some sort of ..Models Resource?

PeachyPeachSM64:
Wrong branch. Source should be based on dev and target should be dev.
Closing it.

PR #1191, opened 2026-04-09
example star dynos attempt 2 github please put this on the dev branch

SprSn64:


Bloxxel64:
This isn't really needed, if you want to add the shine sprite as a star replacement just release it as a dynos pack in the coop server/on the forums .

SprSn64:
its meant to be an example dynos for people to figure out what they do or how to make their own. the coop cafe pipe thing isnt really a good example as it only changes a model thats only seen a few times in the entire game

Blockyyy:
if you really wanted to make an example DynOS pack i think you should include the blend file along with all the C files and put all of that next to where the example mods are, none of the example mods are built in and i don't think this should be as well (this also would go for the coop cafe pipe imo)
The fact that a model isn't seen often in the game shouldn't be a concern as long as it is a good example for others to follow it is fine, regardless of it being a star or a pipe, this isn't really a better example pack because of that (you are missing everything that would make this an example other than the bins)

paradox9955:
There's no real reason to have examples built into the base game, honestly. If people want to learn how to do something, we have thousands of mods available that can give them some insight on how to do something similar. Also, adding more built in options isn't something the devs want to do. In fact, I believe they'd prefer to remove them instead.

PR #1192, opened 2026-04-10
Introduce this new pack in a more proper reviewable manner

AgentXLP:
Ideally I should have made a PR for my new pack and I understand the new geo function may have come out of no where. I've been thinking about adding it for a while and was going to utilize it in a new project I'm working on but I should have explained it first before pushing anything for sure

* Re-adds a recreation of the Coop Cafe Pipe, recently removed by Squishy which is fine
* Adds new geo function `geo_switch_character_type_ext` which can be used in switch cases to set the case to the character index of the local player. In the pack, it's used to switch out the emblem on the pipe. I'm thinking of also making the parameter able to tell the function which local mario index should be referenced for the character. Open to feedback on how to improve this function


Cooliokid956:
for the dynamic part of it, I'd say that it would probably be better to add a system that allows dynos packs to use Lua functions, so that way a mod could be made that activates the pipe's character functionality, but the pack can work just fine without
other uses for this technology would be playermodel packs with long hair that work fine without but would benefit from some hair simulation mod, with the mod providing the geo functions needed

```lua
function geo_func_xyz()
  -- ...
end

dynos_allow_geo_func("geo_func_xyz")
```

PeachyPeachSM64:
> for the dynamic part of it, I'd say that it would probably be better to add a system that allows dynos packs to use Lua functions, so that way a mod could be made that activates the pipe's character functionality, but the pack can work just fine without other uses for this technology would be playermodel packs with long hair that work fine without but would benefit from some hair simulation mod, with the mod providing the geo functions needed
> 
> ```lua
> function geo_func_xyz()
>   -- ...
> end
> 
> dynos_allow_geo_func("geo_func_xyz")
> ```

That already exists with the lua `geo_asm` feature. DynOS packs can embed Lua references, and that's a problem.
They should not be able to run Lua code at all, since those packs are local player only.

That being said, a geo switch function depending on the player's character is interesting and can open to more customization regarding player models.
However, it is barely usable in the current state.

`gMarioState` always refer to the local Mario. This limits the capabilities of this function.
I would suggest the use of `geo_get_mario_state` to retrieve the MarioState associated to the current processed object. For non-Mario objects, it returns the MarioState from their `globalPlayerIndex`. Doing this change would allow this function to be used in player models as well.

Small nitpick: I would call the function `geo_switch_character_type` or `geo_switch_player_character_type`. No need for the `ext`.

Now, here is my opinion about built-in DynOS packs.
I don't think they give much value to the base game and should be included. Like the example mods, there could be an example dynos packs section with these models, including C files and possibly blends.
I don't mind keeping the Coop Café pipe pack built-in, but please separate it from the smoke texture fix. They should be two different packs.

AgentXLP:
Alright I changed some things

AgentXLP:
I think this is ready to go

PR #1193, opened 2026-04-10
Control the Lighting Engine with le_set_enabled in real time

Emeraldsniper:
This is my very first PR ever so if I mess anything up, that's on myself.. (I had to redo the entire fork because I hate git desktop and I honestly hate myself.)

This PR is to create two new functions to control the lighting engine in real time: le_disable and le_enable

The lighting engine in coop was made to only work without turning it off at all. It stays on until the host disconnects the server. Modders had to use a really weird ambient settings in order to lower the lighting engine's properties when using it, but that didn't solve the issue entirely either. So I took the dive to remedy this myself, as I wanted to contribute to coop in some form. (This was WAY too long of a dive. I am really terrible at coding and I learn slower than anyone else...)

Below is a vod preview of the functions working in bitdw.

https://github.com/user-attachments/assets/88b6bf7d-bfc7-4a87-926a-b86bedffe6ef

I'm aware there's some description errors and I forgot to literally fix those again, but I will update this commit with fixing them anyway.

### DIFF
--- a/autogen/lua_definitions/functions.lua
+++ b/autogen/lua_definitions/functions.lua
@@ -5155,13 +5155,9 @@ function le_set_max_lights_per_vertex(count)
     -- ...
 end
 
---- Tells the system to disable the lighting engine when it's currently enabled.
-function le_disable()
-    -- ...
-end
-
---- Tells the system to enable the lighting engine when it's currently disabled.
-function le_enable()
+--- @param value boolean
+--- This will let the user control the lighting engine in real time to disable or enable it.
+function le_set_active(value)
     -- ...
 end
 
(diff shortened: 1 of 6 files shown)


Blockyyy:
can't this be done in one function with a bool passed as a parameter?

Emeraldsniper:
> can't this be done in one function with a bool passed as a parameter?

Actually yes and I took my time to look into it with some helping hands.. now it works with the new function: le_set_active.

Changing the lua file to use:
le_set_active(false) and le_set_active(true)

And it yields the same results as I wanted.

https://github.com/user-attachments/assets/ba5c9eac-fce8-407f-b47c-bdb7dfaafcaf

PeachyPeachSM64:
I think calling it `le_set_enabled` or `le_enable` would be better, to match `le_is_enabled`

Emeraldsniper:
> I think calling it `le_set_enabled` or `le_enable` would be better, to match `le_is_enabled`

Honestly, fair.. I changed it to le_set_enabled. (Before I go insane from the convos)

Lua file will need le_set_enabled(false) and le_set_enabled(true) in conjunction to the change.

PR #1197, opened 2026-04-16
Fix audio support on 32 bit builds

sashaantipov2012:


Isaac0-dev (commented):
Does this really fix audio on 32 bit builds? Even on a completely clean build?
I can see that some of these changes may help, but I'm confused by the changes to the sequence offset code.

The whole compression feature is just disabled for 32 bit builds.
But there's a problem with that too. This requires the vanilla SM64 ROM assets to generate those offsets, but you haven't enabled ROM asset extraction for 32 bits either. 

I don't have a 32 bit machine to test these changes so some more explanation on how you came across these changes would be appreciated.

sashaantipov2012:
Weird, worked on the latest dev branch for me...

Isaac0-dev:
> Weird, worked on the latest dev branch for me...

I was wondering if you could please explain the changes you made here?

AgentXLP:
What if it's AI

Isaac0-dev:
> What if it's AI

Can't rule out that potential, but it's most fair to assume we are talking to a real human here until we have evidence to prove otherwise.

@sashaantipov2012 please provide more context on this PR.

sashaantipov2012:
You are absolutely correct! As an AI language model...... Okay, jokes aside I did get some help from an AI but still checked the changes it recommended me to make and made sure everything works correctly

sashaantipov2012:
For a moment I thought that I accidentally changed it to a draft but now I just realized that you did that

Isaac0-dev:
> For a moment I thought that I accidentally changed it to a draft but now I just realized that you did that

@sashaantipov2012 yes I did that, and it wasn't a mistake. Please explain this PR and the changes made here. If you need to use your AI assistant for that, please do. I just need some explanation on what happened here before I can properly review.

sashaantipov2012:
Oh, my bad. I thought that it was kinda obvious but basically I found the commit that changed the audio support to the new system (which no longer requires a rom) and reverted it. Then I just merged it with the new system so if the architecture is 32 bit, use the old audio system. If its any other architecture, use the new one

Isaac0-dev:
@sashaantipov2012 okay, thanks for that. I'm a bit unsure about whether or not this now needs you to provide a rom when you compile. Currently, we don't require a rom to build the game, but this approach seems to need a rom. But there's a problem. You have to have already extracted the assets from the rom in order to make this code work.
It doesn't look like you've included the code to extract the assets again, so `assemble_sound.py` won't even be executed even with your changes at build time. Due to this, I struggle to believe that this will work on a completely clean build for 32 bit devices.
Can you confirm if this works on a _clean_ build? As in, no source changes besides the ones in your PR, and after running `make distclean`.

EmeraldLoc:
Stale, and unsure if the changes here actually fix anything at all. Feel free to reopen when you have an explanation and/or fixes

PR #1068, opened 2025-12-31
CObject Properties

Cooliokid956:
- Added property members to CObjects via `PROPERTY` macro `PROPERTY(name, get, set)`
  - `name`: property name
  - `get`: `fun(self): value`
  - `set`: `fun(self, value)`
```c
struct SomeObject {
    ...

    PROPERTY(field, get_field, set_field);
}
```

- Prettified `LuaObjectField` struct with unions for function/property value types
- Added properties to `struct ModAudio`
  - `position`: `number`
  - `looping`: `boolean`
  - `frequency`: `number`
  - `volume`: `number`
    - NOTE: These only work with streams—the audio reform will come later

PR #1200, opened 2026-04-19
Fix file.relativePath compatibility issue with sound mods

Baconator2558:
1.4 reworked the `ModAudio` struct to no longer have a `ModFile` field. However, this created compatibility issues with mods that used the `ModFile` field's `relativePath` to find the audio's path, breaking most voice mods. A new struct that stores a copy of the `ModAudio`'s `filepath` field is what I went with to address this. The added struct and field are hidden from documentation.


src/pc/lua/utils/smlua_audio_utils.c
```diff
@@ -301,6 +301,8 @@ struct ModAudio* audio_load_internal(const char* filename, bool isStream) {
 
     // remember file
     audio->filepath = strdup(filepath);
+    struct ModAudioLegacyPath* fileLegacy = { strdup(filepath) };
```
PeachyPeachSM64: That's not how you do it
First, you need to malloc a pointer of ModAudioLegacyPath
Then, you strncpy filepath into fileLegacy->relativePath
Finally, you need to free audio->file when audio is freed

PeachyPeachSM64 (changes requested):
Honestly I'm not sure we want to add a new structure and field to an existing one, just to fix something modders should have updated in their mods...

Cooliokid956:
PLEASEEE review my pr #1068
you could then make `file` a property, with a getter that spits out that same `ModAudio`, then make a `union` to give `filePath` a `relativePath` alias

PeachyPeachSM64:
> PLEASEEE review my pr #1068 you could then make `file` a property, with a getter that spits out that same `ModAudio`, then make a `union` to give `filePath` a `relativePath` alias

merge conflicts

but that seems to be indeed the best solution

src/pc/lua/utils/smlua_audio_utils.c
Baconator2558: Should be correct now. Even if we end up going for Coolio's idea, I figured I might as well fix this too.

Cooliokid956:
> merge conflicts

fixed 🙏🙏🙏🙏🙏

AgentXLP:
Coolio is gonna work on a different implementation of this (file property w/ aliases) that does this but with his new system

PR #1202, opened 2026-04-21
fix lingering kick pvp interaction

Blockyyy:
kicks now check for the MARIO_KICKING flag, meaning a kick will only hit another player for the first 8 frames

first video is before last is after

https://github.com/user-attachments/assets/7e8b765c-82d3-4f81-9007-8ed068af0187

https://github.com/user-attachments/assets/befbfe65-e731-4837-a211-5f784684a94b


wbmarioo:
thank you

PR #1203, opened 2026-04-22
Add HiDPI display support for SDL2

LeoManrique:
Adds HiDPI display support so the game renders at native resolution on high-density screens (Retina Macs, 4K/scaled Windows laptops, Linux with fractional scaling) instead of being upscaled by the OS, making some elements (especially menus and game characters) look a bit pixalated.

- [x] Sharp rendering on HiDPI displays
- [x] Mouse input still lines up with window geometry
- [x] No change on regular (non-HiDPI) displays

The differences are not so noticeable on medium or big size screens like laptops or monitors, but when developing the iOS port the difference was huge.

**Tests on Macbook Pro M5:**

Before:
image (game logo borders, castle towers)
image (stars borders)
image (mario's hair)

After:
image

image

image

**Tests on iOS port:**

Before:
image (game logo borders, castle towers)

image (mario's hair, mario's body)

After:
image

image

### DIFF
--- a/src/pc/configfile.c
+++ b/src/pc/configfile.c
@@ -77,6 +77,7 @@ ConfigWindow configWindow = {
     .exiting_fullscreen = false,
     .settings_changed = false,
     .msaa = 0,
+    .hidpi = false,
 };
 
 ConfigStick configStick = { 0 };
@@ -227,6 +228,7 @@ static const struct ConfigOption options[] = {
     {.name = "window_h",                       .type = CONFIG_TYPE_UINT, .uintValue = &configWindow.h},
     {.name = "vsync",                          .type = CONFIG_TYPE_BOOL, .boolValue = &configWindow.vsync},
     {.name = "msaa",                           .type = CONFIG_TYPE_UINT, .uintValue = &configWindow.msaa},
+    {.name = "hidpi",                          .type = CONFIG_TYPE_BOOL, .boolValue = &configWindow.hidpi},
     // display settings
     {.name = "graphics_backend",               .type = CONFIG_TYPE_UINT, .uintValue = &configGraphicsBackend},
     {.name = "texture_filtering",              .type = CONFIG_TYPE_UINT, .uintValue = &configFiltering},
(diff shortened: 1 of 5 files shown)


EmeraldLoc:
Cool, but a few issues with this:

1. This completely jacks uncapped fps interpolation. While the game claims I'm running at 500+ fps, it instead runs at slideshow speeds, it feels slower than no interpolation.
2. Due to the performance hit this takes, while most Macs should run it fine, others may not want it on all the time due to performance reasons (personally, my fps drops from 1000+ to 500 for little to no gain in clarity). This should be toggleable
3. The UI is very small. Go onto any normal display, it doesn't have to be retina, and you quickly notice the hud is big everywhere by default. This is likely due to djui scale needing a bigger setting, and auto needing to go up to it appropriately

LeoManrique:
@EmeraldLoc I am considering adding a configuration value to toggle this. Only at config file level and not in the UI. Lmk what do you think about it :)

EmeraldLoc:
I'd recommend making it an option in the display settings, just disable changing it if your display doesn't support HiDPI. I don't think there's a reason to hide it in the config file.

LeoManrique:
@EmeraldLoc I added the flag only for devices with this feature available.

Testing on Mac Mini:

https://github.com/user-attachments/assets/f9a6ea37-8ba7-42a1-af63-0e37a601344e

Testing on non HiDPI device:

image

PeachyPeachSM64:
conflicts

LeoManrique:
@PeachyPeachSM64
fixed conflicts :)

Testing again on Macbook Pro:

image
image


Testing again on non-HiDPI device:

image

EmeraldLoc:
This still has the issue of DJUI being too small and uncapped FPS running at slideshow speeds

LeoManrique:
@EmeraldLoc may I ask which device are you using and what do you mean by slideshow speeds, like under 30FPS?
I unfortunately haven't been able to replicate it yet.

Cheers!

EmeraldLoc:
Under 30fps, M1 Pro MacBook Pro macOS Tahoe 26.2 iirc, can’t check rn

PR #1205, opened 2026-04-23
Add `audio_stream_get/set_volume_channel`

kermeow:
Allows modders to play audio streams on channels other than level background music.
4 constants have been added for this purpose:
- `MOD_AUDIO_CHANNEL_MASTER` - sound is only affected by master volume
- `MOD_AUDIO_CHANNEL_MUSIC` - sound is affected by music volume, same as previous behaviour
- `MOD_AUDIO_CHANNEL_SFX` - sound is affected by sfx volume, same as sample behaviour
- `MOD_AUDIO_CHANNEL_ENV` - sound is affected by env volume 
This was done instead of using the existing `SEQ_PLAYER_*` constants to avoid confusion and because there isn't a `NONE`/`MASTER` option.

Additionally, sets the default to `MOD_AUDIO_CHANNEL_MUSIC` as to not break compatibility.

```lua
audio_stream_set_volume_channel(stream, MOD_AUDIO_CHANNEL_SFX) -- wow its just like a sample
audio_stream_get_volume_channel(stream) -- returns MOD_AUDIO_CHANNEL_SFX (its actually 2)
```


Blockyyy:
THANK YOU

Isaac0-dev (approved):

AgentXLP:
@kermeow merge upstream and I'll merge this in

kermeow:
rebased + changed to use property macro

PR #1209, opened 2026-04-26
Add Bad Net

bluelightzero:
Bad Net is a tool for simulating a bad network connection.

It wraps gNetworkSystemSocket and manipulates out-going packets
to recreate the same types of issues caused by an internet connection;
all from the comfort of your direct connection over localhost.

### DIFF
--- a/src/pc/network/badnet.c
+++ b/src/pc/network/badnet.c
@@ -1,0 +1,158 @@
+#include "badnet.h"
+
+#include "socket/socket.h"
+#include "pc/utils/misc.h"
+
+/*
+
+    Bad Net is a tool for simulating a bad network connection.
+
+    It wraps gNetworkSystemSocket and manipulates out-going packets
+    to recreate the same types of issues caused by an internet connection;
+    all from the comfort of your direct connection over localhost.
+
+*/
+
+struct DelayedPacket {
+    u8 localIndex;
+    u8* data;
+    u16 dataLength;
+    f64 time;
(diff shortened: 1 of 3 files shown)


bluelightzero:
This should not be merged in it's current state.

It is currently lacking UI or a way to configure.

Isaac0-dev:
I agree with the conversation, why not just use external tools?

bluelightzero:
> I agree with the conversation,

There is no conversation here, and the conversations I have had elsewhere are disappointing (and quite franky toxic).

Can you explain what it is you are agreeing with?

> why not just use external tools?

 - They are not cross platform (Clumsy is windows only)
 - They are not able to target specific application traffic. Affecting the internet connection of the development system directly is unacceptable.
 - Lack game specific customisation or automation.

To extend to people here what they are missing in context from the other conversation. I was suggesting adding a developer-mode-only set of lua functions to allows lua mods to control the lag. These functions would be completely disabled in normal online play and trying to use them online would result in only script errors.

Having a built in tool with lua mod support allows for recreating online conditions reliably, and share those with other developers/modders.

For example I could write a lua script that moves multiple players around the game or forces them to perform certain inputs and actions and also trigger packet drops which may cause desyncs or other issues. This could then simply be shared with another developer/modder and can recreate it on their system allowing them to fix the issue.

Non of this is currently possible with any of the external tools suggested to me (such as clumsy).

PR #1216, opened 2026-05-02
DJUI color interpolation / combiner editing / transforms

Cooliokid956:
Three sets of features that people of all ages will enjoy!

Color interpolation
---
- `djui_hud_set_color_interpolated`: Takes twice as many arguments as `djui_hud_set_color`
```
u8 prevR, u8 prevG, u8 prevB, u8 prevA, u8 r, u8 g, u8 b, u8 a
```
Combiner editing
---
- `djui_hud_set_combiner`
```
Sets the current DJUI HUD combiner.
Each part uses the following equation: `P = (A - B) * C + D`.
Cycle 2 may be used to extend the equation, with the result of the previous cycle accessible through CS_COMBINED

void djui_hud_set_combiner(u8 cycle, bool alpha, OPTIONAL enum CombinerSource a, OPTIONAL enum CombinerSource b, OPTIONAL enum CombinerSource c, OPTIONAL enum CombinerSource d);
```
- `djui_hud_reset_combiner`: Resets the above, reverting combine modes back to the defaults for each element
- `djui_hud_set_combiner_cycles`: Sets the amount of cycles used in rendering

Transforms
---
- `djui_hud_transform`: Applies a transformation (translate, rotate, scale) to subsequent DJUI HUD elements until closed
- `djui_hud_transform_interpolated`: The interpolated version of the previous function! Takes twice as many arguments
- `djui_hud_close_transform`: Closes the previous transform
- `djui_hud_reset_transform`: Closes all current transforms, restoring a fullscreen state


Cooliokid956:
Hi! I've had this PR in the works for a while but the third and final feature has had me stumped for a while... it all seems to work but somewhere along the way crashing has become prominent even when using regular DJUI functions... so I'd like to ask for guidance on the matter, and maybe also a review on the two already completed features. (@coderabbitai)

Cooliokid956:
Closed in favor of updated PR. Will bring back transform later

PR #1217, opened 2026-05-02
Add built-in preset screen shader effects configurable through Lua

AgentXLP:
# Shader Presets: 
* Hue Shift
* Saturation
* Brightness
* Contrast
* Exposure
* Dithering
* Posterization
* Scanlines

| Function Notation | Description |
| -------------------- | ------------- |
| `clear_all_shader_flags()` | Does something similar to disabling shader flags globably except it goes through and disables/resets each shader individually instead of toggling a global switch. |
| `get_global_shader_flags_enabled(enabled)` | Gets the global enabled state of screen shaders. |
| `set_global_shader_flags_enabled(enabled)` | Toggles the global enabled state of screen shaders, enabled by default. |
| | |
| `get_shader_flag_enabled(flag)` | Gets if a specific shader (`SHADER_FLAG_*`) is enabled. |
| `set_shader_flag_enabled(flag, enabled)` | Enables a specific shader (`SHADER_FLAG_*`) |
| `get_shader_flag_value(flag)` | Gets the value for a specific shader (`SHADER_FLAG_*`). Often ranges from 0.0 to 1.0 |
| `set_shader_flag_value(flag, value)` | Sets the value for a specific shader (`SHADER_FLAG_*`). Often ranges from 0.0 to 1.0 |

This is kind of like a more simple less experienced modder friendly version of Emerald's much more complex custom shader PR.
Also, trick: Set contrast to -1 to get inverted colors

## Hue:
image

## Saturation:
image

## Brightness:
image

## Contrast:
image

## Exposure:
image

## Dithering:
image

## Posterization:
image

## Scanlines:
image

## Dithering/Posterization/Scanlines:
image

Example Mod: [b&w.lua.txt](https://github.com/user-attachments/files/27308159/b.w.lua.txt)
You have to rename it to remove the `.txt` obviously but it's a decent example.

Get builds [here](https://github.com/coop-deluxe/sm64coopdx/actions/runs/25261305067)


Blockyyy:
i don't know if that was fixed yet or not (didn't get to test in a while) but i know there was an issue where the yellow particles (the ones you would see trailing from a 1up for example) would be white instead when i was playtesting this

AgentXLP:
Thanks, I'll look into that now

AgentXLP:
Alright try now

EmeraldLoc:
That fixed it 👍

src/pc/gfx/gfx_opengl.c
```diff
@@ -379,6 +436,15 @@ static struct ShaderProgram *gfx_opengl_create_and_load_new_shader(struct ColorC
         append_line(fs_buf, &fs_len, "uniform vec3 uLightmapColor;");
     }
 
+    if (world_geometry) {
+        char line[64];
```
EmeraldLoc: Probably best to use the syntax above, which would be `fs_len += sprintf(fs_buf + fs_len, "uniform sampler2D uTex%d;\n", t);`

src/pc/gfx/gfx_opengl.c
```diff
@@ -379,6 +436,15 @@ static struct ShaderProgram *gfx_opengl_create_and_load_new_shader(struct ColorC
...
+        snprintf(line, 64, "uniform int uShaderFlags[%d];", SHADER_FLAG_MAX);
+        append_line(fs_buf, &fs_len, line);
+        
+        snprintf(line, 64, "uniform float uShaderFlagValues[%d];", SHADER_FLAG_MAX);
+        append_line(fs_buf, &fs_len, line);
```
EmeraldLoc: Probably best to use the syntax above, which would be `fs_len += sprintf(fs_buf + fs_len, "uniform sampler2D uTex%d;\n", t);` but for the strings here

EmeraldLoc (changes requested):
Everything appears to be working, there's only a single very minor format preference I have, but I like multiple things here, so it's probably better to merge this pr first so I can make changes to my shader pr.

src/pc/gfx/gfx_opengl.c
AgentXLP: Thanks

EmeraldLoc (approved):
This looks good, but someone else may want to verify the DL changes on vanilla models won't have any unintentional side effects

AgentXLP:
`gsSPVertexNonGlobal` is a custom vtx type I wrote for early sm64coopdx that skips any lua light color or vertex color or lighting engine color calculations and just renders raw.
Looking back, I would have went for `gsSPVertexRaw` or something akin to that to make it more clear but can't go back now lol

Me changing them back was because the shaders weren't applying to them since they're skipping all of the extra mod related processing
Me changing them to non global is because they were being affected by the shaders when they shouldn't be (e.g. menu/HUD element)

Everything should be fine but I did miss the sparkle color so yeah you're right

EmeraldLoc:
Was messing with this, wish this was in my review but whatever, for `enable_shader_flags_global ` it'd also be nice to have a `get_shader_flags_global ` as well. It would make things such as toggling it significantly easier, as right now you need to keep track of it's state manually, and another mod can change it to ruin your local state.

AgentXLP:
Good idea, I'll get to work on this

AgentXLP:
Ready for review

AgentXLP:
@coderabbitai alright, I've never used this before but me and my co developers think this PR is good to go. Can you review to see if you can find any issues that could cause problems down the line or something of the sort?

PR #1218, opened 2026-05-03
Unify renderers into one build, add renderer selectionbox

AgentXLP:
This unifies OpenGL and DirectX on Windows into one binary, no more separate binaries.
For people who can't run the OpenGL build at all somehow, you can run sm64coopdx in cmd with `sm64coopdx --backend directx` to force the game to use that backend instead of whatever the config says (OpenGL by default)

Marked as a draft because I need to make sure Linux works and compiles still


AgentXLP:
image

Cooliokid956:
you should probably use the Direct3D naming (direct3d11, d3d11, d3d, etc.)

AgentXLP:
The graphics backend option also switches the window manager as well so I feel like "DirectX 11" is more accurate.

"Graphics Backend" is a simpler label to understand than separate options or "Rendering/Window Backend" so that's why I stuck with that too

AgentXLP:
Linux compiles! Just need to test it as well as various other features I touched with this PR

AgentXLP:
Tested on linux to confirm parity
Mouse input works, controller input works, keybind names work, headless works, the option in the menu works, crash screen works, loading screen works, system path functions work
This is ready for review

Isaac0-dev:
hi, looks like you've got merge conflicts

AgentXLP:
Right, I'll fix those

.github/workflows/build-coop.yaml
```diff
@@ -112,63 +110,13 @@ jobs:
             - name: Zip the game
               run: |
                 cd ./build/us_pc
-                zip -r sm64coopdx_Windows_OpenGL.zip dynos lang mods palettes coop.map discord_game_sdk.dll sm64coopdx.exe
+                zip -r sm64coopdx_Windows.zip dynos lang mods palettes coop.map discord_game_sdk.dll sm64coopdx.exe
 
             - name: Upload artifact
               uses: actions/upload-artifact@v4
               with:
                 name: sm64coopdx-windows-opengl
```
EmeraldLoc: This should be changed to `sm64coopdx-windows`

AgentXLP:
Does someone want to review this? If not I can try myself but I did test every window manager related potential issue I could think of on both backends

EmeraldLoc (commented):
Only have one thing. Tested on windows, macOS, and linux, all platforms compile and work as expected. Windows was tested via the action run. Another pair of eyes should probably look at this.

EmeraldLoc (approved):
Works fine. Looks fine to me but someone else should take a look at this

PeachyPeachSM64:
works as intended. Direct X seems to ignore the FPS cap when VSync is on, but I don't think it's an issue related to this PR.

PeachyPeachSM64 (approved):

Isaac0-dev:
> Direct X seems to ignore the FPS cap when VSync is on

If it locks it to the monitor's refresh rate, that's expected. vsync forcefully does the timing on sdl

PR #1219, opened 2026-05-03
massive performance improvements for add_scroll_target

Isaac0-dev:
The title probably sounds a bit over hype, I know.
To get straight to the point, I noticed that loading the scroll targets was probably the slowest parts to loading a romhack.
The reason for this is that many romhacks can have thousands of calls to `add_scroll_target`.
So, for 4,107 calls to `add_scroll_target`, the time went from ~4.5759 seconds to ~0.0173 seconds in total. 
Changes made:
- Previously, simply finding the material data to scroll was rather slow due to using a linear sub string search across all vertices in all levels. To speed this up, I added a cache.
The cache bypasses checking every level by storing the exact string (rather than the substring) in a hashmap, so lookups become a simple case of a string lookup as a key in the map. It falls back to the full lookup if the cache doesn't hit.
- Changed the vertex buffer management in `scroll_targets.c` to behave closer to a modern dynamic array, where buffer size is doubled each time a new vertex buffer is added, to reduce the number of allocations performed.

This load time improvement was too good to not make a PR about.

### DIFF
--- a/data/dynos_misc.cpp
+++ b/data/dynos_misc.cpp
@@ -10,17 +10,28 @@ extern "C" {
 // Scroll Targets
 //
 
-static std::unordered_multimap<std::string_view, DataNode<Vtx>*> sVertexNodesExactMap;
-static std::vector<GfxData*> sVertexNodesCacheLvls;
+void DynOS_Add_Scroll_Target_Match(u32 index, const char* name, u32 offset, u32 size, DataNode<Vtx>* node) {
+    if (offset >= node->mSize) { return; }
+    u32 finalSize = (size > 0 && size <= (node->mSize - offset)) ? size : (node->mSize - offset);
+    add_vtx_scroll_target(
+        index,
+        &node->mData[offset],
+        finalSize,
+        offset > 0
+    );
+}
 
 void DynOS_Add_Scroll_Target(u32 index, const char* name, u32 offset, u32 size) {
+    static std::unordered_multimap<std::string_view, DataNode<Vtx>*> sVertexNodesExactMap;
+    static std::vector<GfxData*> sLvlGfxDataCache; // cache existing level pointers to know when to rebuild
+
     auto& lvlArray = DynOS_Lvl_GetArray();
 
     // Check if cache needs rebuilding
-    bool rebuild = (lvlArray.size() != sVertexNodesCacheLvls.size());
+    bool rebuild = (lvlArray.size() != sLvlGfxDataCache.size());
     if (!rebuild) {
         for (size_t i = 0; i < lvlArray.size(); ++i) {
-            if (lvlArray[i].second != sVertexNodesCacheLvls[i]) {
+            if (lvlArray[i].second != sLvlGfxDataCache[i]) {
                 rebuild = true;
                 break;
             }
@@ -29,9 +40,9 @@ void DynOS_Add_Scroll_Target(u32 index, const char* name, u32 offset, u32 size) 
 
     if (rebuild) {
         sVertexNodesExactMap.clear();
-        sVertexNodesCacheLvls.clear();
+        sLvlGfxDataCache.clear();
         for (const auto& lvlPair : lvlArray) {
-            sVertexNodesCacheLvls.push_back(lvlPair.second);
+            sLvlGfxDataCache.push_back(lvlPair.second);
             for (const auto& node : lvlPair.second->mVertices) {
                 sVertexNodesExactMap.emplace(std::string_view(node->mName.begin(), node->mName.Length()), node);
             }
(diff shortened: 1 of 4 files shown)


data/dynos_misc.cpp
```diff
@@ -7,14 +10,61 @@ extern "C" {
 // Scroll Targets
 //
 
+static std::unordered_multimap<std::string_view, DataNode<Vtx>*> sVertexNodesExactMap;
+static std::vector<GfxData*> sVertexNodesCacheLvls;
```
PeachyPeachSM64: If I understand correctly, `sVertexNodesExactMap` holds the vertex nodes of all levels with their name as keys, and `sVertexNodesCacheLvls` holds the references to custom levels.
Since both are specific caches used only in `DynOS_Add_Scroll_Target`, they should be defined inside the function, not outside.

That being said, `sVertexNodesCacheLvls` is a basically mirror of `DynOS_Lvl_GetArray` (but only for the GfxData pointers), why not naming it `sLvlGfxDataCache` for better clarity?

As for `sVertexNodesExactMap`, ideally you wouldn't need a multimap... But sadly, the scroll target feature as a whole is not well designed...

data/dynos_misc.cpp
```diff
@@ -7,14 +10,61 @@ extern "C" {
...
+                if (offset >= node->mSize) { continue; }
+                u32 finalSize = (size > 0 && size <= (node->mSize - offset)) ? size : (node->mSize - offset);
                 add_vtx_scroll_target(
                     index,
                     &node->mData[offset],
-                    (size > 0 && size < node->mSize) ? size : node->mSize,
+                    finalSize,
                     offset > 0
                 );
             }
```
PeachyPeachSM64, line 66: Some duplicate code that could be moved into a function
Also, any reason why the name check here is `node->mName.Find(name) >= 0` instead of `node->mName == name`? This is the only place where the comparison is done this way...

src/pc/lua/smlua_functions.c
```diff
@@ -821,8 +821,11 @@ int smlua_func_add_scroll_target(lua_State* L) {
     }
 
     // add_scroll_target used to require offset and size of the vertex buffer to be used
-    if (!smlua_functions_valid_param_range(L, 2, 4)) { return 0; }
     int paramCount = lua_gettop(L);
+    if (paramCount < 2 || paramCount > 4) {
+        LOG_LUA_LINE("Improper param count: Expected (2 - 4), Received %u", paramCount);
+        return 0;
+    }
```
PeachyPeachSM64, line 828: Why this change? This is exactly the same
On that note, the Lua and VSCode documentation for `add_scroll_target` is outdated (2 params instead of 4)

PeachyPeachSM64 (changes requested):
While this is not the purpose of this PR, the scroll target feature would need some rework.

First, you can assign multiple vertex buffers to the same id. And I mean globally. All ids are shared between all levels of all mods. With some `add_scroll_target` manipulation, you can prevent (intentionally or not) other mods levels scroll targets from working properly.

Second, you can assign a scroll target to any vertex buffer existing in any level of any mod. No mod isolation or restriction. (Unfortunately this is an issue with all functions requesting DynOS resources)

Third, you can assign a scroll target *only* to level vertex buffers. Not actor, not `vtx_create` ones.

src/pc/lua/smlua_functions.c
Isaac0-dev: That change was probably the most aggressive optimisation here, because `smlua_func_add_scroll_target` already called `lua_gettop`, I expanded the function to avoid calling it twice, as I understand that the bridge between Lua and C can be particularly slow. That being said, I don't believe that bridge is at all the bottleneck here, but it's just an improvement I saw.

`add_scroll_target` requires 2 parameters officially, but has legacy support for extra parameters.
This is because the system corrects itself later on if you don't provide the parameters. It's better to provide them, but they were originally removed due to the complexity of manually adding these commands.

data/dynos_misc.cpp
Isaac0-dev: Your understanding of the two caches is accurate!
Thanks for the name suggestion. Yeah, I went with the multimap because it was a way to maintain logical consistency (as this is an optimisation).

data/dynos_misc.cpp
Isaac0-dev: Good point.
It's checking for substring matches by doing that, but I don't know why it was originally coded that way.
I'm also hesitant to change it because I don't know if anyone's mod relies on it.

src/pc/lua/smlua_functions.c
PeachyPeachSM64: oh I see, so the 4 params was previous behavior
nothing to change there then

data/dynos_misc.cpp
PeachyPeachSM64: it's fine for now

PeachyPeachSM64 (approved):

PR #1220, opened 2026-05-03
Fix DynOS generation bug with duplicates

PeachyPeachSM64:
When generating multiple actors, if two or more data nodes share the same name, the generated actors can reference the wrong nodes in their display lists/geo layouts.
This results in:
- The wrong nodes compiled into binary files
- Intermediate layouts generated as actors since nothing is referencing them anymore, treating them as root geo layouts

The simplest fix was to first check for the model identifier (renamed to `mDataIdentifier`, since it's no longer related to models) when looking for nodes, to ensure it picks the right node, even if multiple nodes share the same name.


data/dynos_bin_animation.cpp
```diff
@@ -241,11 +241,9 @@ void DynOS_Anim_Table_Load(BinFile *aFile, GfxData *aGfxData) {
     // Data
     String _AnimationName; _AnimationName.Read(aFile);
     if (_AnimationName != "NULL") {
-        for (auto &_AnimData : aGfxData->mAnimations) {
-            if (_AnimData->mName == _AnimationName) {
-                _AnimationPtr = (void *) _AnimData->mData;
-                break;
-            }
+        auto _AnimData = aGfxData->mAnimations.Find(_AnimationName);
```
Isaac0-dev, line 244: wait, `Find(name)`? why not `Find(name, dataIdentifier)`?
Isn't that just the same bug again without the dataIdentifier?
Looks like this is the case elsewhere in this PR too.

data/dynos_bin_gfx.cpp
```diff
@@ -456,6 +457,14 @@ s64 DynOS_Gfx_ParseGfxConstants(const String& _Arg, bool* found) {
     return 0;
 }
 
+template <typename T>
+static DataNode<T> *FindLightNode(GfxData *aGfxData, DataNodes<T> &aDataNodes, String aName) {
+    char *_NameStart = aName.begin() + (aName[0] == '&' ? 1 : 0);
```
Isaac0-dev, line 462: perhaps check `aName` length here (expect at least more than 1 character).
Since you're expecting at least one character after the `&`, and it avoids the redundant lookup later if it's too short.

data/dynos_bin_read.cpp
```diff
@@ -32,8 +32,8 @@ static IfDefPtr GetNearestIfDefPointer(char *pFileBuffer) {
 char *DynOS_Read_Buffer(FILE* aFile, GfxData* aGfxData) {
     fseek(aFile, 0, SEEK_END);
     s32 _Length = ftell(aFile);
-    if (aGfxData && aGfxData->mModelIdentifier == 0) {
-        aGfxData->mModelIdentifier = (u32) _Length;
+    if (aGfxData && aGfxData->mDataIdentifier == 0) {
+        aGfxData->mDataIdentifier = (u32) _Length;
```
Isaac0-dev: the length isn't a very reliable "unique" identifier, we need something unique to avoid the collision problem we're trying to solve here. It also should be consistent with the method in `dynos_bin_lvl.cpp`.

Isaac0-dev (commented):
I'm particularly interested to know why you didn't use the data identifier field in every call to `Find`, isn't that just keeping the original bug?

data/dynos_bin_animation.cpp
PeachyPeachSM64: dataIdentifier is only relevant for the generation part, i.e. C -> binary
for bin loading, you can only search by name

data/dynos_bin_gfx.cpp
PeachyPeachSM64: sorry, but I'm not sure I understand
what would be the benefit of adding a length check here?

data/dynos_bin_read.cpp
PeachyPeachSM64: what do you suggest?

Isaac0-dev: perhaps a global large bit size integer which gets incremented within every `_GeneratePack` function?

Isaac0-dev:
@coderabbitai review

data/dynos_bin_actor.cpp
```diff
@@ -177,8 +177,8 @@ static void DynOS_Actor_Generate(const SysPath &aPackFolder, Array<Pair<u64, Str
         _GfxData->mGeoNodeStack.Clear();
 
         // Parse data
-        PrintNoNewLine("%s.bin: Model identifier: %X - Processing... ", _GeoRootName.begin(), _GfxData->mModelIdentifier);
-        PrintConsole(CONSOLE_MESSAGE_INFO, "%s.bin: Model identifier: %X - Processing... ", _GeoRootName.begin(), _GfxData->mModelIdentifier);
+        PrintNoNewLine("%s.bin: Model identifier: %llX - Processing... ", _GeoRootName.begin(), _GfxData->mDataIdentifier);
+        PrintConsole(CONSOLE_MESSAGE_INFO, "%s.bin: Model identifier: %llX - Processing... ", _GeoRootName.begin(), _GfxData->mDataIdentifier);
```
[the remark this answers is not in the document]
Isaac0-dev: Is that necessary? this file is for handling models specifically. The field was renamed because it's used for non models, but this console log is still accurate because it's referring to the identifier of the model.

data/dynos_bin_col.cpp
```diff
@@ -760,8 +760,8 @@ void DynOS_Col_Generate(const SysPath &aPackFolder, Array<Pair<u64, String>> _Ac
         _GfxData->mLoadIndex = 0;
 
         // Parse data
-        PrintNoNewLine("%s.col: Collision identifier: %X - Processing... ", _ColRootName.begin(), _GfxData->mModelIdentifier);
-        PrintConsole(CONSOLE_MESSAGE_INFO, "%s.col: Collision identifier: %X - Processing... ", _ColRootName.begin(), _GfxData->mModelIdentifier);
+        PrintNoNewLine("%s.col: Collision identifier: %llX - Processing... ", _ColRootName.begin(), _GfxData->mDataIdentifier);
+        PrintConsole(CONSOLE_MESSAGE_INFO, "%s.col: Collision identifier: %llX - Processing... ", _ColRootName.begin(), _GfxData->mDataIdentifier);
```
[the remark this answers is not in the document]
Isaac0-dev: @PeachyPeachSM64 this is a good point ^

data/dynos_bin_tex.cpp
```diff
@@ -256,16 +256,14 @@ DataNode<TexData>* DynOS_Tex_Load(BinFile *aFile, GfxData *aGfxData) {
...
-            }
+        auto _LoadedNode = aGfxData->mTextures.Find(_NodeName);
+        if (_LoadedNode) {
+            _Node->mData->mPngData   = _LoadedNode->mData->mPngData;
+            _Node->mData->mRawData   = _LoadedNode->mData->mRawData;
+            _Node->mData->mRawWidth  = _LoadedNode->mData->mRawWidth;
+            _Node->mData->mRawHeight = _LoadedNode->mData->mRawHeight;
+            _Node->mData->mRawFormat = _LoadedNode->mData->mRawFormat;
+            _Node->mData->mRawSize   = _LoadedNode->mData->mRawSize;
         }
```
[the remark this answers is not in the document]
Isaac0-dev: Not entirely important and not an issue in this PR. I guess it's a potential failure point but not important here.

data/dynos_bin_read.cpp
```diff
@@ -117,9 +122,14 @@ char *DynOS_Read_Buffer(FILE* aFile, GfxData* aGfxData) {
 
 template <typename T>
 static void AppendNewNode(GfxData *aGfxData, DataNodes<T> &aNodes, const String &aName, String *&aDataName, Array<String> *&aDataTokens) {
+    if (aNodes.FindExact(aName, aGfxData->mDataIdentifier)) {
+        PrintDataError("  ERROR: Node \"%s\" already exists for data identifier: %llX", aName.begin(), aGfxData->mDataIdentifier);
+        return;
```
Isaac0-dev: regurgitating coderabbitai comment in my words:
`aDataTokens` cannot be left null here, which will happen if you return without adding something.
If it is null, it seems the code will get stuck in a loop of hitting the `else if (!pDataTokens) {` block, which will be true each time. 
coderabbit mentioned that the `DATA_TYPE_UNUSED` path will set `pDataTokens` to a fake pointer of 1 to bypass this check.
If you choose to copy this behaviour you will need to add some safety check to prevent it from trying to parse the actual content.

data/dynos_bin_col.cpp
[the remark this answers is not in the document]
PeachyPeachSM64: done

data/dynos_mgr_movtexqc.cpp
```diff
@@ -25,16 +25,15 @@ void DynOS_MovtexQC_Register(const char* name, s16 level, s16 area, s16 type) {
...
+        if (node) {
+            // add it
+            _DynosRegisteredMovtexQCs.push_back({
+                .dataNode = node,
+                .level    = level,
+                .area     = area,
+                .type     = type
+            });
         }
     }
```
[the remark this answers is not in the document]
PeachyPeachSM64: not the scope of this PR, but fixed it anyway

data/dynos_bin_read.cpp
PeachyPeachSM64: done
aDataName and aDataTokens will point to the existing node's

data/dynos_bin_read.cpp
```diff
@@ -117,10 +122,16 @@ char *DynOS_Read_Buffer(FILE* aFile, GfxData* aGfxData) {
...
+    } else {
+        _Node = New<DataNode<T>>();
+        _Node->mName = aName;
+        _Node->mDataIdentifier = aGfxData->mDataIdentifier;
+        aNodes.Add(_Node);
+    }
+
     aDataName = &_Node->mName;
     aDataTokens = &_Node->mTokens;
 }
```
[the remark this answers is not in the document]
PeachyPeachSM64: fixed

Isaac0-dev (approved):
nice, looks clean!

PR #1221, opened 2026-05-03
Fix rendering issues with held objects

PeachyPeachSM64:
Add a way to save and load a gfx state in display lists.
Its primary purpose is to save the gfx state before rendering a held object, then restore it later.
It fixes the issue where Mario had opaque legs with the vanish and metal caps when holding an object.

Before:
image

After:
image


src/pc/gfx/gfx_pc.c
```diff
@@ -2245,6 +2255,144 @@ static void OPTIMIZE_O3 djui_gfx_sp_simple_tri1(uint8_t vtx1_idx, uint8_t vtx2_i
     return;
 }
 
+static void gfx_sp_load_or_save_state(uint8_t cmd, uint32_t state) {
+
+    // Load state
+    if (cmd == G_STATE_LOAD) {
+        SUPPORT_CHECK(num_gfx_states > 0);
```
Isaac0-dev: since you've exposed this to dynos, I think this shouldn't crash but instead throw some verbose error. 
It should explain that there are more load state commands than save states.

src/pc/gfx/gfx_pc.c
```diff
@@ -2245,6 +2255,144 @@ static void OPTIMIZE_O3 djui_gfx_sp_simple_tri1(uint8_t vtx1_idx, uint8_t vtx2_i
...
+        }
+        if (state & G_STATE_COLOR_IMAGE) {
+            rdp.color_image_address = gfx_state->rdp.color_image_address;
+        }
+        return;
+    }
+
+    // Save state
+    if (cmd == G_STATE_SAVE) {
+        SUPPORT_CHECK(num_gfx_states < MAX_GFX_STATES);
```
Isaac0-dev: same here, but the error should say there are too many save states

src/game/rendering_graph_node.c
```diff
@@ -1420,8 +1448,32 @@ static s32 obj_is_in_view(struct GraphNodeObject *node, Mat4 matrix) {
...
+}
+
+static void geo_save_object_gfx_state(void) {
+    geo_append_display_list(obj_save_gfx_state, LAYER_OPAQUE);
+    geo_append_display_list(obj_save_gfx_state, LAYER_OPAQUE_DECAL);
+    geo_append_display_list(obj_save_gfx_state, LAYER_OPAQUE_INTER);
+    geo_append_display_list(obj_save_gfx_state, LAYER_ALPHA);
+    geo_append_display_list(obj_save_gfx_state, LAYER_TRANSPARENT);
+    geo_append_display_list(obj_save_gfx_state, LAYER_TRANSPARENT_DECAL);
+    geo_append_display_list(obj_save_gfx_state, LAYER_TRANSPARENT_INTER);
```
Isaac0-dev: It's probably time to have a function called `geo_append_dl_to_all_layers()`, to reduce these repeated patterns.

Isaac0-dev (changes requested):

src/game/rendering_graph_node.c
PeachyPeachSM64: done

src/pc/gfx/gfx_pc.c
PeachyPeachSM64: you're in the middle of the gfx processing, you can't throw an error like in any other part of the code
instead, I'll just silently return without doing anything

src/pc/gfx/gfx_pc.c
PeachyPeachSM64: same here

Isaac0-dev (approved):

Isaac0-dev:
merge conflicts

Isaac0-dev (approved):

PR #1222, opened 2026-05-04
optimised loading the mod cache

Isaac0-dev:
Previously, the mod cache had quadratic time complexity, because it was looking through the entire array every time a new entry was added to the cache.

This optimises the mod cache by adding 2 hashmaps, which store the file path, and the file data as an md5 hash.
For this, a new hashmap type was added to `dynos_cmap.cpp`, which can handle large keys since the existing structures can handle 64 bit keys only; MD5 hashes are 128 bits, and truncating it would increase hash collisions. 

Also, `sModCacheEntries` was changed from a flat array of structs to a dynamic array of pointers to those structs.
Because before, we were copying an entire structure, so this reduces memory copying significantly.

Also added the new field `arrayIndex` to each cache entry. This removes the need to lookup the position of a node to remove it.


data/dynos_cmap.cpp
```diff
@@ -134,6 +135,40 @@ class HMap {
...
+        mMap->clear();
+    }
+
+    size_t size() const {
+        return mMap->size();
+    }
+
+private:
+    std::unique_ptr<std::unordered_map<std::string, void*>> mMap;
+};
```
PeachyPeachSM64: instead of creating a new class, you could template the "key" type of the original class to avoid duplicating the HMap logic
Actually, you could even template the useUnordered parameter, since you're not going to change it once it's created

data/dynos_cmap.cpp
```diff
@@ -185,4 +220,34 @@ void* hmap_next(void* map) {
...
+
+void hmap_data_clear(void* map) {
+    if (!map) { return; }
+    reinterpret_cast<HMapData*>(map)->clear();
+}
+
+void hmap_data_destroy(void* map) {
+    if (!map) { return; }
+    delete reinterpret_cast<HMapData*>(map);
+}
```
PeachyPeachSM64, line 165: of course you can't do templating in C, but you could pass a "type" parameter to the C functions to tell which key type you're handling something like this for example:
```c
enum HMapKeyType {
    HMAP_KEY_INT64,
    HMAP_KEY_STRING,
};

...

void *hmap_create(enum HMapKeyType type, bool useUnordered) {
    switch (type) {
        case HMAP_KEY_INT64: return (void *) new HMap<useUnordered, int64_t>();
        case HMAP_KEY_STRING: return (void *) new HMap<useUnordered, std:string>();
        default: return NULL;
    }
}

and so on...
```

data/dynos_cmap.cpp.h
```diff
@@ -13,5 +13,12 @@ size_t hmap_len(void* map);
 void* hmap_begin(void* map);
 void* hmap_next(void* map);
 
+void* hmap_data_create(void);
+void* hmap_data_get(void* map, const char* data, size_t len);
+void hmap_data_put(void* map, const char* data, size_t len, void* value);
+void hmap_data_del(void* map, const char* data, size_t len);
+void hmap_data_clear(void* map);
+void hmap_data_destroy(void* map);
```
PeachyPeachSM64, line 23: These wouldn't be needed anymore

PeachyPeachSM64 (changes requested):

data/dynos_cmap.cpp.h
Isaac0-dev: I don't think we can remove these, we need to store a key of at least 128 bits, since we're storing md5 hashes as keys.

data/dynos_cmap.cpp
```diff
@@ -3,186 +3,164 @@
...
+    if (!map) { return; }
+    std::string key(data, len);
+    static_cast<IHMap<std::string>*>(map)->put(key, value);
+}
+
+void hmap_data_del(void* map, const char* data, size_t len) {
+    if (!map) { return; }
+    std::string key(data, len);
+    static_cast<IHMap<std::string>*>(map)->erase(key);
+}
```
PeachyPeachSM64: please change parameter name `data` to `key`

data/dynos_cmap.cpp.h
```diff
@@ -13,5 +15,12 @@ size_t hmap_len(void* map);
 void* hmap_begin(void* map);
 void* hmap_next(void* map);
 
+void* hmap_data_create(void);
+void* hmap_data_get(void* map, const char* data, size_t len);
+void hmap_data_put(void* map, const char* data, size_t len, void* value);
+void hmap_data_del(void* map, const char* data, size_t len);
```
PeachyPeachSM64: same here

PeachyPeachSM64 (changes requested):
Sadly I don't think we can get rid of these functions, since they also require the `len` parameter
It's fine to keep them, to make a distinction between integer based keys and arbitrary sized keys

PeachyPeachSM64 (approved):

PR #1225, opened 2026-05-04
Dynamic object limit (Again)

Sunketchupm:
I decided to look into growing arrays again and this time found out how to deallocate in the growing array. The dynamic object limit this time seems to be very stable. There still is a bug where having more than 4096 objects (+ some more) causes newly spawned objects to not load their model, so a limit as been set anyways.

This might be slower performance wise than just increasing the size of the object pool, since every deallocation needs to run a loop to find the object that needs to be deallocated. It may be solvable with a hash map, but I am not able to think of a way to do so that is faster.


AgentXLP:
@coderabbitai review

Sunketchupm:
Part of the reason why I'm pretty much avoiding the "reuse stale or unload later" approach is that fake objects don't crash or cause any issues, which is really the only way to maintain a reference to an unloaded object. So I don't need to complicate the code any further.

src/engine/surface_load.c
```diff
@@ -859,8 +859,13 @@ void clear_dynamic_surfaces(void) {
 
     clear_spatial_partition(&gDynamicSurfacePartition[0][0]);
 
-    for (u16 i = 0; i < OBJECT_POOL_CAPACITY; i++) {
-        gObjectPool[i].numSurfaces = 0;
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but object surfaces were attempted to be cleared.\n");
+        return;
+    }
```
PeachyPeachSM64, line 865: Is it a fatal error though?
It doesn't really matter if there is no object pool, since what the following code does is simply setting the objects numSurfaces to 0
Could be replaced by:
`LOG_ERROR("The object pool was not created but object surfaces were attempted to be cleared.");`

src/engine/surface_load.c
```diff
@@ -859,8 +859,13 @@ void clear_dynamic_surfaces(void) {
...
+    for (u16 i = 0; i < gObjectPool->count; i++) {
+        struct Object* obj = gObjectPool->buffer[i];
+        obj->numSurfaces = 0;
     }
```
PeachyPeachSM64: you can do this now:
```c
growing_array_for_each_(gObjectPool, struct Object, obj) {
    obj->numSurfaces = 0;
}
```

src/game/behaviors/tree_particles.inc.c
```diff
@@ -16,7 +16,7 @@ void bhv_tree_snow_or_leaf_loop(void) {
         obj_mark_for_deletion(o);
     if (o->oTimer > 100)
         obj_mark_for_deletion(o);
-    if (gPrevFrameObjectCount > (OBJECT_POOL_CAPACITY * 212 / 240))
+    if (gPrevFrameObjectCount > OBJECT_POOL_CAPACITY - 200)
```
PeachyPeachSM64: Might be good to have a macro for this, like:
`#define OBJECT_POOL_PARTICLES_THRESHOLD (OBJECT_POOL_CAPACITY - 200)`

src/game/object_helpers.c
```diff
@@ -2529,14 +2524,14 @@ void cur_obj_spawn_particles(struct SpawnParticlesInfo *info) {
     s32 numParticles = info->count;
 
     // If there are a lot of objects already, limit the number of particles
-    if (gPrevFrameObjectCount > (OBJECT_POOL_CAPACITY * 150 / 240) && numParticles > 10) {
+    if (gPrevFrameObjectCount > OBJECT_POOL_CAPACITY - 200 && numParticles > 10) {
```
PeachyPeachSM64: same here

src/game/object_helpers.c
```diff
@@ -2529,14 +2524,14 @@ void cur_obj_spawn_particles(struct SpawnParticlesInfo *info) {
...
         numParticles = 10;
     }
 
 
     // We're close to running out of object slots, so don't spawn particles at
     // all
-    if (gPrevFrameObjectCount > (OBJECT_POOL_CAPACITY * 210 / 240)) {
+    if (gPrevFrameObjectCount > OBJECT_POOL_CAPACITY - 50) {
```
PeachyPeachSM64: same here

src/game/object_list_processor.c
```diff
@@ -89,7 +90,7 @@ struct ObjectNode *gObjectLists;
 /**
  * A singly linked list of available slots in the object pool.
  */
-struct ObjectNode gFreeObjectList;
+UNUSED struct ObjectNode gFreeObjectList;
```
PeachyPeachSM64, line 93: can be removed if unused

src/game/object_list_processor.c
```diff
@@ -587,21 +587,20 @@ void clear_objects(void) {
...
+    for (s32 i = 0; i < 60; i++) {
         gDoorAdjacentRooms[i][0] = 0;
         gDoorAdjacentRooms[i][1] = 0;
     }
 
+    gObjectPool = growing_array_init(gObjectPool, OBJECT_POOL_INIT_CAPACITY, malloc, free);
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: Could not initialize the object pool!\n");
+        return;
+    }
```
PeachyPeachSM64, line 601: however, this is indeed a fatal error and the game should be killed
change to:
`sys_fatal("Could not initialize the object pool!");`

src/game/object_list_processor.c
```diff
@@ -615,7 +614,8 @@ void clear_objects(void) {
 void update_terrain_objects(void) {
     gObjectCounter = update_objects_in_list(&gObjectLists[OBJ_LIST_SPAWNER]);
     //! This was meant to be +=
-    gObjectCounter = update_objects_in_list(&gObjectLists[OBJ_LIST_SURFACE]);
+    // Fixed
```
PeachyPeachSM64: no need for the comment, removing the original comment already indicates that you fixed it

src/game/platform_displacement.c
```diff
@@ -169,6 +170,8 @@ void apply_platform_displacement(struct Object *o, struct Object *platform) {
  */
 void apply_mario_platform_displacement(void) {
     for (s32 i = 0; i < MAX_PLAYERS; i++) {
+        struct NetworkPlayer* np = &gNetworkPlayers[i];
+        if (!np || !np->connected) { continue; }
```
PeachyPeachSM64: unrelated bug fix?
any reason to not use `is_player_active` instead?

src/game/spawn_object.c
```diff
@@ -77,46 +79,51 @@ struct LinkedList *unused_try_allocate(struct LinkedList *destList,
...
-    struct ObjectNode *nextObj = NULL;
+struct Object *try_allocate_object(struct ObjectNode *destList) {
+    if (destList == NULL) {
+        fprintf(stderr, "FATAL ERROR: Failed to try and allocate a object because the destList %p was NULL!\n", destList);
+        return NULL;
+    }
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but an object allocation was attempted!\n");
+        return NULL;
+    }
```
PeachyPeachSM64, line 94: Not fatal, the game can keep running, that's why it's called *try_allocate_object*.
Change the prints to:
`LOG_ERROR("Failed to try and allocate a object because the destList was NULL!");`
and
`LOG_ERROR("The object pool was not created but an object allocation was attempted!");`

src/game/spawn_object.c
```diff
@@ -77,46 +79,51 @@ struct LinkedList *unused_try_allocate(struct LinkedList *destList,
...
 
-    if (destList == NULL || freeList == NULL) {
-        fprintf(stderr, "FATAL ERROR: Failed to try and allocate a object because either the destList %p or freeList %p was NULL!\n", destList, freeList);
+    if (gObjectPool->count >= OBJECT_POOL_CAPACITY) {
         return NULL;
     }
```
PeachyPeachSM64, line 98: Maybe add a `LOG_DEBUG("Reached object pool hard capacity, cannot allocate new objects);`?
But I'm afraid it would generate too much spam in the console...

src/game/spawn_object.c
```diff
@@ -77,46 +79,51 @@ struct LinkedList *unused_try_allocate(struct LinkedList *destList,
...
 
-    geo_remove_child(&nextObj->gfx.node);
-    geo_add_child(&gObjParentGraphNode, &nextObj->gfx.node);
+    // Insert at end of destination list
+    nextObj->header.prev = destList->prev;
+    nextObj->header.next = destList;
+    if (destList->prev != NULL) {
+        destList->prev->next = &nextObj->header;
+    } else {
+        fprintf(stderr, "ERROR: The previous object in the destination list %p was NULL! Unexpected errors may occur.\n", destList);
```
PeachyPeachSM64: LOG_ERROR

PeachyPeachSM64: also, what happens if destList->prev is NULL? doesn't that mean that the list is just empty?

src/game/spawn_object.c
```diff
@@ -135,38 +142,21 @@ void unused_deallocate(struct LinkedList *freeList, struct LinkedList *node) {
...
-        obj++;
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but an object deallocation was attempted!\n");
+        return;
     }
-
-    // End the list
-    obj->header.next = NULL;
+    growing_array_swap_and_pop(gObjectPool, obj);
+    obj = NULL;
```
PeachyPeachSM64: `obj = NULL;`: unneeded line

src/game/spawn_object.c
```diff
@@ -135,38 +142,21 @@ void unused_deallocate(struct LinkedList *freeList, struct LinkedList *node) {
...
-    // Add the first object in the pool to the free list
-    struct Object *obj = &gObjectPool[0];
-    gFreeObjectList.next = (struct ObjectNode *) obj;
-
-    // Link each object in the pool to the following object
-    for (s32 i = 0; i < poolLength - 1; i++) {
-        obj->header.next = &(obj + 1)->header;
-        obj++;
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but an object deallocation was attempted!\n");
```
PeachyPeachSM64: LOG_ERROR

src/game/spawn_object.c
```diff
@@ -135,38 +142,21 @@ void unused_deallocate(struct LinkedList *freeList, struct LinkedList *node) {
...
+        return;
     }
-
-    // End the list
-    obj->header.next = NULL;
+    growing_array_swap_and_pop(gObjectPool, obj);
+    obj = NULL;
+    gObjectPool->buffer[gObjectPool->count] = NULL;
```
PeachyPeachSM64, line 158: No, you shouldn't do that. That's a memory leak.
Growing array items persist even after the count is decremented, to reuse the already allocated slot for future allocations (and technically prevents stale pointer access).

src/pc/network/packets/packet_level_macro.c
```diff
@@ -18,8 +18,12 @@
 
 // TODO: move to common utility location
 static struct Object* get_object_matching_respawn_info(s16* respawnInfo) {
-    for (s32 i = 0; i < OBJECT_POOL_CAPACITY; i++) {
-        struct Object* o = &gObjectPool[i];
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but object respawn infos were attempted to be checked.\n");
```
PeachyPeachSM64: LOG_ERROR

src/pc/network/packets/packet_level_macro.c
```diff
@@ -18,8 +18,12 @@
...
+        return NULL;
+    }
+    for (u32 i = 0; i < gObjectPool->count; i++) {
+        struct Object* o = gObjectPool->buffer[i];
```
PeachyPeachSM64: growing_array_for_each_

src/pc/network/packets/packet_level_macro.c
```diff
@@ -188,16 +192,16 @@ void network_receive_level_macro(struct Packet* p) {
...
             // coin formation
             if (behavior == smlua_override_behavior(bhvCoinFormation)) {
                 o->oBehParams = *respawnInfo;
                 o->oCoinUnkF4 = (o->oBehParams >> 8) & 0xFF;
 
                 u8 childIndex = 0;
-                for (s32 i = 0; i < OBJECT_POOL_CAPACITY; i++) {
-                    struct Object* o2 = &gObjectPool[i];
+                for (u32 i = 0; i < gObjectPool->count; i++) {
+                    struct Object* o2 = gObjectPool->buffer[i];
```
PeachyPeachSM64: growing_array_for_each_

src/pc/network/packets/packet_level_macro.c
```diff
@@ -207,8 +211,8 @@ void network_receive_level_macro(struct Packet* p) {
                 }
                 LOG_INFO("rx macro special: coin formation");
             } else if (behavior == bhvGoombaTripletSpawner) {
-                for (s32 i = 0; i < OBJECT_POOL_CAPACITY; i++) {
-                    struct Object* o2 = &gObjectPool[i];
+                for (u32 i = 0; i < gObjectPool->count; i++) {
+                    struct Object* o2 = gObjectPool->buffer[i];
```
PeachyPeachSM64: growing_array_for_each_

src/pc/network/packets/packet_level_spawn_info.c
```diff
@@ -16,8 +16,12 @@
...
 // TODO: move to common utility location
 static struct Object* get_object_matching_respawn_info(u32* respawnInfo) {
-    for (s32 i = 0; i < OBJECT_POOL_CAPACITY; i++) {
-        struct Object* o = &gObjectPool[i];
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but object respawn infos were attempted to be checked.\n");
+        return NULL;
+    }
+    for (u32 i = 0; i < gObjectPool->count; i++) {
+        struct Object* o = gObjectPool->buffer[i];
```
PeachyPeachSM64: growing_array_for_each_

src/pc/network/packets/packet_level_spawn_info.c
```diff
@@ -16,8 +16,12 @@
 
 // TODO: move to common utility location
 static struct Object* get_object_matching_respawn_info(u32* respawnInfo) {
-    for (s32 i = 0; i < OBJECT_POOL_CAPACITY; i++) {
-        struct Object* o = &gObjectPool[i];
+    if (gObjectPool == NULL) {
+        fprintf(stderr, "FATAL ERROR: The object pool was not created but object respawn infos were attempted to be checked.\n");
```
PeachyPeachSM64, line 20: LOG_ERROR

PeachyPeachSM64 (changes requested):
Pretty solid so far
The 4096 objects bug should probably be investigated, but not in this PR

src/game/platform_displacement.c
Sunketchupm: Related bug fix, since there is (sometimes) a crash otherwise. `is_player_active` also checks if the player is in a bubble and that might affect bubble behavior on moving platforms. That might not matter though, so I will investigate and probably change this.

src/game/spawn_object.c
Sunketchupm: This is the bug that happens when I remove setting the buffer to NULL.
https://github.com/user-attachments/assets/d56cc93a-327e-4e2f-80d8-19f03de9f62f

src/game/object_list_processor.c
```diff
@@ -589,21 +589,20 @@ void clear_objects(void) {
         gMarioStates[i].currentRoom = 0;
     }
 
-    for (i = 0; i < 60; i++) {
+    for (s32 i = 0; i < 60; i++) {
         gDoorAdjacentRooms[i][0] = 0;
         gDoorAdjacentRooms[i][1] = 0;
     }
 
+    gObjectPool = growing_array_init(gObjectPool, OBJECT_POOL_INIT_CAPACITY, malloc, free);
```
Isaac0-dev, line 597: I think the object pool should be initialised on startup, to do this use the `AT_STARTUP` flag with a new function. 
This is because the object pool used to exist at all times during program execution, so let's try to keep that.

PR #1227, opened 2026-05-06
Expose some variables in mario_actions_submerged.c

Sunketchupm:
All of them are now part of the MarioState struct

### DIFF
--- a/autogen/lua_definitions/structs.lua
+++ b/autogen/lua_definitions/structs.lua
@@ -1183,6 +1183,15 @@
 --- @field public floorAngle integer
 --- @field public waterLevel integer
 --- @field public currentRoom integer
+--- @field public swimStrength integer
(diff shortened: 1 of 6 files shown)


Isaac0-dev:
this looks more like something that should be done in `hardcoded.c`, with it's own structure for mario specific constants.

AgentXLP:
Yeah I don't know how I feel about exposing these either, like it feels weird to have specifically swimming related variables be editable which are part of Mario's moveset, but not anything else

Sunketchupm:
I thought about this for a little while. There is no good way to expose them without adding them to the MarioState struct (adding an entire new global for just swimming variables seems pretty bad). I also agree with X's comment. Going to close this, although I'm going to make a new PR that fixes the bug with bobbing.

PR #1228, opened 2026-05-07
Fix sync valid packet issues for the server

EmilyEmmi:
When the server received a sync valid packet from another player, it would also set the host player's level, due to some code that's meant to inform the host also running for the host themselves. While I'm unsure of the exact conditions in which this happens, it can be observed by playing certain mods like MarioHunt or Geoguessr where the issue commonly occurs and viewing debug_info:

```
Fri May 01 13:33:39 2026 [00] [INFO] packet_sync_valid.c: rx sync valid
Fri May 01 13:33:39 2026 [00] [INFO] packet_sync_valid.c: informing server of sync valid
Fri May 01 13:33:39 2026 [00] [INFO] packet_level_area_inform.c: tx level area inform for global 0: (6, 2, 7, 1)
```

The code in question will no longer run for the host, which should fix this issue. I don't think this necessarily fixes the infamous "objects despawning" bug, but at least this fixes another longstanding bug that I've been pulling my hair out over for the past 2 years.

### DIFF
--- a/src/pc/network/network.c
+++ b/src/pc/network/network.c
@@ -617,7 +617,7 @@ void network_update(void) {
             bool inCredits = (np->currActNum == 99);
             if (gNetworkType == NT_SERVER && (npAny == NULL || inCredits)) {
                 // no NetworkPlayer in the level
-                network_send_sync_valid(np, np->currCourseNum, np->currActNum, np->currLevelNum, np->currAreaIndex);
+                network_send_sync_valid(np, np->currCourseNum, np->currActNum, np->currLevelNum, np->currAreaIndex, false);
                 return;
             }
 
--- a/src/pc/network/packets/packet.h
+++ b/src/pc/network/packets/packet.h
@@ -315,7 +315,7 @@ void network_send_area(struct NetworkPlayer* toNp);
 void network_receive_area(struct Packet* p);
 
 // packet_sync_valid.c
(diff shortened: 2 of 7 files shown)


EmeraldLoc (approved):

AgentXLP (dismissed):

EmilyEmmi:
Switched to draft because it apparently breaks stuff? Somehow? Needs more research

AgentXLP:
Yeah I was worried about that so I didn't merge

EmilyEmmi:
This function had a much bigger issue, as it turns out: the server didn't differentiate between packets it was meant to validate for other players, and packets meant to validate _itself_, which are used when the host enters a level where another player is already present. I made some more significant modifications to account for this, and now the server properly differentiates between the two scenarios. ~~Ironically, this means that the server still informs itself about its own sync valid state, even though the original commit prevented this.~~ EDIT: The server doesn't need to inform itself, as Isaac pointed out. Doing so is redundant and calls the HOOK_ON_SYNC_VALID hook twice.

I tested with 4 windows on direct connection, and the original bug has been fixed, and syncing also appears to behave as normal. Also, sync valid packets are a whopping **one byte** larger now, which is insignificant since this packet isn't called every frame.

Isaac0-dev (approved):

PR #1229, opened 2026-05-08
Fix Lua/custom behaviors bugs and flaws

PeachyPeachSM64:
This PR aims to improve custom behaviors by fixing its main flaws. Previously, one couldn't:
- Hook a vanilla behavior more than once
- Hook a custom behavior more than once without creating another behavior
- Mix both a custom and Lua behavior
- Have more than 1 init and update callback per behavior

First, the hooked behavior struct. Removed a lot of redundant (and confusing) fields, now custom behaviors are split into 3 categories:
- **Lua callbacks** (init/update), which run after the object behavior.
- **Lua behaviors**, which consist of Lua callbacks, but fully override the object original behavior.
- **Custom behaviors**, which are loaded from DynOS `.bhv` files and can also run Lua callbacks.

Internally, each hooked behavior has an original behavior id and a custom id.
The **custom id** is, as the name suggests, a unique id given to each new hooked behavior.
The **behavior id** is either the vanilla id if it's a vanilla behavior hook, or the custom id.

How does it work, and what are the benefits?

When the game tries to hook a behavior, it looks for an existing one that has the same **behavior id** or **custom id**.
Which means each hooked behavior is mapped to either a unique vanilla behavior or itself, allowing to hook the same behavior multiple times while keeping the same reference.
This allows to give any behavior (vanilla, Lua, DynOS) as much callbacks (and names, see later) as needed, with or without replacing it.

Caveats/limitations:
- Custom (DynOS) behaviors cannot be overridden with a Lua or another custom behavior (trying to call `hook_behavior` with `replaceBehavior` set to `true` will raise a warning)
- Object list cannot be changed for Lua and custom behaviors (for obvious reasons)
- Calling `hook_behavior` with `replaceBehavior` set to `false` for Lua and custom behaviors has no effect: once overridden, you cannot revert the behavior to its original state

**-- Expected results from the `get_behavior` functions --**

Assume `behavior` is an object behavior (`o.behavior`) and `id` is the result of `get_id_from_behavior(behavior)`.
The five following functions still behave like this:

| Function | Behavior is vanilla | Behavior has lua callbacks (`hook_behavior`, `replaceBehavior` is false) | Behavior is full Lua (`hook_behavior`, `replaceBehavior` is true) | Behavior is custom (DynOS `.bhv`) |
| - | - | - | - | - |
| `get_id_from_behavior(behavior)` | vanilla id | hooked custom id | hooked custom id | hooked custom id |
| `get_id_from_vanilla_behavior(behavior)` | vanilla id | `id_bhv_max_count` | `id_bhv_max_count` | `id_bhv_max_count` |
| `get_behavior_from_id(id)` | vanilla script | vanilla script | hooked script | hooked script |
| `get_vanilla_behavior_from_id(id)` | vanilla script | `nil` | `nil` | `nil` |
| `get_behavior_name_from_id(id)` | vanilla name | hooked name* | hooked name* | hooked name* |

`get_id_from_behavior_name` returns the behavior id corresponding to the provided name. If the name matches a hooked behavior, the hooked **customId** is returned.
It is important to note that it checks hooked behaviors first, so if a mod does `hook_behavior(..., "bhvGoomba")`, `get_id_from_behavior_name("bhvGoomba")` will return the hooked custom id, not the vanilla `id_bhvGoomba`.

*Hooked name refers to the last registered hooked behavior name. That way, a hooked behavior can be given a new name using `hook_behavior`. No name is discarded, meaning a hooked behavior custom id can still be retrieved using an old name.

---

All of this was tested using the `Behaviors on screen` mod (https://discord.com/channels/752682015614173235/755907254318006362/1341527894974791732).


src/pc/lua/smlua_hooks.c
```diff
@@ -815,88 +835,154 @@ int smlua_hook_behavior(lua_State* L) {
...
+    switch (initReferenceType) {
+        case LUA_TNIL: break;
+
+        case LUA_TFUNCTION: {
+            lua_pushvalue(L, 4);
+            initReference = luaL_ref(L, LUA_REGISTRYINDEX);
+        } break;
+
+        default: {
+            LOG_LUA_LINE("Hook behavior: invalid type passed for argument initFunction: %d, should be %d", initReferenceType, LUA_TFUNCTION);
```
Isaac0-dev: Convert the confusing integers to a more verbose name with `lua_typename`.

src/pc/lua/smlua_hooks.c
```diff
@@ -815,88 +835,154 @@ int smlua_hook_behavior(lua_State* L) {
...
+    switch (loopReferenceType) {
+        case LUA_TNIL: break;
+
+        case LUA_TFUNCTION: {
+            lua_pushvalue(L, 5);
+            loopReference = luaL_ref(L, LUA_REGISTRYINDEX);
+        } break;
+
+        default: {
+            LOG_LUA_LINE("Hook behavior: invalid type passed for argument loopFunction: %d, should be %d", loopReferenceType, LUA_TFUNCTION);
```
Isaac0-dev: same here

src/pc/lua/smlua_hooks.c
```diff
@@ -815,88 +835,154 @@ int smlua_hook_behavior(lua_State* L) {
...
+            case LUA_TSTRING: {
+                bhvName = smlua_to_string(L, 6);
+                if (!bhvName || !gSmLuaConvertSuccess) {
+                    LOG_LUA_LINE("Hook behavior: could not parse bhvName");
+                    return 0;
+                }
+            } break;
+
+            default: {
+                LOG_LUA_LINE("Hook behavior: invalid type passed for argument bhvName: %d, should be %d", bhvNameType, LUA_TSTRING);
```
Isaac0-dev: same here

data/behavior_data.c
```diff
@@ -2732,6 +2740,8 @@ const BehaviorScript bhvSunkenShipPart[] = {
 };
 
 const BehaviorScript bhvSunkenShipSetRotation[] = {
+    BEGIN(OBJ_LIST_DEFAULT),
+    ID(id_bhvSunkenShipSetRotation),
```
Isaac0-dev, line 2744: Missing a +2 for this behavior in `bhvInSunkenShip2`?

AgentXLP:
@coderabbitai revieww

src/game/memory.h
```diff
@@ -83,6 +83,9 @@ bool growing_array_swap_and_pop(struct GrowingArray *array, void *ptr);
 void growing_array_free(struct GrowingArray **array);
 void growing_array_debug_print(struct GrowingArray *array, const char *name, s32 x, s32 y);
 
+#define growing_array_for_each_(array, type, item) \
+    for (type *item = array->buffer[0], **_head_ = (type **) array->buffer, **_tail_ = (type **) (array->buffer + array->count); _head_ != _tail_; _head_++, item = *_head_)
```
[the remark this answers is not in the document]
PeachyPeachSM64: I wasn't a fan of the double for loop, so I went with a safer approach which also checks the `array` pointer

src/pc/lua/smlua_hooks.c
```diff
@@ -918,90 +1004,72 @@ int smlua_hook_behavior(lua_State* L) {
...
+        callback->modFile = gLuaActiveModFile;
+    }
 
     // We want to push the behavior into the global LUA state. So mods can access it.
     // It's also used for some things that would normally access a LUA behavior instead.
-    lua_pushinteger(L, customBehaviorId);
+    lua_pushinteger(L, hooked->behaviorId);
     lua_setglobal(L, bhvName);
-    LOG_INFO("Registered custom behavior: 0x%04hX - %s", customBehaviorId, bhvName);
+    LOG_INFO("Registered Lua behavior for behavior id 0x%04hX (custom id: 0x%04hX, custom name: %s)", hooked->behaviorId, hooked->customId, bhvName);
```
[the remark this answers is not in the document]
PeachyPeachSM64: fixed, log info will use last registered name if bhvName is NULL

src/pc/lua/smlua.h
```diff
@@ -19,9 +19,58 @@
...
+            gLuaActiveMod->showedScriptWarning = true; \
+        } \
+        printf("[LUA] Warning: "); \
+        printf(__VA_ARGS__); \
+        printf("\n"); \
+        smlua_mod_warning(); \
+        snprintf(gDjuiConsoleTmpBuffer, CONSOLE_MAX_TMP_BUFFER, __VA_ARGS__); \
+        sys_swap_backslashes(gDjuiConsoleTmpBuffer); \
+        djui_console_message_create(gDjuiConsoleTmpBuffer, CONSOLE_MESSAGE_WARNING); \
+    } \
```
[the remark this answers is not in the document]
PeachyPeachSM64: fixed

src/pc/lua/smlua_hooks.c
PeachyPeachSM64: done

src/pc/lua/smlua_hooks.c
PeachyPeachSM64: done

src/pc/lua/smlua_hooks.c
PeachyPeachSM64: done

data/behavior_data.c
PeachyPeachSM64: good catch, thanks

src/game/memory.h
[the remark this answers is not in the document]
PeachyPeachSM64: thanks, I fixed it

docs/lua/functions-3.md
```diff
@@ -256,10 +256,33 @@ Gets a behavior script from a behavior ID
...
+## [get_vanilla_behavior_from_id](#get_vanilla_behavior_from_id)
+
+### Description
+Gets the behavior script corresponding to the provided `id` if it's a vanilla behavior, `nil` otherwise
+
+### Lua Example
+`local pointerValue = get_vanilla_behavior_from_id(id)`
+
+### Parameters
+| Field | Type |
```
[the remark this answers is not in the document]
PeachyPeachSM64: this is an auto-generated file, please ignore

Isaac0-dev (approved):

Isaac0-dev:
I've done the best I can in reviewing this. The only concern I have is the fact that you moved where the behaviour init callback is called from, which seems to skip the mandatory header which you're enforcing now. 
There is still the chance this may cause issues for mods that expect to be called immediately.

PR #1231, opened 2026-05-09
Add an auto updater

EmeraldLoc:
Thanks to @AgentXLP for laying the main groundwork of this!

This pr adds an updater builtin the game

It works by using a separate executable for updating the game. The updater exists next to the executable. In the game, if an update is available, it checks if this executable exists. If it does it offers to update the game for you. When you hit yes, the game launches the updater and closes itself.

The updater's first job is to copy itself to a temporary directory and then launch that copied updater as a new process and then it closes itself. This new process downloads the game and replaces all the files with the updated files. It then launches the new sm64coopdx exec, and closes itself. That's about how the new system works

This pr should be merged alongside [this pr](https://github.com/coop-deluxe/coopdx-updater/pull/1)

Tested on windows, macOS, and linux

Action run (there's not a whole lot to do here but it's here lol) <https://github.com/EmeraldLoc/sm64coopdx/actions/runs/25589747421>


Blockyyy:
can we please get this for 1.5 this is something that should have existed a long while ago

src/pc/utils/misc.c
```diff
@@ -594,3 +601,47 @@ void str_seperator_concat(char *output_buffer, int buffer_size, char** strings,
         }
     }
 }
+
+static char *get_update_path() {
+#ifdef _WIN32
+    char updateExecFilename[] = "coopdx-updater.exe"; // WIP
```
AgentXLP: `// WIP`?

src/pc/utils/misc.c
```diff
@@ -594,3 +601,47 @@ void str_seperator_concat(char *output_buffer, int buffer_size, char** strings,
...
+#else
+    char updateExecFilename[] = "coopdx-updater";
+#endif
+    static char sUpdateExecFilePath[SYS_MAX_PATH];
+    // this may truncate as sys_exe_path_dir is allocated to be of size SYS_MAX_SIZE, nonetheless such a limit should not be hit during normal use.
+    snprintf(sUpdateExecFilePath, sizeof(sUpdateExecFilePath), "%s%s%s", sys_exe_path_dir(), PATH_SEPARATOR, updateExecFilename);
+    return sUpdateExecFilePath;
+}
+
+bool can_update_game() {
```
AgentXLP: `(void) {`
I'm also gonna be going through and finding other instances of this in the C code since you're supposed to put void when there's no parameters.

AgentXLP (dismissed):
Just a few tiny things

src/pc/utils/misc.c
EmeraldLoc: Whoops, forgot to remove that lol

src/pc/utils/misc.c
EmeraldLoc: Will note!

PR #1232, opened 2026-05-10
fix audio engine crash from reloading banks that are already loaded

Isaac0-dev:
When a custom sequence is loaded, the sound bank is loaded into a temporary cache slot, but we weren't checking if the bank was already loaded in the cache, so we attempted to load it again.

Some romhacks use the same sound bank for multiple sequences. When a new sequence played that uses the same bank, the game would repeatedly reload the same bank into the temporary cache slots. This could cause the currently active copy of the bank to be unloaded while notes were still using it, crashes from stale audio pointers.

This fix checks if the bank is already available before loading it, but if it's loaded we just mark the bank as loaded again.
I put the fix in the custom sequence loading block. The fix I added can actually be found in other places in the sm64 sound engine, we just lacked it specifically for custom sequences.
I also included a safety fix to avoid the crash if a situation like this arises again, which resets the pointers that were going stale.

Old screenshot of the crash screen:
image

The way I was testing this fix is by using the [sm64 decades later romhack port](https://github.com/Isaac0-dev/decades-later-coop), which loads every custom sequence to the extended sound bank (0x2A).
I triggered the crash by going to the first main level (bobomb valley, overrides BOB) and collecting a star to trigger the star celebration fanfare.

### DIFF
--- a/src/audio/heap.c
+++ b/src/audio/heap.c
@@ -228,6 +228,13 @@ void discard_bank(s32 bankId) {
                 note->parentLayer->finished = TRUE;
             }
             note_disable(note);
+
+            // Reset sample state
+            note->sound = NULL;
+            note->samplePosInt = 0;
+            note->samplePosFrac = 0;
+            note->sampleDmaIndex = 0;
+
             audio_list_remove(&note->listItem);
             audio_list_push_back(&gNoteFreeLists.disabled, &note->listItem);
         }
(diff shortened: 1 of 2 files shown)


Isaac0-dev:
@coderabbitai can you review

Blockyyy:
i love DM9 Rebirth

Isaac0-dev:
@coderabbitai configuration

Isaac0-dev:
@coderabbitai configuration

PR #1235, opened 2026-05-12
Multi-line descriptions

Cooliokid956:
They look like this now
image
image

PS: encourage longer descriptions to be formatted like this
image


Cooliokid956:
@coderabbitai is this permissible

PR #1237, opened 2026-05-12
Unexpose update functions

EmeraldLoc:
Blunder made by yours truly, when running autogen the update functions would appear, this adds those functions to the blacklist.


Isaac0-dev:
Thanks!

PR #1240, opened 2026-05-13
much further draw distance option

Isaac0-dev:
The existing draw distance options don't actually increase the draw distance. It only impacts whether or not distant objects are rendered. 

This PR:
- Adds another option to the draw distance setting in the display menu, called "Infinite". It is not truly infinite, but it is significantly larger than what current options allow. Due to it not being truly infinite, we could use the name "Max" instead to be more accurate.
- Exposes a new function to the Lua API `draw_distance_scalar_is_infinite` which returns whether or not the infinite setting is enabled.
- `draw_distance_scalar_is_infinite` is now used in several places in this repo to bypass distance checks for objects if infinite mode is enabled.
- Fixes a bug where you couldn't bypass the distance check in `obj_is_in_view`, meaning you could never disable all object distance culling.
- The infinite setting now forces the far plane to be at least a minimum of `1,000,000`.


Isaac0-dev:
translations provided by AI

src/game/rendering_graph_node.c
```diff
@@ -310,6 +310,10 @@ void patch_mtx_interpolated(f32 delta) {
         f32 fovInterpolated = delta_interpolate_f32(sPerspectiveNode->prevFov, sPerspectiveNode->fov, delta);
         f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(sPerspectiveNode->near, gProjectionMaxNearValue), gOverrideNear);
         f32 far = replace_value_if_not_zero(sPerspectiveNode->far, gOverrideFar);
+
+        // "infinite" draw distance
+        if (configDrawDistance == 6) { far = max(far, 1000000.f); }
```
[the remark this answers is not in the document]
Isaac0-dev: `gBehaviorValues.InfiniteRenderDistance` is not meant to impact the rendering of everything, it's only made for objects.

src/game/rendering_graph_node.c
```diff
@@ -310,6 +312,10 @@ void patch_mtx_interpolated(f32 delta) {
         f32 fovInterpolated = delta_interpolate_f32(sPerspectiveNode->prevFov, sPerspectiveNode->fov, delta);
         f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(sPerspectiveNode->near, gProjectionMaxNearValue), gOverrideNear);
         f32 far = replace_value_if_not_zero(sPerspectiveNode->far, gOverrideFar);
+
+        // "infinite" draw distance
+        if (configDrawDistance == 6) { far = max(far, MAX_FAR_PLANE_DIST); }
```
EmeraldLoc: You're directly overriding `gOverrideFar`, which can be set by Lua mods with `set_override_far`. If `gOverrideFar` is not zero, you shouldn't run this code.

src/game/rendering_graph_node.c
```diff
@@ -650,6 +656,10 @@ static void geo_process_perspective(struct GraphNodePerspective *node) {
     gProjectionVanillaFarValue = node->far;
     f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(node->near, gProjectionMaxNearValue), gOverrideNear);
     f32 far = replace_value_if_not_zero(node->far, gOverrideFar);
+
+    // "infinite" draw distance
+    if (configDrawDistance == 6) { far = max(far, MAX_FAR_PLANE_DIST); }
```
EmeraldLoc: Same here

EmeraldLoc (changes requested):

src/game/rendering_graph_node.c
```diff
@@ -310,6 +312,10 @@ void patch_mtx_interpolated(f32 delta) {
         f32 fovInterpolated = delta_interpolate_f32(sPerspectiveNode->prevFov, sPerspectiveNode->fov, delta);
         f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(sPerspectiveNode->near, gProjectionMaxNearValue), gOverrideNear);
         f32 far = replace_value_if_not_zero(sPerspectiveNode->far, gOverrideFar);
+
+        // "infinite" draw distance
+        if (gOverrideNear != 0 && configDrawDistance == 6) { far = max(far, MAX_FAR_PLANE_DIST); }
```
EmeraldLoc: This still isn't fixed. This is for the far plane, so you should be checking `gOverrideFar == 0`, `gOverrideNear` is unrelated. ~~Also I made a mistake in my last review~~ I misremembered, I said it fine lol, it should be checking if `gOverrideFar` is set to 0. If it's not then override the far plane. That's because `gOverrideFar` default's to zero. If lua overrides it, then it will be nonzero. Hope that makes sense this time lol

src/game/rendering_graph_node.c
```diff
@@ -650,6 +656,10 @@ static void geo_process_perspective(struct GraphNodePerspective *node) {
     gProjectionVanillaFarValue = node->far;
     f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(node->near, gProjectionMaxNearValue), gOverrideNear);
     f32 far = replace_value_if_not_zero(node->far, gOverrideFar);
+
+    // "infinite" draw distance
+    if (gOverrideNear != 0 && configDrawDistance == 6) { far = max(far, MAX_FAR_PLANE_DIST); }
```
EmeraldLoc: This still isn't fixed. This is for the far plane, so you should be checking `gOverrideFar == 0`, `gOverrideNear` is unrelated. ~~Also I made a mistake in my last review~~ I misremembered, I said it fine lol, it should be checking if `gOverrideFar` is set to 0. If it's not then override the far plane. That's because `gOverrideFar` default's to zero. If lua overrides it, then it will be nonzero. Hope that makes sense this time lol

EmeraldLoc (changes requested):

src/game/rendering_graph_node.c
Isaac0-dev: :facepalm: i did this without thinking lol

src/game/rendering_graph_node.c
Isaac0-dev: @EmeraldLoc good catch i think its actually resolved this time

src/game/rendering_graph_node.c
```diff
@@ -310,6 +312,10 @@ void patch_mtx_interpolated(f32 delta) {
         f32 fovInterpolated = delta_interpolate_f32(sPerspectiveNode->prevFov, sPerspectiveNode->fov, delta);
         f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(sPerspectiveNode->near, gProjectionMaxNearValue), gOverrideNear);
         f32 far = replace_value_if_not_zero(sPerspectiveNode->far, gOverrideFar);
+
+        // "infinite" draw distance
+        if (gOverrideNear == 0 && configDrawDistance == 6) { far = max(far, MAX_FAR_PLANE_DIST); }
```
EmeraldLoc: It's still `gOverrideNear` 😭

src/game/rendering_graph_node.c
```diff
@@ -650,6 +656,10 @@ static void geo_process_perspective(struct GraphNodePerspective *node) {
     gProjectionVanillaFarValue = node->far;
     f32 near = get_first_person_enabled() ? 1.f : replace_value_if_not_zero(MIN(node->near, gProjectionMaxNearValue), gOverrideNear);
     f32 far = replace_value_if_not_zero(node->far, gOverrideFar);
+
+    // "infinite" draw distance
+    if (gOverrideNear == 0 && configDrawDistance == 6) { far = max(far, MAX_FAR_PLANE_DIST); }
```
EmeraldLoc: It's still `gOverrideNear` 😭

EmeraldLoc (changes requested):

src/game/rendering_graph_node.c
Isaac0-dev: :facepalm: :facepalm: :facepalm: 
thanks for your patience!

EmeraldLoc (approved):
Ok this looks good now lol

PR #1246, opened 2026-05-16
Resolve issue with windows compilation due to use of `strcasestr`

EmeraldLoc:
Fixes issue introduced by https://github.com/coop-deluxe/sm64coopdx/commit/bb142d5769a2b24e8b3b2db0b5edb3c240992fbe because someone didn't check the C standard to verify it was ok to use (it was me).

`strcasestr` is not apart of the C standard and is a GNU extension. Windows does not play well with this. The solution was to define a quite simple custom function partially based on the glibc implementation. Function should be reviewed.

### DIFF
--- a/src/pc/djui/djui_panel_dynos.c
+++ b/src/pc/djui/djui_panel_dynos.c
@@ -50,7 +50,7 @@ static void djui_panel_dynos_add_packs(struct DjuiBase* base) {
         // filter results
         if (sSearchInputbox != NULL &&
             sSearchInputbox->buffer != NULL &&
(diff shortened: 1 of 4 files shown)


EmeraldLoc:
dont code when you're on the verge of sleeping and misread `cmp` for `case` and think it's a drop in

this is an action run to make sure windows compiles https://github.com/EmeraldLoc/sm64coopdx/actions/runs/25953231699

im going to bed

Isaac0-dev:
> dont code when you're on the verge of sleeping

can confirm, these are words of wisdom

Isaac0-dev (approved):

PR #1249, opened 2026-05-17
dynos pack sequence overrides

Isaac0-dev:
Moved the core sequence overriding code from `smlua_audio_utils` to DynOS, and allowed DynOS packs to override sequences by placing m64 files in the pack directory.
Put the integer settings for the override in the m64 name separated by dots, for example `seq.8.42.100.m64`.
Sequences are auto reloaded upon selecting the pack.


AgentXLP:
Holy peak

Isaac0-dev:
This is ready now

data/dynos.cpp.h
```diff
@@ -648,16 +644,79 @@ struct ActorGfx {
...
+struct AudioOverrideEntry {
+    u8 sequenceId;
+    bool enabled;
+    bool loaded;
+    char* filename;
+    u64 length;
+    u8 bank;
+    u8 defaultVolume;
+    u8* buffer;
+};
```
PeachyPeachSM64, line 660: In `load_sequence_internal`, sequence id is u32, but in `smlua_audio_utils_replace_sequence`, it's u8
which one is correct? what's the actual range of sequence ids?

Same for bank id. Sometimes it's u8, sometimes it's s32. Why?

data/dynos_mgr_audio.cpp
```diff
@@ -0,0 +1,256 @@
...
+            if (i >= SEQ_EVENT_CUTSCENE_LAKITU) {
+                sBackgroundMusicDefaultVolume[i] = 75;
+            } else {
+                sBackgroundMusicDefaultVolume[i] = sBackgroundMusicDefaultVolumeDefault[i];
+            }
+        } else if (hasPack) {
+            AudioOverrideEntry* activePack = DynOS_Audio_GetActivePackOverride(i);
+            sBackgroundMusicDefaultVolume[i] = activePack->defaultVolume;
+        }
+#else
```
PeachyPeachSM64, line 91: not sure if that's really necessary...

data/dynos_mgr_pack.cpp
```diff
@@ -35,6 +54,37 @@ static void ScanPackBins(struct PackData* aPack) {
             _TexName[length - 4] = '\0';
             DynOS_Tex_LoadFromBinary(aPack->mPath, _FileName, _TexName.begin(), true);
         }
+
+        // check for sequences
+        if (length > 4 && !strncmp(&_PackEnt->d_name[length - 4], ".m64", 4)) {
+            String _SeqName = _PackEnt->d_name;
+            _SeqName[length - 4] = '\0';
+            std::string seqName = _SeqName.begin();
+            std::regex re(".*_(\\d+)_(\\d+)_(\\d+)$");
```
PeachyPeachSM64: This should be documented somewhere.
Also, I would make the parameters separated by dots, not underscores.

PeachyPeachSM64 (changes requested):

xLuigiGamerx (commented):
Merge conflicts

data/dynos_mgr_audio.cpp
Isaac0-dev: doesn't hurt

data/dynos.cpp.h
Isaac0-dev: for both, `u8` is correct.
Max sequence ID is 128 (`MAX_AUDIO_OVERRIDE`).
Max bank ID is 64.
I went with `u8` because its correct, `load_sequence_internal` was written by nintendo devs who probably didn't care about integer size.

data/dynos_mgr_pack.cpp
Isaac0-dev: updated it to use dots instead of underscores
documentation is outside the scope of this pr, as nothing else about DynOS packs are documented.

Isaac0-dev:
merge conflicts are resolved

Isaac0-dev:
this pr is complete and ready for review

PeachyPeachSM64 (approved):

Issue #933, opened 2025-08-27
frame rendering is still not perfect


Isaac0-dev:
Even after the recent frame pacing changes on the dev branch, the game sometimes doesn't look smooth still.
In the clip below, there are visible frame drops. But there are 2 fps counters active and you can see that there is no drop in frame rate. 
This implies that there could be an issue with interpolation or delta calculation, or something. 
This does not mean the frame pacing changes are broken, but it does mean there's some other issue under the hood.

The display information used in this video are:
- Fedora Linux, OpenGL + SDL
- Vsync off
- Auto mode enabled, targeting 60fps (60fps monitor)

https://youtube.com/clip/UgkxTtvdH3S6Giag16gJmAhvEnTvW3WXnAL4

Isaac0-dev:
Turns out it was wayland skipping frames if i played in windowed mode.
Playing in fullscreen fixed the issue, and so did changing to x11 (`SDL_VIDEODRIVER=x11 ./sm64coopdx`).

Isaac0-dev:
seems to happen on android and windows too.

AgentXLP:
I don't experience this at all, weird

Isaac0-dev:
#1252 fixed the frame pacing. There are still occasional frame drops, but I believe this is due to the graphics interpreter being unoptimised as I noticed the function `send_display_list` can occasionally take longer than 34ms which forces a lag spike. 
The next step for improving frame rates is to optimise the gfx interpreter.

Isaac0-dev:
This is not completed

PR #1252, opened 2026-05-19
fix frame pacing code

Isaac0-dev:
I've made a lot of changes to the frame pacing code in the past, but I still experience poor frame pacing on my computer: https://github.com/coop-deluxe/sm64coopdx/issues/933.
But now I understand why. By using a tool called Tracy profiler (https://github.com/wolfpld/tracy), I was able to understand where the issues actually lie in the frame pacing.
Previously, I assumed we should swap buffers immediately after delaying, because how can we expect the timing to be right after delaying if the sm64 update loop is called every other frame, causing inconsistent timing? 
This PR takes the industry standard approach, where the buffers are swapped before delaying, so that the GPU has as much time as possible to display the frame. 

Of course, when I initially changed this code some time ago, I did it for a reason - lagging!
But I now know that this is not the frame pacing code (as of this PR), it is the code responsible for interpreting the display lists.

Even with this PR code enabled, I was able to trigger frame drops just by moving around in the game with power saving mode enabled on my laptop to reduce system performance. Here's a screenshot of Tracy, focused on one of those frame drops. You can see that the `send_display_list` call took 35ms alone in the first `produce_one_frame` call. That's way too much, the frame pacing code can't save us from that situation!
Screenshot From 2026-05-20 01-40-30

So I'm opening this PR because I believe it's the right fix for the frame pacing, but we'll need to look into optimising the display list interpreter if we want to fix these further frame drops. For now, this patch fixes the frame pacing code.

### DIFF
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -281,6 +281,7 @@ void produce_interpolation_frames_and_delay(void) {
     // make sure to draw at least one frame to prevent the game from freezing completely
     // (including inputs and window events) if the game update duration is greater than 33ms
     do {
+        curTime = clock_elapsed_f64();
         ++framesDrawn;
 
         // when we know how many frames to draw, use a precise delta
@@ -293,6 +294,7 @@ void produce_interpolation_frames_and_delay(void) {
         if (!gSkipInterpolationTitleScreen) { patch_interpolations(delta); }
         send_display_list(gGfxSPTask);
         gfx_end_frame_render();
+        gfx_display_frame();
 
         // delay if our framerate is capped
         if (shouldDelay) {
@@ -305,8 +307,6 @@ void produce_interpolation_frames_and_delay(void) {
             }
         }
 
-        // send the frame to the screen (should be directly after the delay for good frame pacing)
-        gfx_display_frame();
         sDrawnFrames++;
         if (shouldDelay) { numFramesToDraw--; }
     } while ((curTime = clock_elapsed_f64()) < targetTime && numFramesToDraw > 0);


EmeraldLoc:
Incase you don't know, the display list interpreter goes like this. `send_display_list` calls `gfx_run` with the display list command list, there the frame is setup and all display lists are then ran via `gfx_run_dl`. Ideally there would be a way to trace which display list instructions are taking the longest. Figured I'd give my 2 cents on that. Otherwise, looks fine to me and a good explanation but I wouldn't really know lol

EmeraldLoc (approved):

PeachyPeachSM64:
> Incase you don't know, the display list interpreter goes like this. `send_display_list` calls `gfx_run` with the display list command list, there the frame is setup and all display lists are then ran via `gfx_run_dl`. Ideally there would be a way to trace which display list instructions are taking the longest. Figured I'd give my 2 cents on that. Otherwise, looks fine to me and a good explanation but I wouldn't really know lol

Old screenshot from the built-in profiler in ex-omm (with Render96 model pack)
(columns are: elapsed time per frame, num calls per frame, average time per call)
image

The longest ones are unsurprisingly vertex and triangles processing, not only because of matrix multiplication and stuff, but because they are called so many times it adds a non negligible overhead!

EmeraldLoc:
How much would it improve performance to move matrix calculations to the shader? Vertices are set to clip space from the CPU, not the GPU, how much could it help to move those calc's to the GPU rather than doing them on the CPU?

PeachyPeachSM64:
By a lot.
Ideally all these operations should be performed on the GPU, because that's its main purpose!
And by all, I mean the whole pipeline (or most of it). Repeated memory transfers between CPU and GPU kill performance and should be avoided as much as possible.

EmeraldLoc:
May look into it for my shader stuff then, since once that's done what is set in stone is what is set in stone for the most part. It would also help making shaders if I were to move those to the GPU because then it would be a much more standard layout then what we have right now. Will look into it after I finish the syncing stuff!

PR #1256, opened 2026-05-22
fix custom level data memory leak

Isaac0-dev:
this fixes the giant memory leak that happens due to not freeing custom level data when closing a lobby.
`DynOS_Lvl_ModShutdown` was not freeing the data nodes for the level data, it was only freeing the surface data.
To fix this, I have made it use `DynOS_Gfx_Free` to free the `GfxData` correctly. I found that the level script VM will still be trying to warp from the custom level after `DynOS_Lvl_ModShutdown` is executed, so I added a schedule to simply free it the next frame.
I've made it force the level script to change to a vanilla level during mod shutdown. This is critical to ensure the VM doesn't continue to read from a freed level script.

Removed the explicit deletion of data nodes in `DynOS_MovtexQC_ModShutdown` because `DynOS_Gfx_Free` already frees that, and it's actually data owned by the data node, so it's more appropriate in `DynOS_Gfx_Free`

### DIFF
--- a/data/dynos_level.cpp
+++ b/data/dynos_level.cpp
@@ -21,7 +21,7 @@ extern const BehaviorScript *sWarpBhvSpawnTable[];
 
 #define DYNOS_LEVEL_MARIO_POS_WARP_ID (-1)
 
-extern void *gDynosLevelScriptsOriginal[LEVEL_COUNT];
+extern const void *gDynosLevelScriptsOriginal[LEVEL_COUNT];
 
 void DynOS_Level_ParseScript(const void *aScript, s32 (*aPreprocessFunction)(u8, void *));
 
@@ -158,7 +158,7 @@ void DynOS_Level_Init() {
 
         // Level warps
         for (sDynosCurrentLevelNum = 0; sDynosCurrentLevelNum < LEVEL_COUNT; ++sDynosCurrentLevelNum) {
-            sDynosLevelScripts[sDynosCurrentLevelNum].mLevelScript = gDynosLevelScriptsOriginal[sDynosCurrentLevelNum];
+            sDynosLevelScripts[sDynosCurrentLevelNum].mLevelScript = (void *) gDynosLevelScriptsOriginal[sDynosCurrentLevelNum];
             sDynosLevelScripts[sDynosCurrentLevelNum].mModIndex = DYNOS_LEVEL_MOD_INDEX_VANILLA;
             if (sDynosLevelScripts[sDynosCurrentLevelNum].mLevelScript) {
                 DynOS_Level_ParseScript(sDynosLevelScripts[sDynosCurrentLevelNum].mLevelScript, DynOS_Level_PreprocessScript);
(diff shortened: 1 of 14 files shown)


Isaac0-dev:
~~known bug: rehosting currently crashes due to memory corruption - I believe it's a use after free.~~ (it was) - fixed

Isaac0-dev:
@coderabbitai review

data/dynos_level.cpp
```diff
@@ -216,6 +216,10 @@ const void *DynOS_Level_GetScript(s32 aLevel) {
     return sDynosLevelScripts[aLevel].mLevelScript;
 }
 
+const void *DynOS_Level_GetVanillaScript(s32 aLevel) {
+    return (const void *) gDynosLevelScriptsOriginal[aLevel];
+}
```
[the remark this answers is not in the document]
Isaac0-dev:
```suggestion
const void *DynOS_Level_GetVanillaScript(s32 aLevel) {
    if (aLevel < LEVEL_MIN || aLevel >= LEVEL_COUNT) {
        return NULL;
    }
    return (const void *) gDynosLevelScriptsOriginal[aLevel];
}
```

data/dynos_mgr_builtin.cpp
```diff
@@ -184,46 +184,13 @@ static const void* sDynosBuiltinScriptPtrs[] = {
...
-    define_level_original(LEVEL_UNKNOWN_35, NULL),
-    define_level_original(LEVEL_TTM, level_ttm_entry),
+const void *gDynosLevelScriptsOriginal[] = {
+#define STUB_LEVEL(_0, _1, _2, _3, _4, _5, _6, _7, _8) NULL,
+#define DEFINE_LEVEL(_0, _1, _2, folder, _4, _5, _6, _7, _8, _9, _10) \
+    level_ ## folder ## _entry,
+#include "levels/level_defines.h"
+#undef STUB_LEVEL
+#undef DEFINE_LEVEL
 };
```
PeachyPeachSM64, line 195: !!! you're missing the entry for LEVEL_NONE, which shifts the whole array
to make it more robust, define STUB_LEVEL and DEFINE_LEVEL like this:
```c
#define STUB_LEVEL(_0, levelNum, _2, _3, _4, _5, _6, _7, _8) \
    [levelNum] = NULL,
#define DEFINE_LEVEL(_0, levelNum, _2, folder, _4, _5, _6, _7, _8, _9, _10) \
    [levelNum] = level_ ## folder ## _entry,
```

src/game/level_update.c
```diff
@@ -1541,30 +1541,33 @@ s32 update_current_play_mode() {
...
+        case 11: return LEVEL_HMC;
+        case 12: return LEVEL_CCM;
+        case 13: return LEVEL_RR;
+        case 14: return LEVEL_BITDW;
+        case 15: return LEVEL_PSS;
+        case 16: return LEVEL_TTC;
+        case 17: return LEVEL_WDW;
+    }
+    return LEVEL_CASTLE_GROUNDS;
+}
```
PeachyPeachSM64, line 1569: please make it a static array

PeachyPeachSM64 (changes requested):
partial review

data/dynos_mgr_builtin.cpp
Isaac0-dev: really good catch, thank you!!

Isaac0-dev:
@coderabbitai full review

PeachyPeachSM64 (approved):

PR #1261, opened 2026-05-23
Add function to enable a fullbright mode (`set_shading_fullbright_enabled`)

AgentXLP:
Useful for the up and coming custom shader pipeline when you want to handle shading entirely yourself without the game's default calculations. Only affects materials with `G_LIGHTING` and not vertex shaded materials, though maybe I could add another function for that if needed


Isaac0-dev:
No human review!!

AgentXLP:
My bad

PR #1268, opened 2026-05-25
fix dx11 compilation issues in levels with noise or alpha dithering

EmeraldLoc:
Fix was quite simple, TEXCOORD1 was defined multiple times so changed the name to SCREENPOS. The reason we don't need to change the input element description is because `TEXCOORD1` wasn't actually set there anyways, so it can just be it's own thing. It has always been set here:

```cpp
append_line(buf, &len, ") {");
append_line(buf, &len, "    PSInput result;");
append_line(buf, &len, "    result.position = position;");
if ((cc.cm.use_alpha && cc.cm.use_dither) || ccf.do_noise) {
    append_line(buf, &len, "    result.screenPos = position;");
}
```

Tested by entering VCUTM, which would crash before but now works as intended


AgentXLP (approved):
Yes

PR #1274, opened 2026-05-26
Fix bad window position code in DirectX

AgentXLP:
Fix bad window position code with DXGI
Tested it and it seems to work good
Managed to cobble this together with what knowledge I have on WIN32 and adjacent APIs as well as using Google
I really prefer OpenGL and SDL2 lol

### DIFF
--- a/src/pc/gfx/gfx_dxgi.cpp
+++ b/src/pc/gfx/gfx_dxgi.cpp
@@ -216,15 +216,26 @@ static void update_screen_settings(void) {
     if (configWindow.fullscreen != dxgi.is_full_screen)
         toggle_borderless_window_full_screen(configWindow.fullscreen);
     if (!dxgi.is_full_screen) {
-        // this code is buggy, and I just simply don't care enough about direct x to fix it
-        // when this is enabled, the window will be placed in the wrong spot...
-        const int screen_width = GetSystemMetrics(SM_CXSCREEN);
-        const int screen_height = GetSystemMetrics(SM_CYSCREEN);
+        // use the current monitor work area for centering and apply the real window styles
+        HMONITOR h_monitor = MonitorFromWindow(dxgi.h_wnd, MONITOR_DEFAULTTONEAREST);
+        MONITORINFO monitor_info;
+        monitor_info.cbSize = sizeof(monitor_info);
+        if (!GetMonitorInfo(h_monitor, &monitor_info)) {
+            return;
+        }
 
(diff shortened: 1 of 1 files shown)


src/pc/gfx/gfx_dxgi.cpp
```diff
@@ -216,16 +216,29 @@ static void update_screen_settings(void) {
...
-        const int ypos = (configWindow.y == WAPI_WIN_CENTERPOS) ? (screen_height - configWindow.h) * 0.5 : configWindow.y;
-        RECT wr = { xpos, ypos, xpos + (int)configWindow.w, ypos + (int)configWindow.h };
-        AdjustWindowRect(&wr, WS_OVERLAPPEDWINDOW, FALSE);
-        SetWindowPos(dxgi.h_wnd, NULL, wr.left, wr.top, wr.right - wr.left, wr.bottom - wr.top, SWP_NOACTIVATE | SWP_NOZORDER);
+        // use the current monitor work area for centering and apply the real window styles
+        HMONITOR h_monitor = MonitorFromWindow(dxgi.h_wnd, MONITOR_DEFAULTTONEAREST);
+        MONITORINFO monitor_info;
+        monitor_info.cbSize = sizeof(monitor_info);
+        if (!GetMonitorInfo(h_monitor, &monitor_info)) {
+            return;
```
Isaac0-dev: can you still backup this with the old screen dimensions if this fails?
Just to make resizing still work.

```suggestion
            monitor_info.rcWork = {
                0, 0,
                GetSystemMetrics(SM_CXSCREEN),
                GetSystemMetrics(SM_CYSCREEN)
            };
```

Isaac0-dev (commented):
make sure you account for DPI awareness, since we seem to have that in this file.
probably just need to check that it behaves correctly across different monitors when resizing the window and such.

AgentXLP:
> make sure you account for DPI awareness, since we seem to have that in this file. probably just need to check that it behaves correctly across different monitors when resizing the window and such.

It already supports DPI awareness since the process is set as DPI aware and there isn't really anything else that needs to be done here, it just works. I have a 4K monitor I can plug in and test on but I'm confident I'm not horribly wrong

EmeraldLoc:
Window position is NOT saving on DirectX11. It DOES save on OpenGL. The saved position from OpenGL works on DirectX 11 fine, but the window pos does not save.

AgentXLP:
I'll have to look into this then

AgentXLP:
Yeah I'll redo this but better at a later point

PR #1275, opened 2026-05-27
Partially fix combiners

Blockyyy:
image
image
image
image

First image is what it looks like in the current version of Coop
Second is what it looks like on this pull request
Third is what it looks like in blender
Fourth Image is what it looks like on emulator (Parallel Launcher, Ares)

- Adds back color overflow (openGL)
- Unfloors the noise (openGL)
- Makes dynos actually convert every texture, fixing the bug where a combiner with multitexture would just be black

Shoutouts to the fast64 renderer

This still isn't perfect but is considerably better than what is currently there


data/dynos_bin_gfx.cpp
```diff
@@ -1500,4 +1501,4 @@ void dynos_smlua_clear_gfx_command_cache() {
     sGfxCommandCache.clear();
 }
 
-}
+}
```
Isaac0-dev: please don't delete the newline at the end

```suggestion
}

```

src/pc/gfx/gfx_opengl.c
```diff
@@ -906,4 +903,4 @@ struct GfxRenderingAPI gfx_opengl_api = {
     gfx_opengl_finish_render,
     gfx_opengl_get_name,
     gfx_opengl_shutdown
-};
+};
```
Isaac0-dev: same here


```suggestion
};

```

Isaac0-dev (changes requested):
just at a glance

Isaac0-dev:
@Blockyyy coderabbit left a comment in "outside diff range" section, just in case you missed it.

Blockyyy:
ah, i did miss it, thanks

AgentXLP:
Excellent
This will be good for v1.6

Blockyyy:
i can probably work on this a bit more then

PeachyPeachSM64:
@Blockyyy 
Is it done? Or should we put it as Draft?

Blockyyy:
there is one more thing i wanted to do so i think this can be set to draft for now until i'm done
i'll see if i can finish today though

Blockyyy:
nevermind, consider it done, if i get back to this i'll just make another pr for the other fixes, this will do for now

PeachyPeachSM64 (approved):

PR #1282, opened 2026-05-30
fix cloned marios taking over nametags and pvp

EmeraldLoc:
Fixes an issue where other Mario objects would set Mario's body state


EmeraldLoc:
I don’t like this, closing as I believe I am doing something very wrong

PR #1284, opened 2026-05-30
Built-in voice chat

fgetcat:
This PR adds a built-in voice and proximity chat system so that users don't have to rely on Mumble.

Disabled for public CoopNet lobbies.


paradox9955:
Cool. But how does the voice chat work? Is it all just sending packets over coopnet? Wouldn't that be problematic?

fgetcat:
yeah it sends packets over coopnet, how is that problematic?

src/pc/network/packets/packet.h
```diff
@@ -7,7 +7,7 @@
 #include <assert.h>
 #include <stdbool.h>
 
-#define PACKET_LENGTH 3000
+#define PACKET_LENGTH 4096
```
kermeow: Is there a reason this needed to be changed?

paradox9955:
> yeah it sends packets over coopnet, how is that problematic?

I assumed it used a lot more bandwidth. But I saw you responded to AgentX about it in discord so nvm

src/pc/voice_chat.c
```diff
@@ -0,0 +1,323 @@
...
+#include "game/level_update.h"
+#include "pc/audio/audio_api.h"
+#include "pc/djui/djui_hud_utils.h"
+#include "pc/network/network.h"
+#include "pc_main.h"
+
+#include <opus/opus.h>
+
+#define INTERNAL_SAMPLE_RATE 48000
+#define SAMPLE_RATE 32000
```
kermeow, line 15: Could the output sample rate be increased for quality sake?

src/pc/network/packets/packet_proxchat.c
```diff
@@ -0,0 +1,27 @@
...
+}
+
+void network_receive_proxchat_frame(struct Packet* p) {
+    u32 frame_size;
+    packet_read(p, &frame_size, sizeof(u32));
+
+    u8 frame_data[frame_size];
+    packet_read(p, frame_data, frame_size);
+
+    proxchat_decode_audio(p->localIndex, frame_data, frame_size);
```
kermeow: If I'm not mistaken, this only works on coopnet.
On the socket network system, clients ask the server to broadcast their packets for them, making this always the host's local index unless you are the host yourself. This works fine if there are only 2 clients, but doesn't behave correctly otherwise.

src/pc/proximity_chat.c
```diff
@@ -0,0 +1,323 @@
...
+    buffer_write(&players[id].audio, num_frames * sizeof(s16), pcm);
+}
+
+void proxchat_mix(s16* out_pcm, u32 num_samples) {
+    num_samples = num_samples * INTERNAL_SAMPLE_RATE / SAMPLE_RATE;
+    s16 mixed[num_samples * 2 /* stereo */] = {};
+
+    // skip over player 0 because thats the client
+    for (s32 i = 1; i < MAX_PLAYERS; i++) {
+        if (!proxchat_is_ingame(i)) continue;
```
kermeow: Skipping players in this way leaves their buffer non-empty if they leave or go to a different level. When you see them again, or a new player takes their index, this creates artifacts.

src/pc/proximity_chat.c
```diff
@@ -0,0 +1,323 @@
...
+        vec3f_normalize(to_target);
+
+        float pan = vec3f_dot(right, to_target);
+
+        float pan_mono = 0.5f + pan * 0.5f * STEREO_SPREAD;
+        float vol_right = pan_mono;
+        float vol_left  = 1 - pan_mono;
+
+        s16 player_pcm[num_samples];
+        u32 n = buffer_read(&players[i].audio, sizeof(player_pcm), player_pcm) / sizeof(s16);
```
kermeow: For incoming audio, it might be a good idea to avoid draining the buffer until at least however many frames have been decoded. Otherwise this is ample opportunity for stuttering.

It might also be a good idea to decode opus packets at this point rather than when received so that PLC and/or FEC can be implemented correctly later. A jitter buffer similar to what is used in other VOIP applications could be useful here.

src/pc/proximity_chat.c
```diff
@@ -0,0 +1,323 @@
...
+    s32 sum = 0, avg;
+    for (u32 i = 0; i < num_samples; i++) {
+        s32 with_gain = samples[i] * (int)configProxchatMicrophoneGain / 100;
+        if (with_gain < -32767) with_gain = -32767;
+        if (with_gain > +32767) with_gain = +32767;
+        samples[i] = with_gain;
+        sum += abs(samples[i]);
+    }
+    avg = sum / num_samples;
+    proxchat_mic_level = 1 - powf(1 - avg / 32767.f, 10);
```
kermeow: The difference is minimal, but why not use RMS here?

src/pc/proximity_chat.c
```diff
@@ -0,0 +1,323 @@
...
+        }
+
+        if (err < 0) {
+            fprintf(stderr, "[PROXCHAT] Failed to initialize player %d: %s\n", i, get_opus_error(err));
+            proxchat_error[i] = PROXCHAT_ERR_FAILED_TO_INITIALIZE;
+        }
+        else proxchat_error[i] = PROXCHAT_ERR_NONE;
+
+        players[i].audio.capacity = FRAME_SIZE * sizeof(s16) * 4;
+        players[i].audio.dynamic = true;
```
kermeow: I do not recommend keeping this enabled. In the case of lag, this will cause voices to become extremely delayed as the buffer will just grow to account for late packets.

Instead, a different way of buffering should probably be considered. See other comment about jitter buffers.

src/pc/network/packets/packet_proxchat.c
```diff
@@ -0,0 +1,27 @@
...
+#include "pc/proximity_chat.h"
+
+void network_send_proxchat_frame(void) {
+    struct Packet p = {};
+
+    u8 frame[4096];
+    u32 size = proxchat_encode_audio(frame, sizeof(frame));
+    if (size == 0) return;
+
+    packet_init(&p, PACKET_PROXCHAT_FRAME, false, PLMT_NONE);
```
kermeow: Since the packets are not reliable or ordered, they need to be timestamped as there is no guarantee they will be received in order. This also ties in with the PLC and jitter buffer mentioned in other comments.

kermeow (commented):
I probably missed a few things. Otherwise, looks good so far! Excited to see if this goes anywhere.

src/pc/network/packets/packet.h
fgetcat: I increased it when I was sending raw PCM data over the network and forgot to change it back

src/pc/voice_chat.c
fgetcat: This is actually the sample rate that's used in `src/pc/audio/audio_sdl.c`, I assume that changing this would break the game's audio system

src/pc/network/packets/packet_proxchat.c
fgetcat: Oh, what's the proper way of getting the index then?

src/pc/proximity_chat.c
fgetcat: Alright, will drain the buffer then

src/pc/proximity_chat.c
fgetcat: I wasn't aware of it until now

src/pc/proximity_chat.c
fgetcat: Ah, I did intend on keeping this disabled but forgot to change it

src/pc/network/packets/packet_proxchat.c
fgetcat: Do I just switch the bool to `true`?

src/pc/network/packets/packet_proxchat.c
kermeow: There isn't one afaik, you'd have to do it yourself as part of the packet data or disable outside of coopnet.

src/pc/network/packets/packet_proxchat.c
kermeow: I wouldn't recommend making them reliable, you should timestamp them yourself.

Blockyyy:
i'm hoping we get a proper lua api for this

src/pc/djui/djui_slider.c
```diff
@@ -70,9 +70,9 @@ static void djui_slider_on_cursor_down(struct DjuiBase* base) {
     f32 x = slider->rect->base.elem.x;
     f32 w = slider->rect->base.elem.width;
     f32 cursorX = gCursorX;
-    cursorX = fmax(cursorX, x);
-    cursorX = fmin(cursorX, x + w);
-    *value = ((cursorX - x) / w) * (max - min) + min;
+    if (cursorX < x) *value = min;
+    else if (cursorX > x + w) *value = max;
+    else *value = ((cursorX - x) / w) * (max - min) + min;
```
fgetcat, line 75: The previous code was prone to floating point errors, which could cause the slider to be stuck at `min+1` or `max-1`. This new code ensures this doesn't happen.

fgetcat:
Right I forgot, @coderabbitai you can review.

src/pc/nametags.c
```diff
@@ -171,6 +172,23 @@ void nametags_render(void) {
...
+        else if (gVoicePlayers[playerIndex].clientMutedState & VOICECHAT_MUTE_LOCAL)    mic = "texture_microphone_muted";
+        else if (gVoicePlayers[playerIndex].talking) mic = "texture_microphone";
+        if (mic) {
+            struct TextureInfo texture;
+            dynos_texture_get(mic, &texture);
+            djui_hud_render_texture_interpolated(&texture,
+                  e->prevPos[0] - prevHalfWidth - 36 *   e->prevScale, prevNametagPosY,   e->prevScale * 2,   e->prevScale * 2,
+                nametag->pos[0] - currHalfWidth - 36 * nametag->scale, currNametagPosY, nametag->scale * 2, nametag->scale * 2
+            );
+        }
```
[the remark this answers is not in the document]
fgetcat: Cannot happen. The strings and texture list that `dynos_texture_get` pulls from are both hardcoded.

Blockyyy:
could you make it so we can choose a microphone in the settings?, it doesn't detect the right one for me

fgetcat:
> could you make it so we can choose a microphone in the settings?, it doesn't detect the right one for me

Yeah I have plans to do that

fgetcat:
Didn't realize that `SDL_GetDefaultAudioInfo` isn't a thing on Ubuntu 22.04

kermeow:
I saw this was no longer a draft but I do believe a few concerns from before have gone unaddressed. Has this been tested yet with more than 2 clients or even outside of the same computer?

fgetcat:
> I saw this was no longer a draft but I do believe a few concerns from before have gone unaddressed. Has this been tested yet with more than 2 clients or even outside of the same computer?

Yes, it has been tested with 5 people in a private lobby and it works well

xLuigiGamerx (commented):
Merge conflicts

xLuigiGamerx (approved):
Merge conflicts resolved

Blockyyy:
image

happens when i try to host with proximity chat, happened like twice and never again, don't know if its related (windows, opengl)

Blockyyy:
would be nice if we could choose the talkign distance for proximity cvhat in server settings or something

Blockyyy:
it would also be cool to have it so rooms muffle your voice and caps have effects on your voice, similarly to that old proximity chat mod, unless mods can apply modifiers to the voice

fgetcat:
> it would also be cool to have it so rooms muffle your voice and caps have effects on your voice, similarly to that old proximity chat mod, unless mods can apply modifiers to the voice

thats outside of the scope of this patch. implementing that needs a bunch of other code and machinery, so it makes sense for it to be a separate patch. i aim to implement the basic functionality in this patch.

EmeraldLoc:
Lota conflicts!

EmeraldLoc:
@coderabbitai full review

PR #1286, opened 2026-05-30
Fix Mario clones stealing bodystate from player Marios

PeachyPeachSM64:
When spawning fake Marios, these objects would steal the bodystate from valid Marios, resulting in nametags rendering in the wrong place and PvP attacks using these fake as origin.

Before:
before

After:
after

Also works with mirror Marios:
mirror

### DIFF
--- a/src/game/mario_misc.c
+++ b/src/game/mario_misc.c
@@ -364,7 +364,7 @@ s8 geo_get_processing_mario_index(struct Object *obj) {
     }
 
     index = obj->oBehParams - 1;
-    return (index >= MAX_PLAYERS) ? -1 : index;
+    return (index < 0 || index >= MAX_PLAYERS) ? -1 : index;
 }
 
 struct MarioState *geo_get_mario_state(void) {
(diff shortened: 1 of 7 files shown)


AgentXLP (approved):
Looks good, better implementation than Emerald's approach ngl (no offense)
Good catch on the cap displaylist incorrect player index type and I suppose the "associated" typo too

src/game/mario_misc.c
```diff
@@ -377,6 +377,27 @@ struct MarioBodyState *geo_get_body_state(void) {
     return &gBodyStates[index];
 }
 
+struct Object *geo_get_mario_object(void) {
```
Isaac0-dev, line 382: Some description would be appreciated here since we're intentionally not returning the object that's actually being rendered.
Could do the same one you put for the header file, or:

```suggestion
// Get the real Mario object represented by the object being rendered
// e.g. if rendering Mirror Mario, return the real Mario object for that player
struct Object *geo_get_mario_object(void) {
```

src/game/mario_misc.c
```diff
@@ -377,6 +377,27 @@ struct MarioBodyState *geo_get_body_state(void) {
...
+    struct Object *obj = gCurGraphNodeProcessingObject;
+    if (obj == NULL) {
+        return NULL;
+    }
+
+    s8 mirrorIndex = geo_get_processing_mirror_mario_index(obj);
```
Isaac0-dev, line 388: Traced it through, just to confirm we're on the same page. 
Mirror Mario seems to be the only type of object that is never spawned as a regular sm64 object, it's just a `GraphNodeObject`.
It can still be safely processed by this function here because regular sm64 objects have their `GraphNodeObject` as the first field. 
However, if `geo_get_processing_mirror_mario_index` ever fails, we immediately read from oob memory and may crash depending on the system.
This is usually safe because we exit early for Mirror Mario, but technically we can't be sure that we're reading from a `struct Object *` until after this condition.
As I say this is safe but if you wanted to be explicit you could take the pointer as a `GraphNodeObject`.. There's other places in `rendering_graph_node.c` which also cast this incorrectly so whatever lol.

Isaac0-dev:
Looks good, just wanted to send these comments so you can consider them @PeachyPeachSM64 ^^

src/game/mario_misc.c
PeachyPeachSM64: `geo_get_processing_object_index` has the same issue (field `globalPlayerIndex`) and we just assume that a pointer outside of the gMirrorMario range is a valid Object pointer. I don't think I'm gonna fix that lol

src/game/mario_misc.c
PeachyPeachSM64: sure, why not

Isaac0-dev (approved):

PR #1301, opened 2026-06-06
Fix Mario knockback

PeachyPeachSM64:
- Restore Mario's "doh" knockback sound on high speeds
- Fix incorrect knockback speed with Heave-ho and Chuckya

As far as I know, only Arena uses custom knockbacks, so enforcing `PVP_ATTACK_KNOCKBACK_ACTION_ARG` shouldn't be an issue with other mods

### DIFF
--- a/mods/arena/arena-player.lua
+++ b/mods/arena/arena-player.lua
@@ -141,7 +141,7 @@ function mario_local_hammer_check(m)
                 vec3f_normalize(vel)
                 vec3f_mul(vel, 80 + 10 * (1 - mario_health_float(cmvictim)))
 
-                set_mario_action(m, ACT_BACKWARD_AIR_KB, 0)
+                set_mario_action(m, ACT_BACKWARD_AIR_KB, PVP_ATTACK_KNOCKBACK_ACTION_ARG)
                 m.invincTimer = 20
                 m.knockbackTimer = 10
                 m.vel.x = vel.x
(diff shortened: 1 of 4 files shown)


EmeraldLoc:
By

> only Arena uses custom knockbacks, so enforcing PVP_ATTACK_KNOCKBACK_ACTION_ARG shouldn't be an issue with other mods

I assume you mean that this is for setting Marios action to backwards or forwards kb and setting velocities manually.

Was this speed change done in the action in vanilla sm64? If not, what's the justification for adding it here and potentially breaking mods?

Actual code changes look fine, mainly just concerned about mods and if this is the correct solution or not.

PeachyPeachSM64:
> I assume you mean that this is for setting Marios action to backwards or forwards kb and setting velocities manually.

yes, if you do `set_mario_action(m, ACT_BACKWARD_AIR_KB, 0)` without `PVP_ATTACK_KNOCKBACK_ACTION_ARG`, any x/z/forward velocity you'll set will be overridden by the vanilla vels.

> Was this speed change done in the action in vanilla sm64? If not, what's the justification for adding it here and potentially breaking mods?

The speed is (and was always) reset every frame in vanilla sm64 in `common_air_knockback_step` (https://github.com/n64decomp/sm64/blob/master/src/game/mario_actions_airborne.c#L1101).

EmeraldLoc (approved):

Isaac0-dev (approved):

PR #1305, opened 2026-06-08
Custom goddard heads using dynos

DorfDork:
ExcellentGamer worked on updating the Goddard head swapper originally from SiSilicon to be compatible with modern repositories. Uses dynos packs to swap between heads with the file format of .gdbin. Heads will also track when player changes characters such as Mario and all between characters to Wario. If no custom .gd head is used, defaults to normal sm64 head.


DorfDork:
Forgot to post some preview images

Screenshot 2026-06-07 225555
Screenshot 2026-06-07 220250

data/dynos_main.cpp
```diff
@@ -45,10 +45,26 @@ void DynOS_Mod_Update() {
...
+        DynOS_Goddard_ModShutdown();
         DynOS_Gfx_ModShutdown();
     }
 }
 
 void DynOS_Mod_Shutdown() {
     sDynosModShutdown = true;
 }
+
+void DynOS_Restart_Reset() {
```
Isaac0-dev: unused function! `dynos_restart_reset` is not used either.

data/dynos_mgr_pack.cpp
```diff
@@ -1,14 +1,147 @@
...
+    BinFile::Close(_File);
+
+    DynOS_Goddard_ActiveMarioHeadBinData() = _Data;
+    DynOS_Goddard_ActiveMarioHeadBinSize() = _Size;
+    sLoadedPath = _Path;
+
+    printf("[DynOS] Goddard: loaded mario_head.gdbin (%d bytes) from %s\n", _Size, _Path.c_str());
+}
+
+extern "C" void DynOS_Goddard_RecomputeActiveMarioHeadBin() {
```
Isaac0-dev: should not be directly building a C function here, we code in C++ for DynOS, and then create a C bridge in `dynos_c.cpp`.
Also, this looks like avoiding the entire DynOS pack system in the first place. This logic belongs in `DynOS_Pack_SetEnabled`!
You can safely only run `DynOS_Goddard_LoadActiveMarioHeadBinIfNeeded` if the active goddard actually changed by having a getter which calculates what the active goddard head is, and checks if any change actually occurred.

src/goddard/skin.c
Isaac0-dev, line 1: Are any of the changes to skin.c necessary? They don't seem relevant.

src/goddard/gd_main.h
```diff
@@ -10,6 +10,7 @@
 // "#define printf(...) /* nothing */" wasn't an option.)
 // This macro is separate from the gd_printf function; one probably
 // forwarded to the other, but it is hard to tell in which direction.
+#if 0
```
Isaac0-dev: what are these changes for?

src/goddard/gd_math.c
```diff
@@ -7,6 +7,7 @@
 #include "gd_types.h"
 #include "macros.h"
 #include "renderer.h"
+#include <stdio.h>
```
Isaac0-dev: didn't it compile without this?

data/dynos_mgr_pack.cpp
```diff
@@ -1,14 +1,147 @@
...
+    return DynOS_Goddard_ActiveMarioHeadBinData();
+}
+
+s32 DynOS_Goddard_GetActiveMarioHeadBinSize() {
+    return DynOS_Goddard_ActiveMarioHeadBinSize();
+}
+
+const SysPath& DynOS_Goddard_GetActiveMarioHeadBin() {
+    return DynOS_Goddard_ActiveMarioHeadBin();
+}
```
Isaac0-dev: please do not have 2 functions that do the exact same thing.
just have a getter function and that's it (for the sake of the C bridge). 
Also, please just use a static file scoped variabledefinition for each, its not very readable in practice.

data/dynos_mgr_pack.cpp
```diff
@@ -1,14 +1,147 @@
 #include <deque>
+#include <stdio.h>
 #include "dynos.cpp.h"
 extern "C" {
 #include "engine/graph_node.h"
 }
 
+#define MAX_CHARACTER_HEADS 16
+static const char* sCharacterHeadNames[MAX_CHARACTER_HEADS] = {
```
Isaac0-dev: did you have any reason for allocating 16 slots for characters?
This array should not allocate more than it uses, don't predefine the number of slots.
Also you can use C++ std data structures here, perhaps `std::array`, and remove the `NULL` entry.
Also add a static assertion to ensure the size of `sCharacterHeadNames` is the same as `CT_MAX`.
That'll warn anyone who tries to add a new character without changing this.

data/dynos_mgr_pack.cpp
```diff
@@ -1,14 +1,147 @@
...
+    
+    // Get the character index from save file
+    s32 charIndex = (configPlayerModel >= CT_MAX) ? CT_MARIO : configPlayerModel;
+    
+    // First try to find a character-specific head, then fall back to mario_head
+    for (auto& _Pack : DynosPacks()) {
+        if (!_Pack.mEnabled) { continue; }
+        
+        // Try character-specific head first
+        if (charIndex > 0 && charIndex < MAX_CHARACTER_HEADS && !_Pack.mGoddardCharacterHeadBins[charIndex].empty()) {
```
Isaac0-dev: `charIndex` should be unsigned, because `configPlayerModel` is. then, remove `charIndex > 0 && charIndex < MAX_CHARACTER_HEADS && ` because that's guaranteed to be true.

src/pc/configfile.c
```diff
@@ -963,5 +965,8 @@ void configfile_save(const char *filename) {
         functionOptions[i].write(file);
     }
 
+    // Recompute Goddard head for the new character
+    DynOS_Goddard_RecomputeActiveMarioHeadBin();
```
Isaac0-dev: this is almost certainly the wrong place to do this.

data/dynos_bin_behavior.cpp
```diff
@@ -1913,8 +1913,23 @@ s64 DynOS_Bhv_ParseBehaviorScriptConstants(const String &_Arg, bool *found) {
...
+            }
+            best = node;
+        }
+    }
+    return best;
+}
+
 static BehaviorScript ParseBehaviorScriptSymbolArgInternal(GfxData *aGfxData, DataNode<BehaviorScript> *aNode, u64 &aTokenIndex, bool *found) {
     String _Arg = aNode->mTokens[aTokenIndex++];
+    u64 _ModelIdentifier = aNode->mDataIdentifier;
```
Isaac0-dev: this is unused

data/dynos_bin_behavior.cpp
Isaac0-dev, line 1: unrelated and unused changes here

Isaac0-dev (changes requested):
The DynOS goddard bin code should not be in `dynos_mgr_pack.cpp`, it should be in it's own file, like `dynos_bin_goddard.cpp`.

Shouldn't be using the DynOS name in names for symbols outside DynOS.

Why are you removing print statements in the goddard code?

data/dynos_main.cpp
ExcellentGamer: This function was used in the Render96dx repo, handled "restarting" the game differently than coop, will remove.

data/dynos_mgr_pack.cpp
ExcellentGamer: A lot of this system was made somewhat blindly, will adjust the code to actually use more actual DynOS systems! The C++ code note is important and I intend to further note that upon my future changes if I make them.

src/goddard/skin.c
ExcellentGamer: iirc I had a config to disable the shine in the render96dx repo, likely not

xLuigiGamerx (commented):
Merge conflicts

data/dynos_bin_goddard.cpp
Isaac0-dev, line 1: This code is from inside a dynos file, so symbols should use the dynos naming style!

data/dynos_bin_goddard.cpp
```diff
@@ -0,0 +1,149 @@
...
+    "mario_head.gdbin",
+    "luigi_head.gdbin",
+    "toad_head.gdbin",
+    "waluigi_head.gdbin",
+    "wario_head.gdbin",
+};
+
+static_assert(sCharacterHeadNames.size() == CT_MAX, "sCharacterHeadNames size must match CT_MAX");
+
+static SysPath& Goddard_ActiveMarioHeadBin() {
```
Isaac0-dev: please separate the reference symbol from the data type with a space.

data/dynos_bin_goddard.cpp
```diff
@@ -0,0 +1,149 @@
...
+    return Goddard_ActiveMarioHeadBinData();
+}
+
+s32 goddard_get_active_mario_head_bin_size() {
+    return Goddard_ActiveMarioHeadBinSize();
+}
+
+const SysPath& goddard_get_active_mario_head_bin() {
+    return Goddard_ActiveMarioHeadBin();
+}
```
Isaac0-dev: why do these extra getters still exist?

data/dynos_bin_goddard.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    s32 _Size = _File->Size();
+    if (_Size <= 0) {
+        BinFile::Close(_File);
+        printf("[DynOS] Goddard: mario_head.gdbin is empty: %s\n", _Path.c_str());
+        return;
+    }
+
+    u8* _Data = (u8*) calloc(_Size, 1);
```
Isaac0-dev: don't calloc that, use malloc (calloc resets memory to 0, which is wasted compute because you overwrite that immediately):

```suggestion
    u8* _Data = (u8*) malloc(_Size);
```

src/goddard/renderer.c
```diff
@@ -913,6 +989,12 @@ void *gd_allocblock(u32 size) {
 
 /* 24A318 -> 24A3E8 */
 void *gd_malloc(u32 size, u8 perm) {
+    /*C MEM*/
+    size = ALIGN(size, 8);
+    sAllocMemory += size;
+    return malloc(size);
```
Isaac0-dev: this completely disables the code below!!
If that's intentional, should explain what is happening here in a comment, and delete or comment out the code below.

Isaac0-dev (changes requested):

Isaac0-dev:
Marking as draft as this is still pending changes.

data/dynos_bin_goddard.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (!_GoddardDir) { return; }
+
+    struct dirent *_GoddardEnt = NULL;
+    while ((_GoddardEnt = readdir(_GoddardDir)) != NULL) {
+        s32 length = strlen(_GoddardEnt->d_name);
+
+        // check for goddard heads
+        if (length > 3 && !strncmp(&_GoddardEnt->d_name[length - 3], ".gd", 3)) {
+            std::string _HeadName(_GoddardEnt->d_name, length - 3);
+            aPack->mGoddardHeads[_HeadName] = fstring("%s/%s", _GoddardFolder.c_str(), _GoddardEnt->d_name);
```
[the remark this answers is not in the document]
DorfDork: the format is .gd now, .gdbin was removed

Isaac0-dev:
conflicts

src/goddard/dynlist_proc.c
```diff
@@ -1003,11 +1003,17 @@ void chk_shapegen(struct ObjShape *shape) {
...
             }
 
-            if (vtxdata->count >= VTX_BUF_SIZE) {
+            if (vtxdata->count == 0) {
+                fatal_printf("shapegen() no vertices");
+            }
+
+            // Avoid fixed-size vertex staging buffer; allocate based on vertex count.
+            // Keep a sanity cap to prevent pathological allocations.
+            if (vtxdata->count > 200000) {
```
Isaac0-dev: this should be a constant rather than magic value. Use `VTX_BUF_SIZE`, and if you need it this large, update the constant.

src/goddard/renderer.c
```diff
@@ -110,9 +117,11 @@ static OSContPad sPrevFrameCont[4]; // @ 801BAE88
 static u8 D_801BAEA0;
 static struct ObjGadget *sTimerGadgets[GD_NUM_TIMERS]; // @ 801BAEA8
 static u32 D_801BAF28;                                 // RAM addr offset?
-static s16 sTriangleBuf[13][8];                          // [[s16; 8]; 13]? vert indices?
+static s16 sTriangleBuf[GD_CFG_TRIANGLE_BUF_SIZE][8];
+static u32 unref_801bb000[3];
```
Isaac0-dev: what's this for? it doesn't seem to be referenced. please delete it unless there is a reason it is here.

src/goddard/renderer.c
```diff
@@ -110,9 +117,11 @@ static OSContPad sPrevFrameCont[4]; // @ 801BAE88
...
 static u8 *sMemBlockPoolBase; // @ 801BB00C
 static u32 sAllocMemory;      // @ 801BB010; malloc-ed bytes
+static u32 unref_801bb014;
```
Isaac0-dev: same here, and for all added `unref_` symbols. they are not used.

src/goddard/renderer.c
```diff
@@ -199,19 +215,27 @@ static struct ObjView *sMenuView = NULL; // @ 801A86E8
 static u32 sItemsInMenu = 0;             // @ 801A86EC
 static s32 sDebugViewsCount = 0;               // number of elements in the sDebugViews array
 static s32 sCurrDebugViewIndex = 0;             // @ 801A86F4; timing activate cool down counter?
+static u32 unref_801a86F8 = 0;
 static struct GdDisplayList *sCurrentGdDl = NULL; // @ 801A86FC
 static u32 sGdDlCount = 0;                        // @ 801A8700
-UNUSED static struct DynListBankInfo sDynLists[] = {     // @ 801A8704
+static struct DynListBankInfo sDynLists[] = {     // @ 801A8704
```
Isaac0-dev: why are we removing the `UNUSED` flag? it is literally unused.
same for all added unused stuff below.

data/dynos_mgr_pack.cpp
```diff
@@ -1,12 +1,13 @@
...
+#include <stdio.h>
 #include <regex>
 #include <charconv>
 #include "dynos.cpp.h"
 extern "C" {
 #include "engine/graph_node.h"
 }
 
-static std::deque<PackData>& DynosPacks() {
+std::deque<PackData>& DynosPacks() {
```
Isaac0-dev: this should be remaining static!

data/dynos_mgr_pack.cpp
```diff
@@ -86,6 +87,8 @@ static void ScanPackBins(struct PackData* aPack) {
             }
         }
     }
+
+    DynOS_Goddard_ScanPack(aPack);
```
Isaac0-dev: no, do not read the entire pack directory twice! remove `DynOS_Goddard_ScanPack` and shift this code up into the loop above, because you shouldn't be reading the directory twice in a row.

data/dynos_bin_goddard.cpp
```diff
@@ -0,0 +1,139 @@
...
+    if (_CharacterIndex >= CT_MAX) { _CharacterIndex = CT_MARIO; }
+
+    SysPath _HeadPath = DynOS_Goddard_FindHead(sHeadOverride);
+    if (_HeadPath.empty()) { _HeadPath = DynOS_Goddard_FindHead(sCharacterHeadNames[_CharacterIndex]); }
+    if (_HeadPath.empty()) { _HeadPath = DynOS_Goddard_FindHead(sCharacterHeadNames[CT_MARIO]); }
+    if (_HeadPath == sHeadPath) { return; }
+
+    sHeadPath = _HeadPath;
+    DynOS_Goddard_LoadHead();
+}
```
Isaac0-dev: This method of updating the active goddard model is not ideal. 
Instead of constantly checking if something changed, only check if something changed when a DynOS pack is updated via `DynOS_Pack_SetEnabled`, and if a mod is activated/deactivated by simply adding and removing the Goddard data. 
Checking if a model change occurred every frame is inefficient.

Isaac0-dev (changes requested):
this is definitely better than last time, but there are some major changes needed here.
- The DynOS integration is still not complete. currently you're doing a lot of code manually instead of adopting the internal DynOS structures, so this is kinda a standalone unit which hooks into DynOS instead. 
To effectively use DynOS, please use a similar method to the one I used in `data/dynos_mgr_audio.cpp`.
- Why did you add so much unused code, and in some cases removed the `UNUSED` flag from code that is still unused?

Isaac0-dev:
don't forget to expand the collapsed comments from coderabbit, it has a real, important observation

src/goddard/renderer.c
DorfDork: backport from render96ex which is using an older code base, will fix

data/dynos_mgr_pack.cpp
DorfDork: whoops

data/dynos_bin_goddard.cpp
DorfDork: I agree, will update

DorfDork:
@coderabbitai try review

PR #1318, opened 2026-06-18
Paintings: reduce entries with macros

Cooliokid956:
Split from #1310 
There is no need to `memcpy` singular structs, simple assignment will do

### DIFF
--- a/src/game/paintings.h
+++ b/src/game/paintings.h
@@ -48,6 +48,11 @@ typedef struct {
 	signed char	n[3];	/* normal */
 } Vtx_Interp;
 
+#define DEF_PAINTING(name, ...) \
+struct Painting name = __VA_ARGS__; \
+const struct Painting default_ ## name = __VA_ARGS__
+
+#define RESTORE_PAINTING(name) name = default_ ## name
 struct Painting
 {
     s16 id;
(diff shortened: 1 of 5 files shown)


src/game/paintings.h
```diff
@@ -48,6 +48,11 @@ typedef struct {
 	signed char	n[3];	/* normal */
 } Vtx_Interp;
 
+#define DEF_PAINTING(name, ...) \
+struct Painting name = __VA_ARGS__; \
+const struct Painting default_ ## name = __VA_ARGS__
+
+#define RESTORE_PAINTING(name) name = default_ ## name
 struct Painting
```
PeachyPeachSM64: prefer `DEFINE_PAINTING` over `DEF_PAINTING` (like `DEFINE_LEVEL`, `DEFINE_COURSE`, etc...)
add a newline between `RESTORE_PAINTING` and the struct definition

PeachyPeachSM64 (changes requested):

src/game/paintings.h
Cooliokid956: OK

PeachyPeachSM64 (approved):

PR #1329, opened 2026-06-24
Remove NPOT texture check in DJUI

eros71-dev:
Doesn't seem to crash anymore, not needed, tested and NPOT textures work under both renderers (OpenGL and DX11)


xLuigiGamerx (commented):
It's not just about whether it crashes or not, the engine just doesn't handle NPOT well, did you confirm that functions like djui_hud_render_texture_tile work correctly with it and with NPOT tilewidths and tileheights ans such

eros71-dev:
> It's not just about whether it crashes or not, the engine just doesn't handle NPOT well, did you confirm that functions like djui_hud_render_texture_tile work correctly with it and with NPOT tilewidths and tileheights ans such

You're right, let me check that

eros71-dev:
https://github.com/user-attachments/assets/04396430-43b2-455b-a24d-ef633c77d559


https://github.com/user-attachments/assets/fa64a498-060c-4f4d-97c4-153bbb8fe578


A couple of examples of the interpolation and rendering functions working with NPOT textures, 201x212 for the FNAF logo and 200x200 for the heart icon. Works in tiled and regular textures, interpolated or not. @AgentXLP helped me out with the heart test since the other one was a bit hard to see and I wanted to make sure I was doing it right as well as checking with other setups. 

Should be good to go now and merge, but feel free to test under another OS maybe just in case.

src/pc/djui/djui_hud_utils.c
```diff
@@ -755,11 +755,6 @@ static inline bool is_power_of_two(u32 n) {
 }
 
 static void djui_hud_render_texture_raw(const Texture* texture, u32 width, u32 height, u8 fmt, u8 siz, f32 x, f32 y, f32 scaleW, f32 scaleH, struct InterpHud *interp) {
-    if (!is_power_of_two(width) || !is_power_of_two(height)) {
```
AgentXLP, line 829: Make sure to delete the `inline bool is_power_of_two` function above and I'll approve this

AgentXLP (commented):
Perfect except one small thing

AgentXLP (approved):

Isaac0-dev:
its great to hear that it doesn't crash, but why did it stop crashing?
why did it crash in the first place?
I think we should probably understand the answer to those questions before merging this.

anzz1:
Has the fundamental rule of there being a performance penalty for using NPOT textures changed with modern GPUs?
If not, I think the check should be kept in so mod creators are forced to follow the correct practice.

Also, is there any need for anyone to actually use NPOT textures? It can help later development if you know that textures are always POT and can base logic on that assumption - the very reason why back in the day GPU's *required* POT textures, while later on NPOT support was bolted on but had performance penalties since most of the pipeline still relied on POT logic, so NPOT textures were simply converted to POT in the pipeline. Not sure if this still is the case or relevant.

It reminds me of GoldSrc engine, there is certain samplerate used for audio samples which are then mixed by the engine, but if you put in audio with wrong sample rate, it needs to convert it on the fly causing a performance penalty. But the only thing it does is log in debug console "warning: sample X has wrong samplerate and is causing runtime samplerate conversion!". Of course, mod makers don't notice/understand this and use sounds without pre-converting the samplerate now causing performance penalties everywhere.

If the design simply went with "fail fast", and didn't play wrong samplerate sounds at all, then every mod maker would have been forced to do the right thing and convert the samplerate beforehand, and no performance penalty would be caused.

EmeraldLoc:
Hey, you know what render api is the problem here? OpenGL! Because it's slow, crappy, and garbage. I don't know why you like it so much when its old and outdated!

Nonetheless, non power of 2 textures were figured out a LOOOOOONG time ago, on modern gpus there is no performance penalty, especially on modern render apis.

What Isaac said is very true, but idk what you're yapping about anzz1

anzz1:
That was my question exactly, whether the performance penalty of NPOT textures is still the case in "modern" GPUs from the last 10-15 years. I guess not then.

And OpenGL is perfect for the project for its wide compatibility, ease-of-use and maintainability, and keeping the game released in 1996 compatible with low-end hardware to serve as large userbase as possible. The very reason why so many people like it and why its not going anywhere.

PR #1337, opened 2026-07-09
Standardize the window manager to always use sdl2 leaving the backend specific code separate

xLuigiGamerx:
This pr basically standardizes the window manager so it's always using SDL2 rather than switching based on the backend like dxgi, this makes handling window events and keyboard and mouse input much more convenient and less confusing rather than having to figure out how to use dxgi for those.
This also fixes certain issues with dxgi such as the window titlebar or edges hiding the mouse


src/pc/gfx/gfx_window_manager.c
```diff
@@ -106,101 +88,37 @@ static void gfx_sdl_reset_dimension_and_pos(void) {
...
-    wnd = SDL_CreateWindow(
-        window_title,
-        xpos, ypos, configWindow.w, configWindow.h,
-        SDL_WINDOW_OPENGL | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE
-    );
-    ctx = SDL_GL_CreateContext(wnd);
+    currBackend = configGraphicsBackend;
 
-    gfx_sdl_set_vsync(configWindow.vsync);
+    sBackends[currBackend]->init(window_title);
```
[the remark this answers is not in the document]
xLuigiGamerx: GAPI_MAX is never allowed since it's just the count and it's not actually a backend itself

src/pc/gfx/gfx_window_manager.c
```diff
@@ -43,16 +22,14 @@
 #include "pc/mods/mod_import.h"
 #include "pc/rom_checker.h"
 
-#ifndef GL_MAX_SAMPLES
-#define GL_MAX_SAMPLES 0x8D57
-#endif
+static struct GfxBackendAPI *sBackends[GFX_BACKEND_COUNT] = {
+    [GFX_BACKEND_OPENGL] = &gfx_gl,
+    [GFX_BACKEND_DIRECTX] = &gfx_dxgi,
+};
```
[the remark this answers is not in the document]
xLuigiGamerx: Addressed

src/pc/gfx/gfx_dummy.c
```diff
@@ -199,28 +147,15 @@ static const char* gfx_dummy_renderer_get_name(void) {
...
-    gfx_dummy_wm_get_clipboard_text,
-    gfx_dummy_wm_set_clipboard_text,
-    gfx_dummy_wm_set_cursor_visible,
-    gfx_dummy_wm_delay,
-    gfx_dummy_wm_get_max_msaa,
-    gfx_dummy_wm_set_window_title,
-    gfx_dummy_wm_reset_window_title,
-    gfx_dummy_wm_has_focus
+    gfx_dummy_wm_get_max_msaa
 };
```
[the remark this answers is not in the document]
xLuigiGamerx: Addressed

src/pc/gfx/gfx_window_manager.h
```diff
@@ -0,0 +1,63 @@
...
+    GFX_BACKEND_OPENGL,
+#if defined(_WIN32)
+    GFX_BACKEND_DIRECTX,
+#endif
+    GFX_BACKEND_DUMMY,
+    GFX_BACKEND_COUNT,
+    GFX_BACKEND_MAX = GFX_BACKEND_COUNT - 1,
+};
+
+struct GfxBackendAPI {
```
EmeraldLoc: I heavily dislike this. It would make more sense for this to be integrated with the render api, since that is what this really is apart of. This is a very confusing naming scheme that is not at all proper. Either:

1. Name it to align with the window manager
2. Add the funcs to the render api

src/pc/gfx/gfx_window_opengl.c
EmeraldLoc, line 1: The name of this whole thing is frankly abysmal. Either

1. Align it with the window, so something akin to `gfx_window_gl`, and for future apis, `gfx_window_metal`, `gfx_window_vulkan`, etc. Make dxgi `gfx_window_dxgi`
2. Throw it in the same spot as the render api

src/pc/gfx/gfx_dummy.c
```diff
@@ -199,28 +147,15 @@ static const char* gfx_dummy_renderer_get_name(void) {
 static void gfx_dummy_renderer_shutdown(void) {
 }
 
-struct GfxWindowManagerAPI gfx_dummy_wm_api = {
+struct GfxBackendAPI gfx_dummy_backend = {
```
EmeraldLoc: this whole name is very unclear as stated above

src/pc/pc_main.c
```diff
@@ -222,30 +221,27 @@ static void select_graphics_backend(void) {
     }
 
 #if defined(_WIN32)
-    if (configGraphicsBackend == GAPI_GL && !gfx_sdl_check_opengl_compatibility()) {
-        configGraphicsBackend = GAPI_D3D11;
+    if (configGraphicsBackend == GFX_BACKEND_OPENGL && !gfx_gl_check_compatibility()) {
+        configGraphicsBackend = GFX_BACKEND_DIRECTX;
     }
 #endif
     int backend = configGraphicsBackend;
```
EmeraldLoc: Use `enum GfxBackend` here instead

src/pc/djui/djui_panel_display.c
```diff
@@ -72,12 +72,12 @@ void djui_panel_display_create(struct DjuiBase* caller) {
         djui_checkbox_create(body, DLANG(DISPLAY, SHOW_FPS), &configShowFPS, NULL);
         djui_checkbox_create(body, DLANG(DISPLAY, VSYNC), &configWindow.vsync, djui_panel_display_apply);
 
-        if (GAPI_MAX > 1) {
-            char* gfxBackendChoices[2] = {
+        if (GFX_BACKEND_MAX > 1) {
```
EmeraldLoc: Only run this code if you are on Windows, please wrap in ifdef so linux and macOS users don't see a compile error!

src/pc/gfx/gfx_window_opengl.c
```diff
@@ -0,0 +1,173 @@
...
+
+    int xpos = (configWindow.x == WAPI_WIN_CENTERPOS) ? SDL_WINDOWPOS_CENTERED : configWindow.x;
+    int ypos = (configWindow.y == WAPI_WIN_CENTERPOS) ? SDL_WINDOWPOS_CENTERED : configWindow.y;
+
+    wnd = SDL_CreateWindow(
+        window_title,
+        xpos, ypos, configWindow.w, configWindow.h,
+        SDL_WINDOW_OPENGL | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE
+    );
+    ctx = SDL_GL_CreateContext(wnd);
```
EmeraldLoc: Where'd the shutdown func go that cleans this up? While not strictly necessary, it is going backwards, not forwards.

src/pc/gfx/gfx_window_manager.c
```diff
@@ -0,0 +1,320 @@
...
+    SDL_SetWindowSize(wnd, configWindow.w, configWindow.h);
+    SDL_SetWindowPosition(wnd, xpos, ypos);
+}
+
+void gfx_wm_init(const char *window_title) {
+    if (gCLIOpts.headless) { return; }
+#if defined(_WIN32)
+    SetProcessDPIAware();
+#endif
+
```
EmeraldLoc, line 104: You've removed `SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, "0");`! You also removed `SDL_Init(SDL_INIT_VIDEO);`! Best to readd them

src/pc/pc_main.c
```diff
@@ -222,30 +221,27 @@ static void select_graphics_backend(void) {
...
-        case GAPI_D3D11:
-            gWindowApi = &gfx_dxgi;
+        case GFX_BACKEND_DIRECTX:
             gRenderApi = &gfx_direct3d11_api;
             gAudioApi  = &audio_sdl;
             break;
 #endif
         default:
-            gWindowApi = &gfx_sdl;
             gRenderApi = &gfx_opengl_api;
```
EmeraldLoc: This can crash if the config is set to gfx dummy and the window manager doesn't initialize because of it! Either make default use the dummy renderer or add an explicit switch case for for the dummy renderer

EmeraldLoc (changes requested):

src/pc/djui/djui_panel_display.c
xLuigiGamerx: This is already protected bro, MAX is 1 when directx doesn't exist

src/pc/pc_main.c
xLuigiGamerx: Can never happen, gCLIOpts.backend is manually set to a backend based on the string and dummy isn't an option

src/pc/djui/djui_panel_display.c
EmeraldLoc: > src/pc/djui/djui_panel_display.c:78:17: warning: excess elements in
      array initializer [-Wexcess-initializers]
   78 |                 "DirectX 11"

Ok you can find another way to solve the warning then :D

src/pc/pc_main.c
EmeraldLoc: I wasn't talking about the cli opts, I was talking about the config option and editing it manually there, either way, fixed now so thanks

src/pc/djui/djui_panel_display.c
```diff
@@ -64,20 +64,20 @@ void djui_panel_display_create(struct DjuiBase* caller) {
...
 
     {
         djui_checkbox_create(body, DLANG(DISPLAY, FULLSCREEN), &configWindow.fullscreen, djui_panel_display_apply);
         djui_checkbox_create(body, DLANG(DISPLAY, FORCE_4BY3), &configForce4By3, djui_panel_display_apply);
         djui_checkbox_create(body, DLANG(DISPLAY, SHOW_FPS), &configShowFPS, NULL);
         djui_checkbox_create(body, DLANG(DISPLAY, VSYNC), &configWindow.vsync, djui_panel_display_apply);
 
-        if (GAPI_MAX > 1) {
-            char* gfxBackendChoices[2] = {
+        if (GFX_WINDOW_BACKEND_MAX > 1) {
```
EmeraldLoc, line 75: This is still broken and still gives a compilation warning! It doesn't matter that `GFX_WINDOW_BACKEND_MAX` will always be less than or equal to 1, the compiler will stile compile the below expression and return

```
src/pc/djui/djui_panel_display.c:78:17: warning: excess elements in
      array initializer [-Wexcess-initializers]
   78 |                 "DirectX 11"
```

EmeraldLoc (changes requested):

EmeraldLoc (approved):

PeachyPeachSM64 (approved):

PR #1343, opened 2026-07-12
More control on hook_chat_command

ProfeJavix:
Sometimes you may need a dynamic chat command, but currently there's no way of disabling it once it is hooked. In order to allow it:
-Now the third param can be the usual function, but also a function table with fields `callback` and `enabled`.
-`callback` works as the hook function, while `enabled` returns a boolean to check if the command should be triggered and displayed in the command list
-The documentation and autogen were updated with these changes


EmeraldLoc:
I understand why it is what it is, but Im not a fan of using a table to have a callback and boolean. I think I would prefer having a `enable_chat_command` func and `disable_chat_command` func (or just a `set_chat_command_enabled` func) and have `hook_chat_command` return an index to that chat command to be used inside those functions. But someone else may provide input

ProfeJavix:
> I understand why it is what it is, but Im not a fan of using a table to have a callback and boolean. I think I would prefer having a `enable_chat_command` func and `disable_chat_command` func (or just a `set_chat_command_enabled` func) and have `hook_chat_command` return an index to that chat command to be used inside those functions. But someone else may provide input

Yeah, you've got a point. I guess it's more practical with `set_chat_command_enabled` since there aren't that many cases where the toggle is needed. Though, having it in a single place can help to keep the code clean.
Anyway, if needed, I have no problem on changing it myself👌

PR #1346, opened 2026-07-18
Give more control over dialog type and sound

PeachyPeachSM64:
- Allow mods to assign a dialog type (boss start, boss stop, race, star sound) to any dialog id.
- Unhardcode default dialog ids to types to use `gBehaviorValues.dialogs` instead.
- Pass `dialogId` as second parameter of hook `HOOK_DIALOG_SOUND`.
- Allow `HOOK_DIALOG_SOUND` to return a sound id instead of a speaker to play any sound.


src/audio/external.c
```diff
@@ -2452,15 +2452,21 @@ void set_sound_moving_speed(u8 bank, u8 speed) {
...
-    if (!IS_VALID_VANILLA_DIALOG(dialogID)) {
-        dialogID = 0;
+    if (IS_VALID_VANILLA_DIALOG(dialogID)) {
+        speaker = sDialogSpeaker[dialogID];
     }
 
-    speaker = sDialogSpeaker[dialogID];
-    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, &speaker);
-    if (speaker < DS_MAX && speaker != 0xff) {
+    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, dialogID, &speaker);
```
Isaac0-dev, line 2461: Issue spotted by coderabbitai (https://github.com/coop-deluxe/sm64coopdx/pull/1346/changes#r3608375706), but my proposed solution.

```suggestion
    s32 hookSpeakerValue = speaker;
    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, dialogID, &hookSpeakerValue);
    if (hookSpeakerValue >= 0) {
        speaker = hookSpeakerValue;
    }
```

src/audio/external.c
```diff
@@ -2459,8 +2459,15 @@ void play_dialog_sound(s32 dialogID) {
...
+
+    // Hook returned a sound id
+    if (speaker > 0xFF) {
+        play_sound(speaker, gGlobalSoundSource);
+    }
+
+    // Hook returned a speaker id
+    else if (speaker < DS_MAX) {
         play_sound(sDialogSpeakerVoice[speaker], gGlobalSoundSource);
```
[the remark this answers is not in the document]
Isaac0-dev: Agreed on the issue, I disagree on the solution. 
The first `play_sound` call shouldn't see the negative value, so fix the issue immediately after `smlua_call_event_hooks`.


```C
    s32 hookSpeakerValue = speaker;
    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, dialogID, &hookSpeakerValue);
    if (hookSpeakerValue >= 0) {
        speaker = hookSpeakerValue;
    }
```

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -250,6 +290,64 @@ s32 smlua_text_utils_allocate_dialog(void) {
...
+    dialogIdType->dialogType = dialogType;
+}
+
+void smlua_text_utils_dialog_reset_type(enum DialogId dialogId) {
+    if (!sDialogIdTypes) {
+        return;
+    }
+
+    for (u32 i = 0; i < sDialogIdTypes->count; i++) {
+        const struct DialogIdType *dialogIdType = sDialogIdTypes->buffer[i];
```
Isaac0-dev, line 343: why not do this for consistency

```suggestion
    growing_array_for_each_(sDialogIdTypes, struct DialogIdType, dialogIdType) {
```

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -250,6 +290,64 @@ s32 smlua_text_utils_allocate_dialog(void) {
...
+        if (dialogIdType->dialogId == dialogId) {
+            growing_array_swap_and_pop_index(sDialogIdTypes, i);
+            return;
+        }
+    }
+}
```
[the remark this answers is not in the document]
Isaac0-dev: `smlua_text_utils_reset_all` doesn't leak, but agreed that `smlua_text_utils_dialog_reset_type` does.
should add `sDialogIdTypes->free(dialogIdType);` before the `growing_array_swap_and_pop_index` call.

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -250,6 +290,64 @@ s32 smlua_text_utils_allocate_dialog(void) {
...
+
+void smlua_text_utils_dialog_reset_type(enum DialogId dialogId) {
+    if (!sDialogIdTypes) {
+        return;
+    }
+
+    for (u32 i = 0; i < sDialogIdTypes->count; i++) {
+        const struct DialogIdType *dialogIdType = sDialogIdTypes->buffer[i];
+        if (dialogIdType->dialogId == dialogId) {
+            growing_array_swap_and_pop_index(sDialogIdTypes, i);
```
Isaac0-dev, line 345: memory leak unless you free here, as spotted by coderabbit.

```suggestion
            sDialogIdTypes->free(dialogIdType);
            growing_array_swap_and_pop_index(sDialogIdTypes, i);
```

Isaac0-dev (changes requested):
besides what coderabbit already picked up on, this is pretty good.

src/audio/external.c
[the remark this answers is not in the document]
PeachyPeachSM64: > Agreed on the issue, I disagree on the solution. The first `play_sound` call shouldn't see the negative value, so fix the issue immediately after `smlua_call_event_hooks`.
> 
> ```c
>     s32 hookSpeakerValue = speaker;
>     smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, dialogID, &hookSpeakerValue);
>     if (hookSpeakerValue >= 0) {
>         speaker = hookSpeakerValue;
>     }
> ```

The problem is that, since soundBits are s32, sounds with bank 8 and above appear as negative (for example, SOUND_OBJ2_BOWSER_ROAR: 0x90040080 == -1878785919)
`if (speaker < 0 || speaker > 0xFF)` seems to be the best solution to handle sounds

src/audio/external.c
PeachyPeachSM64: see: https://github.com/coop-deluxe/sm64coopdx/pull/1346#discussion_r3632658007

src/pc/lua/utils/smlua_text_utils.c
PeachyPeachSM64: because growing_array_swap_and_pop_index that requires the index

src/pc/lua/utils/smlua_text_utils.c
PeachyPeachSM64: Not a leak. The popped element memory will be reused at the next growing array alloc. Even if count decreases, all elements up to capacity are freed when the growing array is freed.

src/pc/lua/utils/smlua_text_utils.c
Isaac0-dev: fair enough

src/audio/external.c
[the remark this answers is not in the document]
Isaac0-dev: @PeachyPeachSM64 good point, agreed.

src/pc/lua/utils/smlua_text_utils.c
Isaac0-dev: true, that's the point of a growing array.

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -163,6 +200,9 @@ void smlua_text_utils_reset_all(void) {
     } else {
         dialog_table_shutdown();
     }
+
+    // Dialog types
```
EmeraldLoc, line 204: Uber nitpicky, is this comment really necessary? Like, nah, really???, it's dialog types?!?!? No way!!

src/audio/external.c
```diff
@@ -2452,15 +2452,21 @@ void set_sound_moving_speed(u8 bank, u8 speed) {
...
+    if (IS_VALID_VANILLA_DIALOG(dialogID)) {
+        speaker = sDialogSpeaker[dialogID];
     }
 
-    speaker = sDialogSpeaker[dialogID];
-    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, &speaker);
-    if (speaker < DS_MAX && speaker != 0xff) {
+    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, dialogID, &speaker);
+
+    // Hook returned a sound id
```
EmeraldLoc, line 2463: Is the convention to start comments uppercase or lowercase? Dk if it matters lol

src/audio/external.c
```diff
@@ -2452,15 +2452,21 @@ void set_sound_moving_speed(u8 bank, u8 speed) {
...
+    if (speaker < 0 || speaker > 0xFF) {
+        play_sound(speaker, gGlobalSoundSource);
+    }
+
+    // Hook returned a speaker id
```
EmeraldLoc, line 2468: Same here

src/pc/lua/utils/smlua_text_utils.c
```diff
@@ -17,7 +18,43 @@ extern s32 gInGameLanguage;
...
 
-#define INVALID_COURSE_NUM(courseNum) (smlua_level_util_get_info_from_course_num(courseNum) == NULL && !COURSE_IS_VALID_COURSE(courseNum))
+  //////////////////
+ // Dialog types //
+//////////////////
+
+struct DialogIdType {
+    union {
+        enum DialogId dialogId;
+        enum DialogId *dialogIdPtr;
```
EmeraldLoc, line 28: This may be a bit nitpicky and/or incorrect. Perhaps I haven't read the code entirely right, but, why is there a union? Can't this just be the dialog id? You only use dialog id ptr once, and that one time you dereference it, you never set it, so you're wasting 4 bits by not having it just be a dialog id no?

If the pointer doesn't need to be a pointer, then the union definitely makes no sense. There's 2 separate arrays, `sDialogIdTypes`, which is dynamic and always uses the `dialogId`, and `sDefaultDialogIdTypes`, which uses the dialog ptr, but there's no reason it can't just use a `dialogId` directly. It seems like an overcomplication, but maybe I am wrong ¯\\_(ツ)_/¯

EmeraldLoc (changes requested):

src/pc/lua/utils/smlua_text_utils.c
Isaac0-dev: if anything we'd want more content in each comment!
yes, I think its valuable. When scanning a file for a piece of code, the plain english comment is noticed first for some.

EmeraldLoc: While I do disagree with the notion of commenting small things like allocating dialog id types (the code explains itself), if it is preferred I'd imagine a bit more than "Dialog types" would be ideal, even just something as simple as "initialize dialog types", but I've never really heard it from the perspective of scanning through a codebase so that's an interesting thought to ponder

src/pc/lua/utils/smlua_text_utils.c
PeachyPeachSM64: The `gBehaviorValues.dialogs` ids need to be pointers because their value can change at runtime. Dereferencing at compile time would just fix their values and ignore all future changes.
I didn't feel like using two different types for what's basically the same thing. If you think that should be changed, that's fine.

src/audio/external.c
PeachyPeachSM64: idk
sometimes it is, sometimes it is not

EmeraldLoc (approved):

Isaac0-dev:
@coderabbitai full review

src/pc/lua/utils/smlua_text_utils.h
```diff
@@ -42,6 +52,12 @@ void smlua_text_utils_dialog_restore(enum DialogId dialogId);
 bool smlua_text_utils_dialog_is_replaced(enum DialogId dialogId);
 /* |description|Allocates a new dialog entry|descriptionEnd|*/
 s32 smlua_text_utils_allocate_dialog(void);
+/* |description|Gets the type of a `dialogId`|descriptionEnd|*/
+enum DialogType smlua_text_utils_dialog_get_type(enum DialogId dialogId);
+/* |description|Sets the type of a `dialogId`|descriptionEnd|*/
+void smlua_text_utils_dialog_set_type(enum DialogId dialogId, enum DialogType dialogType);
+/* |description|Resets the type of a `dialogId`|descriptionEnd|*/
+void smlua_text_utils_dialog_reset_type(enum DialogId dialogId);
```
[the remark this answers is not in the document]
Isaac0-dev: good observation, but that is not a unique issue to this PR.
please open an issue detailing it. The bug exists at `autogen/convert_functions.py:959`

src/audio/external.c
```diff
@@ -2452,15 +2452,21 @@ void set_sound_moving_speed(u8 bank, u8 speed) {
...
     }
 
-    speaker = sDialogSpeaker[dialogID];
-    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, &speaker);
-    if (speaker < DS_MAX && speaker != 0xff) {
+    smlua_call_event_hooks(HOOK_DIALOG_SOUND, speaker, dialogID, &speaker);
+
+    // Hook returned a sound id
+    if (speaker < 0 || speaker > 0xFF) {
+        play_sound(speaker, gGlobalSoundSource);
```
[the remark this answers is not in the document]
Isaac0-dev: the value of `DS_NONE` is `0xFF`

Isaac0-dev (approved):

PR #1347, opened 2026-07-19
Menu options for resolution and filter

busdriverflix:
A menu was added to enable the user to choose the internally rendered resolution down to the original N64 resolution. Additionally a seperate menu was added to enable the user to choose a filter that makes it look like composite video output.

Menu
Unbenannt


EmeraldLoc:
Branch needs to point to dev!

busdriverflix:
I will develop this feature a bit further and then make another pull request this time for the dev branch :)

PR #1335, opened 2026-07-03
Custom shaders, metal render api, cull toggles, and probably more

EmeraldLoc:
This is a big one

# Shaders

For documentation on how shaders work, [read the docs](https://github.com/EmeraldLoc/sm64coopdx/blob/shader/docs/lua/guides/shaders.md)! This pr description will mainly go over all hooks and functions added along with the systems in place.

## Shader parser, GLSlang, and SPIRV-Cross

To summarize, a parser will go through and add all relevant info to a `Shader`struct to be used for cross compiling with SPIRV-Cross. SPIRV-Cross does reflection to get uniforms at the proper locations and stores it in the `Shader` structure. It also does strict sanitization to make sure there's no funny business that'll happen when cross compiling to DirectX or Metal. 

The general sanitization parser works by going through and looking for uniforms, inputs, and outputs. It'll take those and get the data from it and add in hardcoded locations for inputs and outputs and store those in the shader (it does not store locations for uniforms). When finished it'll double check that all the inputs and outputs line up, and it will add all global uniforms into a Uniform Buffer. The GLSL version is increased to version 450 due to support for explicit uniform buffer locations, and explicit binding locations

SPIRV-Cross is used to cross compile a SPIRV shader to HLSL, MSL or older versions of GLSL. SPIRV is first generated from the sanitized GLSL shader code using glslang. After compiling the shader into SPIRV, we can use that in SPIRV Cross. During that process for HLSL, MSL, etc., we reflect to get uniform data to be used for shaders. Our input and output locations are already good info so we don't need to reflect on that.

With that done a successful conversion has been made!

## The Metal Render API

Metal is Apple's graphics API. Due to macOS being stuck at OpenGL 4.1, it was necessary for a metal renderer to be made to ensure it doesn't get stuck as a legacy render api (more on that below). The metal renderer follows a similar structure to DX11, implementing all required functions. It uses `metal-cpp` so the renderer is written in c++ rather than Obj-C (which sucks, I hate obj-c). The `metal-cpp` files were chosen to be inside the `include` directory, they should realistically be a submodule but I HATE submodules and it shouldn't need to be updated very often at all. Not a whole lot to write on this, everything works in the metal renderer, it's very fast, cool 👍 

## Legacy Render API's

[...]

PR #1349, opened 2026-07-20
Add internal render resolution setting with N64 composite video filters

busdriverflix:
- Adds a new "Internal Resolution" setting (Options → Display) that renders the 3D game at a lower resolution than the window — from common resolutions down to the N64's native 240p — while menus/HUD (DJUI) always stay sharp at native window resolution.
- Adds two new upscale filters alongside Nearest/Linear: "Composite" (soft video-cable look) and "Composite Capture" (mimics a recorded capture of real N64 composite output — scanlines, edge ringing, chroma bleed, analog grain).
- Implemented for both OpenGL and DirectX 11.
- Translated the new menu strings into all supported languages.

Menu
Unbenannt

The D3D11 implementation is not tested yet

### DIFF
--- a/include/PR/gbi_extension.h
+++ b/include/PR/gbi_extension.h
@@ -9,7 +9,7 @@
 // Please update the following table when implementing a new command.
 //
 // RSP ->                     09 0a 0b 0c 0d 0e 0f
-//             14 15 16 17 18 19 1a 1b 1c 1d 1e 1f
+//                15 16 17 18 19 1a 1b 1c 1d 1e 1f
 // 20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d 2e 2f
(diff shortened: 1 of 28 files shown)


Cooliokid956:
Is the D3D11 implementation tested yet

xLuigiGamerx:
Is the D3D11 implementation tested yet x2

ManIsCat2:
Is the Vulkan implementation tested yet

EmeraldLoc:
Heads up! We won't be reviewing or merging this pr due to a different plan being preferred. With a custom multipass shader system in the works (see <https://github.com/coop-deluxe/sm64coopdx/pull/1335>), this is heavily incompatible with that! Currently the plan is to allow post process shaders to be loaded for a client, so you'll be able to do this effect with a shader when the pr is merged. This is a cool pr just unfortunate timing :/

PR #1357, opened 2026-07-30
run hooked bhv loop callbaks on the first tick

Beckowl:
Loop callbacks would never run on an object's first update tick, so ``o.oTimer == 0`` inside a loop would never be true since oTimer is already 1 by the time loop first runs

Fixed ``smlua_call_behavior_hook`` so loop runs every tick, including the first, alongside init to match vanilla where execution falls straight through into the loop on the same tick after init, so the loop function's first call also happens while oTimer is still 0


PeachyPeachSM64:
While I am usually in favor of this kind of bug fixes (if we can call it a bug), I can't agree with this one.
Since the beginning of custom behaviors, loop functions always ran 1 frame after the init. And the custom behaviors rework didn't (couldn't) change that.
Because mods already account for this delay: check `oTimer <= 1`, or call the behavior loop function once at the end of init.

PR #1360, opened 2026-08-03
(borked PR, wrong branch)

anzz1:
There is a serious flaw in current security which needs to be remedied ASAP, loading external binary chunks is allowed, with obvious security implications.

From the [manual](https://www.lua.org/manual/5.4/manual.html)
> It is safe to load malformed binary chunks; load signals an appropriate error. However, Lua does not check the consistency of the code inside binary chunks; running maliciously crafted bytecode can crash the interpreter. 

Now let's get ahead of the CVE and fix this before it happens. There is no reason why binary chunks should have ever been allowed as there is no performance win to using binary chunks and the chunks are in fact often larger than the text version which increases the data transferred over the wire for no reason.

Note that any possible POC or such should *not* be posted here before this fix is pushed to clients and the exploit is patched.


anzz1:
Wrong branch.

PR #1366, opened 2026-08-05
Interpolate the carpet

Cooliokid956:
Interpolate the carpet


EmeraldLoc:
Bro interpolated the carpet

src/game/geo_misc.c
```diff
@@ -132,48 +157,52 @@ Gfx *geo_exec_flying_carpet_create(s32 callContext, struct GraphNode *node, UNUS
...
+            displayListHead = displayList;
+    
+            if (verts == NULL || displayList == NULL) {
+                return NULL;
+            }
 
-        if (verts == NULL || displayList == NULL) {
-            return NULL;
-        }
+            for (n = 0; n <= 20; n++) {
```
Isaac0-dev: this magic value is now repeated. use `NUM_FLYING_CARPET_VERTICES`.

```suggestion
            for (n = 0; n < NUM_FLYING_CARPET_VERTICES; n++) {
```

src/game/geo_misc.c
```diff
@@ -124,6 +127,28 @@ Gfx *geo_exec_flying_carpet_timer_update(s32 callContext, UNUSED struct GraphNod
...
+    sPrevFlyingCarpetRippleTimer = sFlyingCarpetRippleTimer;
+}
+
+void patch_carpet_interpolated(f32 delta) {
+    if (sCachedCarpetGfx && sFlyingCarpetRippleTimer != sPrevFlyingCarpetRippleTimer) {
+        s16 n, row, col, timer;
+        Vtx *verts = (Vtx *)sCachedCarpetGfx[1].words.w1;
+        timer = delta_interpolate_s32(sFlyingCarpetRippleTimer - 0x400, sFlyingCarpetRippleTimer, delta);
+        
+        for (n = 0; n <= 20; n++) {
```
Isaac0-dev:
```suggestion
        for (n = 0; n < NUM_FLYING_CARPET_VERTICES; n++) {
```

src/game/geo_misc.c
```diff
@@ -124,6 +127,28 @@ Gfx *geo_exec_flying_carpet_timer_update(s32 callContext, UNUSED struct GraphNod
...
 
+static Gfx *sCachedCarpetGfx;
+
+void patch_carpet_before(void) {
+    sCachedCarpetGfx = NULL;
+    sPrevFlyingCarpetRippleTimer = sFlyingCarpetRippleTimer;
+}
+
+void patch_carpet_interpolated(f32 delta) {
+    if (sCachedCarpetGfx && sFlyingCarpetRippleTimer != sPrevFlyingCarpetRippleTimer) {
```
Isaac0-dev, line 138: ~~since `sPrevFlyingCarpetRippleTimer` is only updated once every game tick, the carpet visual update rate is effectively capped to 60hz even though coop is capable of near infinite frame generation.
why did you include this condition `sFlyingCarpetRippleTimer != sPrevFlyingCarpetRippleTimer`? As far as I can tell, it should simply be removed.~~

Ignore this, it's wrong...

src/game/geo_misc.c
```diff
@@ -124,6 +127,28 @@ Gfx *geo_exec_flying_carpet_timer_update(s32 callContext, UNUSED struct GraphNod
...
+        s16 n, row, col, timer;
+        Vtx *verts = (Vtx *)sCachedCarpetGfx[1].words.w1;
+        timer = delta_interpolate_s32(sFlyingCarpetRippleTimer - 0x400, sFlyingCarpetRippleTimer, delta);
```
Isaac0-dev: use `sPrevFlyingCarpetRippleTimer` instead of the `0x400` magic value, since you're already storing that - and technically its more correct.

Also, this code needs to be wrap aware. `sFlyingCarpetRippleTimer` normally wraps when it hits the `s16` limit, so the `now - before` calculation needs to be wrapped. Should be done before the interpolation.

src/game/shadow.c
```diff
@@ -349,11 +346,11 @@ void make_shadow_vertex_at_xyz(Vtx *vertices, s8 index, f32 relX, f32 relY, f32
...
     if (sMarioOnFlyingCarpet) {
-        vtxX += 5;
-        vtxY += 5;
-        vtxZ += 5;
+        relX += 5;
+        relY += 5;
+        relZ += 5;
     }
-    make_vertex(vertices, index, vtxX, vtxY, vtxZ, (u16)textureX << 5, (u16)textureY << 5, 255, 255, 255,
+    make_vertex(vertices, index, relX, relY, relZ, (u16)textureX << 5, (u16)textureY << 5, 255, 255, 255,
```
Isaac0-dev: this change applies to more objects than just the carpet! It's smoother, but that change is beyond the scope of this PR. 
It probably could be discussed though, since it is technically better.

src/game/geo_misc.c
```diff
@@ -124,6 +127,28 @@ Gfx *geo_exec_flying_carpet_timer_update(s32 callContext, UNUSED struct GraphNod
     return NULL;
 }
 
+static Gfx *sCachedCarpetGfx;
```
Isaac0-dev: please add `= NULL` for readability.

```suggestion
static Gfx *sCachedCarpetGfx = NULL;
```

Isaac0-dev (changes requested):

src/game/geo_misc.c
Cooliokid956: I agree!

src/game/geo_misc.c
Cooliokid956: Done

PeachyPeachSM64:
Bro interpolated the carpet

Isaac0-dev (approved):

PeachyPeachSM64 (approved):

PR #1370, opened 2026-08-06
Fix UB in `level_cmd_place_object` when gCurrActNum is <= 0

Beckowl:
We ran into an issue in our romhack port where scrolling textures wouldn't work in some levels. It turned out the current act number was `0`, so `level_cmd_place_object` was doing:

```c
actFlags & (1 << (gCurrActNum - 1))
```

Since `gCurrActNum == 0`, that became a left shift by `-1`, which is undefined behavior, so our editor_Scroll_Textures wouldn't spawn.

`level_cmd_place_object` now handles `gCurrActNum <= 0` by only spawning objects marked with `ALL_ACTS` or `ALL_ACTS_MACRO`. Other act numbers behave just like it did before.

Shoutouts to @ManIsCat2 for spotting this

### DIFF
--- a/src/engine/level_script.c
+++ b/src/engine/level_script.c
@@ -538,7 +538,7 @@ static void level_cmd_place_object(void) {
     u8 actMatch;
 
     if (gCurrActNum > 0) {
-        actMatch = actFlags & (1 << (gCurrActNum - 1));
+        actMatch = actFlags & (1 << (gCurrActNum - 1)) || actFlags == ALL_ACTS_MACRO;
     } else {
         actMatch = (actFlags == ALL_ACTS_MACRO) || (actFlags == ALL_ACTS);
     }
--- a/src/engine/level_script.c
+++ b/src/engine/level_script.c
@@ -531,11 +531,19 @@ static void level_cmd_init_mario(void) {
 }
 
 static void level_cmd_place_object(void) {
-    u8 val7 = gCurrActNum > 0 ? (1 << (gCurrActNum - 1)) : ALL_ACTS;
     u16 model;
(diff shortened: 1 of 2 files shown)


EmeraldLoc (approved):

src/engine/level_script.c
```diff
@@ -531,11 +531,19 @@ static void level_cmd_init_mario(void) {
...
 
-    if (sCurrAreaIndex != -1 && (gLevelValues.disableActs || (CMD_GET(u8, 2) & val7) || CMD_GET(u8, 2) == 0x1F)) {
+    u8 actFlags = CMD_GET(u8, 2);
+    u8 actMatch;
+
+    if (gCurrActNum > 0) {
+        actMatch = actFlags & (1 << (gCurrActNum - 1)) || actFlags == ALL_ACTS_MACRO;
+    } else {
+        actMatch = (actFlags == ALL_ACTS_MACRO) || (actFlags == ALL_ACTS);
+    }
```
Isaac0-dev, line 544: Why not just make literally every act have objects defined with `ALL_ACTS`?


```suggestion
    u8 actMatch = actFlags & (1 << (gCurrActNum - 1)) || actFlags == ALL_ACTS_MACRO || actFlags == ALL_ACTS;
```

Isaac0-dev: oh, lol that ignores the whole issue of UB

Isaac0-dev (approved):

src/engine/level_script.c
Beckowl: because that makes objects spawn on the title screen/credits when they aren't supposed to

Issue #1369, opened 2026-08-06
MSAA 16x causes reported opengl version 0 (aka no opengl support) on linux version v1.5.1


gamagama69:
**Describe the bug**
When enabling MSAA 16x in the settings gui or in the config file, the game, upon a restart to apply the settings, will refuse to load with an error stating opengl version = 0. This does not happen for 8x MSAA or lower

**To Reproduce**
1. Launch sm64 coopdx's compiled linux binary
2. Go into the graphics settings
3. Select MSAA 16x
4. Restart the game
5. Game will refuse to launch with an opengl error message

**Expected behavior**
The game should restart with 16x MSAA enabled

**Screenshots**

<img width="298" height="224" alt="Image" src="https://github.com/user-attachments/assets/819412a6-be61-4bc9-b5ce-f8243d2b72dd" />

**Desktop (please complete the following information):**
 - OS: Opensuse Tumbleweed 20260729 x64
 - GPU: Nvidia RTX 3080 10gb LHR
 - GPU Driver Version: Nvida 580.159.03
 - Application version: v1.5.1
 - DE: KDE Wayland

frog8412:
same here, CachyOS, Plasma on Wayland, NVIDIA GeForce GTX 750 Ti, version 1.6 (in-dev)

EmeraldLoc:
to clarify, the 16x option appears in the settings menu correct? But that option is effectively a nuke right?

That's a bit worrying since we are already querying the max msaa as OpenGL sees it.

If that's the case, this appears to be an issue with gpu drivers. Perhaps it's Nvidia being a pain on Linux? Idk

That's just what I have observed though

paradox9955:
> to clarify, the 16x option appears in the settings menu correct? But that option is effectively a nuke right?

For me, it does not. But when adding it to the configfile, it still crashes.
> If that's the case, this appears to be an issue with gpu drivers. Perhaps it's Nvidia being a pain on Linux? Idk

Unlikely, Nvidia works great on Linux it's just the development that is a pain

EmeraldLoc:
> For me, it does not. But when adding it to the configfile, it still crashes.

Great to know! I'll work on a fix for this then since I know the problem and why that happens. @frog8412 @gamagama69 can you see the 16x msaa setting?

frog8412:
> [@frog8412](https://github.com/frog8412) can you see the 16x msaa setting?

as of 1.5.1, yes

PR #1373, opened 2026-08-09
Query gpu sample limit for msaa in OpenGL before init

EmeraldLoc:
Fixes a portion, or fully, the issue presented in <https://github.com/coop-deluxe/sm64coopdx/issues/1369>

If the configfile ever had a msaa setting set that was above what a persons computer could handle, the game could crash depending on the os and gpu drivers. The fix was querying compatibility beforehand in init. The code mainly follows the compatibility function below, but modifying it to query msaa compatibility.

### DIFF
--- a/src/pc/gfx/gfx_window_opengl.c
+++ b/src/pc/gfx/gfx_window_opengl.c
@@ -74,6 +74,12 @@ static int clamp_window_msaa_before_init() {
         }
     }
 
+#ifdef USE_GLES
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
+#endif
+
     // hidden window
     SDL_Window *window = SDL_CreateWindow(
         "",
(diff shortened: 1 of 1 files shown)


anzz1:
Good catch. You'd assume this is something SDL would catch on its own and abstract away (if MSAA set to higher to what gpu can handle, cap it to maximum it can, and disable altogether if msaa unavailable) but apparently not.

This might be worth reporting to SDL upstream as it seems something it should do, and it might be a bug that it doesn't.

src/pc/gfx/gfx_window_opengl.c
```diff
@@ -61,7 +67,50 @@ static void gfx_window_opengl_reset_dimension_and_pos(void) {
...
+    configWindow.msaa = MIN(configWindow.msaa, maxMsaa);
+
+    SDL_GL_DeleteContext(ctx);
+    SDL_DestroyWindow(window);
+
+    return maxMsaa;
+}
+
 static void gfx_window_opengl_init(const char *window_title) {
+    clamp_window_msaa_before_init();
```
PeachyPeachSM64, line 109: Why do you return a value if you're not using it?
Either update `configWindow.msaa` in `clamp_window_msaa_before_init` and return nothing or return `maxMsaa` and set `configWindow.msaa` here.
Currently, the failure points all return 0, but `configWindow.msaa` is not updated with that value. What was the intended behavior? Set msaa to 0 on failure, or ignore it and continue with a potentially invalid msaa value?

PeachyPeachSM64 (changes requested):

src/pc/gfx/gfx_window_opengl.c
EmeraldLoc: Very annoying that I keep doing this. That was old code, this is not meant to return anything. Fixed

PeachyPeachSM64 (approved):

PR #1374, opened 2026-08-10
Remove extension restrictions in mods and mod fs and add smlua fs utils

xLuigiGamerx:
This might seem like a crazy change but hear me out:
There is NO reason you shouldn't be allowed to use any extension

Let me explain, mods already network valid extensions regardless of the maliciousness of the content, and you might think that making it possible to network executable files makes this more dangerous, but no, not unless you cache it, now instead of doing that, I get all the files in a mod directory BUT on receiving those files, I decide to ONLY cache in safe extensions which we already have a list of, this removes any extra potential danger that might have come from this making it as safe as not having it.

Now to address actual additions
Added `smlua_fs_utils.c` 
Added `mod_file_read`
Renamed `get_mod_files` -> `mod_files_get` for consistency and moved to fs utils, deprecated the old name
Added `-- unrestricted-extensions` which is off by default to make old mods not network all uncompiled assets and only network cacheable files

### `mod_file_read()`
`mod_file_read(mod, relativeFileName)` returns a bytestring if the file was found

### `-- unrestricted-extensions`
`-- unrestricted-extensions: boolean, pattern, pattern2, ...` enables loading of unrestricted extensions, if true, the patterns become blacklists, if false they become whitelists, accepts filenames, wildcards, and directories


src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,187 @@
...
+static char *sReadBuffer = NULL;
+static size_t sReadBufferSize = 0;
+
+const char *smlua_fs_utils_mod_file_read(struct Mod *mod, const char *fileName) {
+    if (path_ends_with(fileName, PATH_SEPARATOR) || path_ends_with(fileName, PATH_SEPARATOR_ALT)) {
+        LOG_LUA_LINE("Cannot read '%s' because it is a directory", fileName);
+        return NULL;
+    }
+
+    bool activeFileInMod = (gLuaActiveModFile >= &mod->files[0] && gLuaActiveModFile < &mod->files[mod->fileCount]);
```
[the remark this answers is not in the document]
xLuigiGamerx: Can you elaborate further, what do you recommend i do instead

src/pc/network/packets/packet_download.c
```diff
@@ -341,8 +352,22 @@ static void open_mod_file(struct Mod* mod, struct ModFile* file) {
...
+                    mod_file_create_directories(mod, file);
+                    file->fp = fopen(fullPath, "wb");
+                    cacheableExtension = true;
+                    break;
+                }
+            }
+        }
+        if (!cacheableExtension) {
+            file->fp = f_open_w(fullPath);
+        }
```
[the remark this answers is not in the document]
xLuigiGamerx: this is actually not true, i tested it and i was able to read files of the server a joined without having the files cached

src/pc/lua/utils/smlua_fs_utils.c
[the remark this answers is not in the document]
xLuigiGamerx: "These pointers are related and you are explicitly comparing within the range of the mod files array so I'd imagine this is fine" - EmeraldLockdown

src/pc/network/packets/packet_download.c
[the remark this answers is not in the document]
xLuigiGamerx: I don't follow, this code is basically just doing the same thing as an uncached mod file because it's tagged with a [WIP] tag, why would its behavior be any different, elaborate or check again

src/pc/lua/utils/smlua_fs_utils.c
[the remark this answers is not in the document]
EmeraldLoc: Feel free to correct, since we are doing a range check from the start of the mod array to the end of the mod array, if the pointer lies in between that range, isn't it always within that array? Order doesn't matter because we have an upper bound and a lower bound afaik, or is that wrong?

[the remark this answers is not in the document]
EmeraldLoc: @coderabbitai respond to previous comment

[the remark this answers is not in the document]
xLuigiGamerx: @coderabbitai he was talking to you

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,187 @@
...
+
+    if (needsClose) {
+        if (!inMemory) {
+            fclose(fp);
+        } else {
+            f_close(fp);
+        }
+    }
+
+    return sReadBuffer;
```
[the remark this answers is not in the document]
xLuigiGamerx: Will fix, remember this for the future tho, there is a `ByteString` struct that needs to be returned instead of a const char* to make autogen use a custom smlua push bytestring function instead of lua_pushstring

anzz1:
If I understand you correctly, you want to create a separate filesystem of unpacked arbitrary files alongside the ModFs, which was designed precisely so such a thing wouldn't be needed.

The primary question here is, why? ModFs is exactly designed to be able to hold arbitrary files in a neat controlled package, its a solid system.

And since the engine can only use certain types of files anyway, why remove the list of safe extensions? This doesn't seem to make any sense unless I'm missing something here.

And with ModFs, you can just use .bin as the extension for any binary file, and .txt for any text file. Why the need for arbitrary extensions (and potentially unsafe, nobody can accidentally launch a .bin file by just clicking on it, especially after its wrapped in modFS)?

src/pc/mods/mods_utils.c
```diff
@@ -208,6 +208,31 @@ bool path_ends_with_filepath(const char *path, const char *filepath) {
...
+
+        if ((p == '?' && t != '\0') || tolower((unsigned char)p) == tolower((unsigned char)t)) {
+            pattern++;
+            text++;
+        } else {
+            return false;
+        }
+    }
+    return *pattern == '\0' && *text == '\0';
+}
```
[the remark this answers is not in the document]
xLuigiGamerx: what do you recommend it becomes

src/pc/mods/mod.c
```diff
@@ -326,27 +521,40 @@ static bool mod_load_files_dir(struct Mod* mod, char* fullPath, const char* subD
...
+        
+        bool allowedFile = false;
+        if (blacklist) {
+            // blacklist
+            allowedFile = !matched;
+        } else {
+            // whitelist, cacheable extensions are networked regardless of matching
+            allowedFile = matched || mod_check_file_cacheable(relativePath);
+        }
+        if (!allowedFile) { continue; }
```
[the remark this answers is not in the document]
xLuigiGamerx: elaborate because a modfile shouldn't be able to read mod_cache_add or update if it never makes it into the mod's files

src/pc/mods/mod.c
```diff
@@ -222,12 +233,139 @@ void mod_clear(struct Mod* mod) {
...
+                // allocate a copy for pattern
+                size_t len = strlen(token) + 1;
+                char* pattern = calloc(len, sizeof(char));
+                if (pattern) {
+                    memcpy(pattern, token, len);
+                    fp->patterns[fp->count++] = pattern;
+                }
+            }
+        }
+        token = strtok(NULL, ",");
```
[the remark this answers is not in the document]
xLuigiGamerx: explain how plz

src/pc/mods/mod.c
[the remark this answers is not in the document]
xLuigiGamerx: but when the players receive the downloads, they also check the extensions against the allowed cacheable extensions before ever writing the file, can you check packet_download.c

src/pc/mods/mod.c
```diff
@@ -282,7 +420,64 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+        return path_ends_with(relativePath, pattern + 1);
+    }
+
+    // literal strings
+    if (strchr(pattern, '*') == NULL && strchr(pattern, '?') == NULL) {
+        if (str_ends_with(pattern, "/") || str_ends_with(pattern, "\\")) {
+            return str_starts_with(relativePath, pattern);
+        }
+        return path_ends_with_filepath(relativePath, pattern);
+    }
```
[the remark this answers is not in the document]
xLuigiGamerx: what do you propose

src/pc/mods/mod.c
```diff
@@ -282,7 +420,64 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+    const char *lastDot = strrchr(path, '.');
+    if (lastDot > lastSlash) {
+        for (const char **ext = MOD_FILE_CACHEABLE_EXTENSIONS; *ext; ext++) {
+            if (strcasecmp(lastDot, *ext) == 0) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
xLuigiGamerx: do you mean i should check if both exist

src/pc/mods/mod.c
```diff
@@ -18,6 +18,17 @@
...
+    ".lua", ".luac",                    // script
+    ".txt", ".json", ".ini", ".sav",    // text
+    ".bin", ".col",                     // actors
+    ".bhv",                             // behaviors
+    ".tex", ".png",                     // textures
+    ".lvl",                             // levels
+    ".m64", ".aiff", ".mp3", ".ogg",    // audio
+    NULL
+};
+
```
EmeraldLoc, line 31: Do NOT cache `pngs`. While `.txt`, `.json` and `.ini` weren't cached before, and I am unsure if they should be cached now, they are probably ok. But pngs need to not be cached

src/pc/mods/mod.c
```diff
@@ -222,12 +233,138 @@ void mod_clear(struct Mod* mod) {
         growing_array_free(&mod->customObjectFields);
     }
 
+    if (mod->filePatterns != NULL) {
+        for (size_t i = 0; i < mod->filePatterns->count; i++) {
+            if (mod->filePatterns->patterns[i] != NULL) {
```
EmeraldLoc: Unecessary check, `free` can take in `NULL` just fine, so you can remove the `if` inside the `for` block

src/pc/mods/mod.c
```diff
@@ -222,12 +233,138 @@ void mod_clear(struct Mod* mod) {
...
+            if (blacklistToken) {
+                fp->blacklist = (!strcmp(token, "true"));
+                blacklistToken = false;
+            } else if (fp->count < MOD_FILE_PATTERNS_MAX) {
+                // allocate a copy for pattern
+                size_t len = strlen(token) + 1;
+                char* pattern = calloc(len, sizeof(char));
+                if (pattern) {
+                    memcpy(pattern, token, len);
+                    fp->patterns[fp->count++] = pattern;
```
EmeraldLoc, line 284: Pointer alignment is right, not left. Don't use `if (code) return;`, use `if (code) { return; }`. Same for while loops. Every single block must have `{}` to go along with it

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+            if (strcasecmp(lastDot, *ext) == 0) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+static bool mod_matches_pattern(const char *relativePath, const char *pattern) {
+    if (!relativePath || !pattern || !*pattern) return false;
```
EmeraldLoc:
```c
if (!relativePath || !pattern || !*pattern) { return false; }
```

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+
+    // simple matching
```
EmeraldLoc: When trying to decipher this code, this comment stood out as quite useless. Funnily enough if you wrote no comment instead I probably wouldn't have commented on this lol. Do

`// check for pattern after the .`

and for the next long if do

`// check for pattern after the *.`

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+            char normalizedPattern[SYS_MAX_PATH] = { 0 };
+            snprintf(normalizedPattern, sizeof(normalizedPattern), "%s", pattern);
+
+            size_t length = strlen(normalizedPattern);
+            if (normalizedPattern[length - 1] == '\\') {
+                normalizedPattern[length - 1] = '/';
+            }
+
+            return str_starts_with(relativePath, normalizedPattern);
+        }
```
EmeraldLoc: Uh no. Use `normalize_path` with `PATH_SEPARATOR_ALT` and `PATH_SEPARATOR`

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+        return path_ends_with_filepath(relativePath, pattern);
+    }
+
+    // complex wildcards
+    char patBuf[SYS_MAX_PATH] = { 0 };
+    snprintf(patBuf, sizeof(patBuf), "%s", pattern);
+
+    // append "*" to match contents inside folder
+    if (str_ends_with(patBuf, "/") || str_ends_with(patBuf, "\\")) {
```
EmeraldLoc: You should be able to compare with `PATH_SEPARATOR`, `pathBuf` should be normalized with `normalize_path`

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+        strncat(patBuf, "*", sizeof(patBuf) - strlen(patBuf) - 1);
+    }
+
+    // wildcard match for full relativePath
+    if (wildcard_match(patBuf, relativePath)) {
+        return true;
+    }
+
+    // match files in the root if **/file
+    if (str_starts_with(patBuf, "**/") || str_starts_with(patBuf, "**\\")) {
```
EmeraldLoc: Same here, you'd need to allocate a temporary buffer but it should really use `PATH_SEPARATOR` and the `pathBuf` should be normalized

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+                normalizedPattern[length - 1] = '/';
+            }
+
+            return str_starts_with(relativePath, normalizedPattern);
+        }
+        return path_ends_with_filepath(relativePath, pattern);
+    }
+
+    // complex wildcards
+    char patBuf[SYS_MAX_PATH] = { 0 };
```
EmeraldLoc: This should certainly be called `pathBuf`, was this a typo?

src/pc/mods/mods_utils.c
```diff
@@ -208,6 +208,36 @@ bool path_ends_with_filepath(const char *path, const char *filepath) {
     return matchPtr == path || *(matchPtr - 1) == *PATH_SEPARATOR || *(matchPtr - 1) == *PATH_SEPARATOR_ALT;
 }
 
+bool wildcard_match(const char* pattern, const char* text) {
+    const char* wildcard = NULL;
```
EmeraldLoc: Pointer should be right aligned!

src/pc/mods/mods_utils.c
```diff
@@ -208,6 +208,36 @@ bool path_ends_with_filepath(const char *path, const char *filepath) {
...
+
+    while (*text) {
+        char p = (*pattern == '\\') ? '/' : *pattern;
+        char t = (*text == '\\') ? '/' : *text;
+
+        if (p == '*') {
+            wildcard = pattern++;
+            retryText = text;
+        } else if ((p == '?' && t != '\0') ||
+                   tolower((unsigned char)p) == tolower((unsigned char)t)) {
```
EmeraldLoc: Maybe I'm missing something, but why this awful cast? Unless there is a reason, remove this cast and make the if a single line

src/pc/mods/mods_utils.c
```diff
@@ -208,6 +208,36 @@ bool path_ends_with_filepath(const char *path, const char *filepath) {
     return matchPtr == path || *(matchPtr - 1) == *PATH_SEPARATOR || *(matchPtr - 1) == *PATH_SEPARATOR_ALT;
 }
 
+bool wildcard_match(const char* pattern, const char* text) {
+    const char* wildcard = NULL;
+    const char* retryText = NULL;
+
+    while (*text) {
+        char p = (*pattern == '\\') ? '/' : *pattern;
+        char t = (*text == '\\') ? '/' : *text;
```
EmeraldLoc: Normalize the path beforehand and use `PATH_SEPARATOR` so you don't have to do this!

src/pc/network/packets/packet_download.c
```diff
@@ -457,10 +472,19 @@ after_group:;
 
                     // Write cachedPath here so the file doesn't end up in mod.cache
                     if (!should_cache_mod(mod)) {
-                        char modFilePath[SYS_MAX_PATH] = { 0 };
-                        concat_path(modFilePath, mod->basePath, modFile->relativePath);
-                        normalize_path(modFilePath);
-                        modFile->cachedPath = strdup(modFilePath);
+                        const char *lastSlash = strrchr(modFile->relativePath, '/');
```
EmeraldLoc: Relative path for a mod file is normalized, use `PATH_SEPARATOR`

src/pc/network/packets/packet_download.c
```diff
@@ -341,8 +342,22 @@ static void open_mod_file(struct Mod* mod, struct ModFile* file) {
...
+        if (lastDot > lastSlash) {
+            for (const char **ext = MOD_FILE_CACHEABLE_EXTENSIONS; *ext; ext++) {
+                if (strcasecmp(lastDot, *ext) == 0) {
+                    mod_file_create_directories(mod, file);
+                    file->fp = fopen(fullPath, "wb");
+                    cacheableExtension = true;
+                    break;
+                }
+            }
+        }
```
EmeraldLoc: If you exposed `mod_check_file_cacheable`, couldn't you use that? That'd be a lot nicer

src/pc/network/packets/packet_download.c
EmeraldLoc: If you exposed `mod_check_file_cacheable`, couldn't you use that? That'd be a lot nicer

src/pc/mods/mod.c
```diff
@@ -282,7 +419,72 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
     return file;
 }
 
-static bool mod_load_files_dir(struct Mod* mod, char* fullPath, const char* subDir, const char** fileTypes, bool recursive) {
+static bool mod_check_file_cacheable(const char *path) {
+    if (!path) return false;
+    const char *lastSlash = strrchr(path, '/');
```
EmeraldLoc: I fully understand when this function is called the relative path uses `/`, however if you could normalize it or copy and normalize it before usage with this function, it would make this more available. This function should really also be used in `packet_download.c`, but can't be because of the forward slash

EmeraldLoc (changes requested):
While I still think the `assets` folder approach discussed earlier would be better, this should be ok

src/pc/mods/mod.c
xLuigiGamerx: I kinda disagree, pngs are small enough to not make a difference and you can blacklist them

src/pc/mods/mod.c
xLuigiGamerx: Probably

src/pc/network/packets/packet_download.c
xLuigiGamerx: True

src/pc/mods/mods_utils.c
EmeraldLoc: I was missing something. `tolower` expects a `unsigned char`, it not being unsigned is undefined behavior

src/pc/mods/mod.c
```diff
@@ -23,7 +23,7 @@ const char *MOD_FILE_CACHEABLE_EXTENSIONS[] = {
     ".txt", ".json", ".ini", ".sav",    // text
     ".bin", ".col",                     // actors
     ".bhv",                             // behaviors
-    ".tex", ".png",                     // textures
+    ".tex", /* ".png", */               // textures (pngs should be enabled later when tex generation is deprecated/removed since we do support png textures in dynos)
```
EmeraldLoc: This is an unnecessary comment, nor has this been discussed with the dev team, please remove it

src/pc/mods/mod.c
```diff
@@ -250,25 +248,25 @@ void mod_clear(struct Mod* mod) {
...
     snprintf(str, sizeof(str), "%s", input);
 
-    char* token = strtok(str, ",");
+    char *token = strtok(str, ",");
     bool blacklistToken = true;
 
     while (token != NULL) {
         // leading spaces
-        while (isspace((u8)*token)) token++;
+        while (isspace((u8)*token)) { token++; }
```
EmeraldLoc, line 266: This cast is wrong. If you were to cast to anything it'd be a `s32` since that's what `isspace` takes in, but you don't need to cast

src/pc/mods/mod.c
xLuigiGamerx: Commented out

src/pc/mods/mod.c
xLuigiGamerx: Done

src/pc/mods/mod.c
xLuigiGamerx: Done

src/pc/mods/mod.c
xLuigiGamerx: Done

src/pc/mods/mod.c
xLuigiGamerx: Done

src/pc/mods/mod.c
xLuigiGamerx: Fixed

src/pc/mods/mods_utils.c
xLuigiGamerx: Adjusted

src/pc/mods/mods_utils.c
xLuigiGamerx: Done

src/pc/network/packets/packet_download.c
xLuigiGamerx: Done

src/pc/network/packets/packet_download.c
xLuigiGamerx: Done

src/pc/mods/mod.c
```diff
@@ -539,9 +539,18 @@ static bool mod_load_files_dir(struct Mod* mod, char* fullPath, const char* subD
         bool blacklist = (mod->filePatterns != NULL && mod->filePatterns->blacklist);
         bool matched = false;
         if (mod->filePatterns != NULL) {
+            normalize_path(relativePath);
```
EmeraldLoc: Why normalize here? Just normalize the path after it's initialized with snprintf

src/pc/mods/mod.c
xLuigiGamerx: Done

src/pc/mods/mod.c
xLuigiGamerx: Removed

src/pc/mods/mod.c
xLuigiGamerx: Fixed

src/pc/mods/mod.c
xLuigiGamerx: Done

EmeraldLoc (changes requested):

src/pc/mods/mod.c
xLuigiGamerx: My brother, it's a todo comment, why are you against it, everyone leaves todo comments all around the code

EmeraldLoc: > This is an unnecessary comment, nor has this been discussed with the dev team, please remove it

There has never an agreement that tex will become deprecated in the future, nor is this a necessary comment at all, please remove it

src/pc/mods/mod.c
EmeraldLoc: ^ I'm wrong, per the c standard:

> The behavior is undefined if the value of ch is not representable as unsigned char and is not equal to [EOF](https://en.cppreference.com/c/io).

A `u8` is equivalent to an unsigned char

src/pc/mods/mod.c
xLuigiGamerx: tex deprecation means that it just shouldn't be generated anymore, it can still be used but new tex files shouldn't be generated, this should be added in the future because it skips an initial generation step since dynos supports pngs perfectly fine

src/pc/mods/mod.c
xLuigiGamerx: why tho, it's not needed before here and when the path initialized, it's still the same loop so im not doing it multiple times here

src/pc/mods/mod.c
EmeraldLoc: This has not been discussed, nor does it have a plan on being implemented by anyone on the dev team or even agreeing with the idea of it at all, so there should not be a comment saying what we should do. That's the last I'll say on that. Leave it out of the codebase for now, thanks

src/pc/mods/mod.c
EmeraldLoc: Because it's random and not very clean and affects the function calls below which may be counterintuitive. It just doesn't look nice. Better to have it in clear view imo

src/pc/mods/mod.c
xLuigiGamerx: alright, i'll add it back myself when i deprecate tex generation later

src/pc/mods/mod.c
xLuigiGamerx: wdym affects function calls below, it's where it's needed, it will also affect function calls below regardless, this just makes it only run if there are actually patterns

EmeraldLoc: `allowedFile = matched || mod_check_file_cacheable(relativePath);`
`struct ModFile* file = mod_allocate_file(mod, relativePath);`

It does affect the string going into these functions

It doesn't break anything, these functions just run normalize again, it's just a weird spot to put it and it would match the rest of the codebase and make a **lot** more sense to have it after it's set by `snprintf`. If for some reason you are super against changing it, you can wait for another developer to chime in, but I'd imagine the response would be to just move it up

xLuigiGamerx: it's more of that there are also functions above that i don't want to affect

EmeraldLoc: `mod_load_files_dir`? It does not care if the path is normalized or not so it's fine

xLuigiGamerx: aight, i'll move it up

src/pc/mods/mod.c
xLuigiGamerx: removed

src/pc/mods/mod.c
xLuigiGamerx: moved

EmeraldLoc (approved):

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
+#include "smlua_fs_utils.h"
+#include "pc/mods/mods_utils.h"
+#include "pc/lua/smlua.h"
+#include "pc/lua/smlua_utils.h"
+#include "pc/fs/fmem.h"
+
+ByteString mod_file_read(struct Mod *mod, const char *fileName) {
```
PeachyPeachSM64, line 7: it's `filename`, not `fileName`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    bool activeFileInMod = gLuaActiveMod == mod && gLuaActiveModFile != NULL;
+
+    char baseDir[SYS_MAX_PATH] = { 0 };
+    if (activeFileInMod && strlen(gLuaActiveModFile->relativePath) > 0) {
+        snprintf(baseDir, SYS_MAX_PATH, "%s", gLuaActiveModFile->relativePath);
+
+        char* lastSeparator = strrchr(baseDir, *PATH_SEPARATOR);
+        char* lastAltSeparator = strrchr(baseDir, *PATH_SEPARATOR_ALT);
```
PeachyPeachSM64, line 26: `char *`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+            *lastSeparator = '\0';
+        } else {
+            baseDir[0] = '\0';
+        }
+    }
+    
+    char targetPath[SYS_MAX_PATH] = { 0 };
+    resolve_relative_path(baseDir, fileName, targetPath);
+
+    struct ModFile* file = NULL;
```
PeachyPeachSM64, line 42: `struct ModFile *`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    for (int i = 0; i < mod->fileCount; i++) {
```
PeachyPeachSM64, line 43: `u16 i`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    if (!file) {
+        LOG_LUA_LINE("File '%s' not found in '%s'", fileName, mod->name);
+        return byteString;
+    }
+
+    static char *sReadBuffer = NULL;
+    static size_t sReadBufferSize = 0;
+    FILE* fp = file->fp;
```
PeachyPeachSM64, line 61: `FILE *`
Why is `fp` defined here? You don't use it until line 77

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+
+    size_t fileSize = file->size;
+    if (sReadBufferSize < fileSize) {
+        char *newBuffer = realloc(sReadBuffer, fileSize);
+        if (newBuffer) {
+            sReadBuffer = newBuffer;
+            sReadBufferSize = fileSize;
+        } else {
+            LOG_LUA_LINE("Could not set read buffer for '%s'", file->relativePath)
```
PeachyPeachSM64, line 70: Missing `;`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+        return byteString;
+    }
+
+    if (!sReadBuffer) {
+        if (needsClose) {
+            f_close(fp);
+        }
+        LOG_LUA_LINE("Ran out of memory while reading '%s'", file->relativePath);
+        return byteString;
+    }
```
PeachyPeachSM64, line 93: Why is it here? You don't need `fp` to check the buffer!

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+
+    f_rewind(fp);
+
+    size_t bytesRead;
+    bytesRead = f_read(sReadBuffer, 1, fileSize, fp);
```
PeachyPeachSM64, line 98: make it a single line

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+
+    if (needsClose) {
+        f_close(fp);
+    }
+
+    if (bytesRead != fileSize) {
+        LOG_LUA_LINE("Filesize mismatched with the amount of bytes read for file '%s'", file->relativePath);
```
PeachyPeachSM64, line 105: That's cool, but it would be useful to see the value of bytes read vs the expected value

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+        return byteString;
+    }
+
+    byteString.bytes = sReadBuffer;
+    byteString.length = bytesRead;
+
+    return byteString;
+}
+
+bool mod_file_exists(const char *filename) {
```
PeachyPeachSM64, line 115: `mod_file_read` can use any mod but this one can only check the current mod? Why?

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    if (gLuaActiveMod == NULL) { return false; }
```
PeachyPeachSM64, line 116: Unusable during loading...

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    char normPath[SYS_MAX_PATH] = { 0 };
+    char normRelative[SYS_MAX_PATH] = { 0 };
+
+    if (snprintf(normPath, sizeof(normPath), "%s", filename) < 0) {
+        LOG_ERROR("Failed to copy filename for normalization: %s", filename);
+    }
+
+    normalize_path(normPath);
+
+    for (s32 i = 0; i < gLuaActiveMod->fileCount; i++) {
```
PeachyPeachSM64, line 127: `fileCount` is `u16`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+        struct ModFile *file = &gLuaActiveMod->files[i];
+        strcpy(normRelative, file->relativePath);
```
PeachyPeachSM64, line 129: Use `strncpy` or `snprintf`!

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+        normalize_path(normRelative);
+        if (!strcmp(normRelative, normPath)) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+LuaTable mod_files_get(struct Mod *mod, OPTIONAL const char *subDirectory, OPTIONAL bool relative) {
```
PeachyPeachSM64, line 139: Technically, you don't get the mod files, but only the filenames... A more appropriate name would be `mod_files_list`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    normalize_path(normalizedSubDir);
+
+    size_t subDirLen = strlen(normalizedSubDir);
+    if (subDirLen > 0 && subDirLen + 1 < SYS_MAX_PATH && normalizedSubDir[subDirLen - 1] != *PATH_SEPARATOR) {
+        strcat(normalizedSubDir, PATH_SEPARATOR);
+        subDirLen = strlen(normalizedSubDir);
+    }
+
+    struct lua_State *L = gLuaState;
+    if (!L) { return 0; }
```
PeachyPeachSM64, line 160: Move that at the top and simplify the `if (!mod)` condition please
This function should also error and return nothing if `mod` is NULL, like `mod_file_read`

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+
+    LUA_STACK_CHECK_BEGIN_NUM(L, 1);
+
+    lua_newtable(L);
+
+    int luaTableIndex = 1;
+    for (int i = 0; i < mod->fileCount; i++) {
```
PeachyPeachSM64, line 167: `u16 i`

src/pc/lua/utils/smlua_fs_utils.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "pc/lua/smlua_utils.h"
+
+/* |description|Reads a file in a given mod, relative to the current lua file if it's the active mod, and returns its content as a bytestring|descriptionEnd|*/
+ByteString mod_file_read(struct Mod *mod, const char *fileName);
+/* |description|Checks if a file exists inside of a mod|descriptionEnd| */
+bool mod_file_exists(const char* filename);
+/* |description|Gets all files a mod contains|descriptionEnd| */
+LuaTable mod_files_get(struct Mod* mod, OPTIONAL const char* subDirectory, OPTIONAL bool relative);
```
PeachyPeachSM64, line 10: `const char *`, `struct Mod *`, `const char *`

src/pc/mods/mod.h
```diff
@@ -29,6 +35,12 @@ struct ModFile {
     char* cachedPath;
 };
 
+struct ModFilePatterns {
+    bool blacklist;
+    const char *patterns[MOD_FILE_PATTERNS_MAX];
```
PeachyPeachSM64, line 40: Why isn't this a growing array?

src/pc/mods/mod.h
```diff
@@ -37,6 +49,7 @@ struct Mod {
     char relativePath[SYS_MAX_PATH];
     char basePath[SYS_MAX_PATH];
     struct ModFile* files;
+    struct ModFilePatterns* filePatterns;
```
PeachyPeachSM64, line 52: Why is it a pointer?

src/pc/mods/mod.c
```diff
@@ -222,12 +233,136 @@ void mod_clear(struct Mod* mod) {
...
+        char *end = token + strlen(token);
+        while (end > token && isspace((u8)*(end - 1))) {
+            *--end = '\0';
+        }
+
+        if (*token != '\0') {
+            if (blacklistToken) {
+                fp->blacklist = (!strcmp(token, "true"));
+                blacklistToken = false;
+            } else if (fp->count < MOD_FILE_PATTERNS_MAX) {
```
PeachyPeachSM64, line 278: Immediately return if you hit the limit. And you wouldn't have to worry about if it was a growing array...

src/pc/mods/mod.c
```diff
@@ -222,12 +233,136 @@ void mod_clear(struct Mod* mod) {
...
+                // allocate a copy for pattern
+                size_t len = strlen(token) + 1;
+                char *pattern = calloc(len, sizeof(char));
+                if (pattern) {
+                    memcpy(pattern, token, len);
```
PeachyPeachSM64, line 283: That's `strdup`

src/pc/mods/mod.c
```diff
@@ -222,12 +233,136 @@ void mod_clear(struct Mod* mod) {
...
+    // open file
+    FILE* f = fopen(path, "rb");
+    if (f == NULL) {
+        LOG_ERROR("Failed to open '%s'", path);
+        return;
+    }
+    fseek(f, 0, SEEK_SET);
+
+    // read line-by-line
+    #define BUFFER_SIZE MAX(MAX(MOD_NAME_SIZE, MOD_INCOMPATIBLE_SIZE), MAX(MOD_FILE_PATTERNS_SIZE, MOD_DESCRIPTION_SIZE))
```
PeachyPeachSM64, line 322: Please move this out of the function or `static const size_t` it

src/pc/mods/mod.c
```diff
@@ -282,7 +417,74 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
     return file;
 }
 
-static bool mod_load_files_dir(struct Mod* mod, char* fullPath, const char* subDir, const char** fileTypes, bool recursive) {
+bool mod_check_file_cacheable(const char *path) {
+    if (!path) return false;
```
PeachyPeachSM64, line 421: `if (!path) { return false; }`

src/pc/mods/mod.c
```diff
@@ -282,7 +417,74 @@ static struct ModFile* mod_allocate_file(struct Mod* mod, char* relativePath) {
...
+        for (const char **ext = MOD_FILE_CACHEABLE_EXTENSIONS; *ext; ext++) {
+            if (strcasecmp(lastDot, *ext) == 0) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+static bool mod_matches_pattern(char *relativePath, char *pattern) {
```
PeachyPeachSM64, line 441: `const char *`, `const char *`

src/pc/mods/mod.c
```diff
@@ -356,48 +581,13 @@ static bool mod_load_files_dir(struct Mod* mod, char* fullPath, const char* subD
...
 
 static bool mod_load_files(struct Mod* mod, char* fullPath) {
-    // read single lua file
     if (!mod->isDirectory) {
+        // read single file mods
         return (mod_allocate_file(mod, mod->relativePath) != NULL);
+    } else {
+        // read folder mods
+        return (mod_load_files_dir(mod, fullPath, "", true));
     }
```
PeachyPeachSM64, line 590: Unneeded parentheses. You also don't need `else` after a `return`.

PeachyPeachSM64 (changes requested):
I don't like how the unrestricted extensions thing is handled. The whole true/false for blacklist/whitelist as first token is kind of confusing and unintuitive.
What if I want to allow `.txt` files but not `this_particular_file.txt`? Or blacklist every `.txt` file except `this_mandatory_file.txt`?
`.gitignore` files have a straight forward way to handle this: everything listed is blacklisted (or whitelisted, depending on your need), unless prefixed with `!`. It's simple and you don't have to think twice about how it works.
Pattern matching is cool, but I feel like you're trying to reinvent the wheel. You should check existing (and well tested) regex C libs like [this one](https://github.com/kokke/tiny-regex-c).

EmeraldLoc:
I forgot to mention this! Please document these changes!

src/pc/lua/utils/smlua_fs_utils.c
xLuigiGamerx: That was already the case when I moved it, I might overload them all to accept omitting the first param and have this one have 2 params

src/pc/lua/utils/smlua_fs_utils.c
xLuigiGamerx: I didn't modify the function, it's existing behavior, I just moved it

src/pc/mods/mod.h
xLuigiGamerx: I don't know how to set it up, can you help me on discord later

src/pc/mods/mod.h
xLuigiGamerx: No real reason probably, I probably thought i needed it

src/pc/mods/mod.c
xLuigiGamerx: Refer to my reply to your first growing array review

src/pc/mods/mod.c
xLuigiGamerx: I mean it was there before I even touched it, ask whichever dev added it, probably dj

src/pc/mods/mod.c
xLuigiGamerx: Emerald told me to switch them

src/pc/mods/mod.h
PeachyPeachSM64: Sure I can

src/pc/mods/mod.c
EmeraldLoc: Because you explicitly showed me this when asking

```c
static bool mod_matches_pattern(const char *relativePath, const char *pattern) {
    if (!relativePath || !pattern || !*pattern) { return false; }

    normalize_path(relativePath); // here
    normalize_path(pattern); // here
```

And since you were mutating `relativePath` and `pattern`, it needed to not be a `const`.

But since you ended up not doing that, they can be `const` now

PeachyPeachSM64: oh, because of `path_basename` being `char *`, right?

*sigh*

PeachyPeachSM64: Well, even with this stupidity, `pattern` has no reason to not be `const char *` at least

Amy54Desu:
> There is NO reason you shouldn't be allowed to use any extension

Security reasons probably

xLuigiGamerx:
> > There is NO reason you shouldn't be allowed to use any extension
> 
> Security reasons probably

For the love of God please read before commenting

docs/lua/functions-7.md
```diff
@@ -373,6 +373,84 @@ Checks if the surface is a painting warp
...
+
+[:arrow_up_small:](#)
+
+<br />
+
+## [mod_file_exists](#mod_file_exists)
+
+### Description
+Checks if a file exists inside of a mod
+
```
xLuigiGamerx:
```suggestion

### Lua Example
`local booleanValue = mod_file_exists(mod, filename)`

### Parameters
| Field | Type |
| ----- | ---- |
| mod | [Mod](structs.md#Mod) |
| filename | `string` |

### Returns
- `boolean`

### C Prototype
`bool mod_file_exists_in_mod(struct Mod *mod, const char *filename);`

---
### Description
Checks if a file exists inside of the current mod

```

docs/lua/functions-7.md
```diff
@@ -373,6 +373,84 @@ Checks if the surface is a painting warp
...
+`bool mod_file_exists(const char* filename);`
+
+[:arrow_up_small:](#)
+
+<br />
+
+## [mod_files_get](#mod_files_get)
+
+### Description
+Gets all files a mod contains
```
xLuigiGamerx:
```suggestion
Gets all files in a mod within `subDirectory` and with upper path prefixes stripped if relative
```

docs/lua/functions-7.md
```diff
@@ -373,6 +373,84 @@ Checks if the surface is a painting warp
...
+<br />
+
+
+## [mod_file_read](#mod_file_read)
+
+### Description
+Reads a file in a given mod, relative to the current lua file if it's the active mod, and returns its content as a bytestring
+
+### Lua Example
+`local stringValue = mod_file_read(mod, fileName)`
```
xLuigiGamerx:
```suggestion
`local stringValue = mod_file_read(mod, filename)`
```

src/pc/lua/utils/smlua_fs_utils.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "pc/lua/smlua_utils.h"
+
+/* |description|Reads a file in a given mod, relative to the current lua file if it's the active mod, and returns its content as a bytestring|descriptionEnd|*/
+ByteString mod_file_read(struct Mod *mod, const char *fileName);
```
xLuigiGamerx:
```suggestion
OVERLOAD(mod_file_read) ByteString mod_file_read_in_mod(struct Mod *mod, const char *filename);
/* |description|Reads a file in the current mod, relative to the current lua file, and returns its content as a bytestring|descriptionEnd|*/
OVERLOAD(mod_file_read) ByteString mod_file_read(const char *filename);

```

src/pc/lua/utils/smlua_fs_utils.c
xLuigiGamerx:
```suggestion
bool mod_file_exists(const char *filename) {
    mod_file_exists_in_mod(gLuaActiveMod, filename);
}

LuaTable mod_files_list_in_mod(struct Mod *mod, OPTIONAL const char *subDirectory, OPTIONAL bool relative) {
```

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+        normalize_path(normalizedModPath);
+
+        if (!strcmp(normalizedModPath, targetPath) || !strcmp(normalizedModPath, fileName)) {
+            file = &mod->files[i];
+            break;
+        }
+    }
+
+    if (!file) {
+        LOG_LUA_LINE("File '%s' not found in '%s'", fileName, mod->name);
```
xLuigiGamerx:
```suggestion
        LOG_LUA_LINE("File '%s' not found in '%s'", filename, mod->name);
```

src/pc/lua/smlua_functions_autogen.c
```diff
@@ -31161,6 +31162,73 @@ int smlua_func_network_player_palette_to_color(lua_State* L) {
...
+
+    return 1;
+}
+
+int smlua_func_mod_file_exists(lua_State* L) {
+    if (L == NULL) { return 0; }
+
+    int top = lua_gettop(L);
+    if (top != 1) {
+        LOG_LUA_LINE("Improper param count for '%s': Expected %u, Received %u", "mod_file_exists", 1, top);
```
xLuigiGamerx:
```suggestion
    if (top != 2 && top != 1) {
        LOG_LUA_LINE("Improper param count for 'mod_file_exists': Expected 2 or 1, Received %u", top);
```

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    return false;
+}
+
+LuaTable mod_files_get(struct Mod *mod, OPTIONAL const char *subDirectory, OPTIONAL bool relative) {
+    if (!mod) {
+        struct lua_State *L = gLuaState;
+        if (L) {
+            lua_newtable(L);
+            return smlua_to_lua_table(L, -1);
+        }
```
xLuigiGamerx:
```suggestion
        LOG_LUA_LINE("Attempted to read a nil mod");
        return 0;
    }    
    
    if (!gLuaState) {
        LOG_LUA_LINE("'gLuaState' was not found due potential corruption or a bug");
```

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+    char targetPath[SYS_MAX_PATH] = { 0 };
+    resolve_relative_path(baseDir, fileName, targetPath);
+
+    struct ModFile* file = NULL;
+    for (int i = 0; i < mod->fileCount; i++) {
+        char normalizedModPath[SYS_MAX_PATH] = { 0 };
+        snprintf(normalizedModPath, SYS_MAX_PATH, "%s", mod->files[i].relativePath);
+        normalize_path(normalizedModPath);
+
+        if (!strcmp(normalizedModPath, targetPath) || !strcmp(normalizedModPath, fileName)) {
```
xLuigiGamerx:
```suggestion
        if (!strcmp(normalizedModPath, targetPath) || !strcmp(normalizedModPath, filename)) {
```

src/pc/lua/utils/smlua_fs_utils.c
```diff
@@ -0,0 +1,186 @@
...
+
+ByteString mod_file_read(struct Mod *mod, const char *fileName) {
+    ByteString byteString = { NULL, 0 };
+    if (!mod) {
+        LOG_LUA_LINE("Attempted to read a nil mod");
+        return byteString;
+    }
+
+    if (path_ends_with(fileName, PATH_SEPARATOR) || path_ends_with(fileName, PATH_SEPARATOR_ALT)) {
+        LOG_LUA_LINE("Cannot read '%s' because it is a directory", fileName);
```
xLuigiGamerx:
```suggestion
    if (path_ends_with(filename, PATH_SEPARATOR) || path_ends_with(filename, PATH_SEPARATOR_ALT)) {
        LOG_LUA_LINE("Cannot read '%s' because it is a directory", filename);
```

anzz1:
Well there are security reasons - if I understood correctly the confusing explanation, that with unrestricted extensions you could send any extension over the network. And on the other hand if you don't do that, then some of the files are not cached, that means you are only sending *some* of the mod's files, but not all, so the receiver ends up with a broken mod that doesn't have all the files needed for it to run properly? That would be a bad, bad thing, and break the whole mod distribution system.

Most importantly, *why* do you need arbitrary extensions? That question was left unanswered. Why do you need file types that the game cannot load anyway? And if you're storing data or something, modFS already exists.

You claimed there is "NO reason you shouldn't be allowed to use any extension", there are some reasons. On the other hand, there is NO reason why you *should* be allowed to use any extension. Or if there is, you didn't make that use-case clear.

xLuigiGamerx:
> Well there are security reasons - if I understood correctly the confusing explanation, that with unrestricted extensions you could send any extension over the network. And on the other hand if you don't do that, then some of the files are not cached, that means you are only sending _some_ of the mod's files, but not all, so the receiver ends up with a broken mod that doesn't have all the files needed for it to run properly? That would be a bad, bad thing, and break the whole mod distribution system.

What mod distribution, if you didn't get the hint, I don't care about your open source ideology, the game does send all files but it only caches safe extensions and that's fine because again, I don't CARE about your open source ideology

> Most importantly, _why_ do you need arbitrary extensions? That question was left unanswered. Why do you need file types that the game cannot load anyway? And if you're storing data or something, modFS already exists.

Because I want to use it, I left you unanswered on purpose

> You claimed there is "NO reason you shouldn't be allowed to use any extension", there are some reasons. On the other hand, there is NO reason why you _should_ be allowed to use any extension. Or if there is, you didn't make that use-case clear.

Organization purposes, my statement specifically refers to the restriction existing in the first place which shouldn't be the case since extensions are arbitrary

anzz1:
I carefully went through your messages in this PR, twice, and I honestly cannot find any "hint" towards you not caring about the functionality of the open source mod distribution system. Except if you meant that seems to be the purpose of the PR, although not stated as such. When making a PR towards, in any project, you should explicitly state your intentions, goals and effects of a PR without relying anyone getting a "hint", nobody can get inside your head and know what you're thinking. Communication is key.

And to answer to "what mod distribution system", well that'd be the one which has existed ever since mods has existed for the coop project, from the very beginning. If your intention is to change the system in a major way, that it breaks the entire system's current functionality as right now the mods are sent and cached as whole, so they can be re-hosted, edited, etc. they work out of the box as-is, and after the change only parts of the mod would be cached breaking the mod from working, that is definitely something you should explicitly state, otherwise it seems dishonest like slipping a breaking change under the door while not mentioning it at all. Maybe you didn't mention that by accident and there is no deliberate hiding involved, but you have to realize not communicating such change is not how things are done.

EmeraldLoc:
Realistically I don't *think* this pr is getting merged as is, at least with the extension stuff (the read file stuff is fine, in fact I really want it for some of my mods for clarity sake). A lot of the dev team (including myself) doesn't really like it because it's complicated, convoluted, and there's easier to understand solutions.

A few alternatives suggested in Discord are as follows:

1. a gitignore style list for files you explicitly want to include in your mod, it's simple and intuitive enough 
2. use a mod manifest and allow that to explicitly add/remove files with clean json syntax
3. create an `assets` folder where any assets of any extension can go and be networked

Those are the three I think are pretty decent alternatives, in no particular order

That all being said, these are all in a broader scope, not specifically for extensions. These are for files in general, and that is how it should be, that way you can have development files on a local build but not send all that out over the network when it isn't necessary. These ideas can accomplish what Luigi wants but in a clearer manner. I prefer 2 and 3, with 1 being a bit meh since it's another file to include in your mod and reserve.

Mainly posting this here so anyone can chime in with ideas they have on a good way to approach this, since it is an actual problem. Some mods may want files in their mod (shader files, json files, xml files, binary files, etc.) that they want to load for everyone, but because they aren't sent or synced over the network, they can't right now, and lua can't see them. The only way they can is using a hacky system where you create a `.lua` file and return the contents of whatever file you are making into a string and then require it, but that's ugly and kills any vscode syntax highlighting and IntelliSense and isn't at all necessary if we implement the right system.

xLuigiGamerx:
> Realistically I don't _think_ this pr is getting merged as is, at least with the extension stuff (the read file stuff is fine, in fact I really want it for some of my mods for clarity sake). A lot of the dev team (including myself) doesn't really like it because it's complicated, convoluted, and there's easier to understand solutions.
> 
> A few alternatives suggested in Discord are as follows:
> 
> 1. a gitignore style list for files you explicitly want to include in your mod, it's simple and intuitive enough
> 2. use a mod manifest and allow that to explicitly add/remove files with clean json syntax
> 3. create an `assets` folder where any assets of any extension can go and be networked
> 
> Those are the three I think are pretty decent alternatives, in no particular order
> 
> That all being said, these are all in a broader scope, not specifically for extensions. These are for files in general, and that is how it should be, that way you can have development files on a local build but not send all that out over the network when it isn't necessary. These ideas can accomplish what Luigi wants but in a clearer manner. I prefer 2 and 3, with 1 being a bit meh since it's another file to include in your mod and reserve.

I have no problem with 1 and 2 but I've already shown my disagreement with 3 and agent x already showed that `assets` might already be used by mod makers as a raw assets folder.
Also i think 2 might be out of the scope of this pr since it's making an entirely new unrelated system to the pr, if you could bring back your mod manifest pr, I'll happily rely on that.

xLuigiGamerx:
> I carefully went through your messages in this PR, twice, and I honestly cannot find any "hint" towards you not caring about the functionality of the open source mod distribution system. Except if you meant that seems to be the purpose of the PR, although not stated as such. When making a PR towards, in any project, you should explicitly state your intentions, goals and effects of a PR without relying anyone getting a "hint", nobody can get inside your head and know what you're thinking. Communication is key.
> 
> And to answer to "what mod distribution system", well that'd be the one which has existed ever since mods has existed for the coop project, from the very beginning. If your intention is to change the system in a major way, that it breaks the entire system's current functionality as right now the mods are sent and cached as whole, so they can be re-hosted, edited, etc. they work out of the box as-is, and after the change only parts of the mod would be cached breaking the mod from working, that is definitely something you should explicitly state, otherwise it seems dishonest like slipping a breaking change under the door while not mentioning it at all. Maybe you didn't mention that by accident and there is no deliberate hiding involved, but you have to realize not communicating such change is not how things are done.

There was never a mod distribution system, the tmp cache folder was never meant for that, it was simply a way to not have to re-download a mod when rejoining a server, snooping around the tmp folder to steal mods was actually banned before the [WIP] tag and it still is if someone accidentally didn't add it to their private/WIP mods, your idea that there ever was a mod distribution system is wrong and mod distribution always was handled via discord and more recently the coop mod sites. The game never stated that all mods need to be open source or to be shared publicly and never will because that's not how it is.

anzz1:
> Some mods may want files in their mod (shader files, json files, xml files, binary files, etc.) that they want to load for everyone, but because they aren't sent or synced over the network, they can't right now, and lua can't see them.

I think this should be alleviated simply by adding the extensions that can be loaded to the allowlist. Like .glsl .hlsl for shaders, .json, .xml, whatever. Binary files can just be a .bin. I mean what the game can actually load is already known beforehand. Let's say you add a feature to load shaders, just add the file extension to the allow list at the same time. Adding those extensions to the allowlist when adding features that can load such things seems like a no-brainer to me. This is much easier than having a disallow list for dangerous extensions like .exe, .dll, .com, etc. since the list is long. Its not impossible to do that way of course, but *why* is the question. Since we know what extensions the game engine can load, just add those extensions to allow list. This is how almost every game with from-server downloadable assets work.

> There was never a mod distribution system, the tmp cache folder was never meant for that

There absolutely was, originally there was no .tmp folder at all and all mods simply got downloaded straight to the mods folder, as mod distribution was *always* the design. The .tmp folder and the folder id system was added later to alleviate the problem of collisions when people used same names. It was certainly never meant to gatekeep mods, and yes there *always* was a mod distribution system, from day 1, and there still is. Something you seem to want to break because of, I don't know?

> snooping around the tmp folder to steal mods was actually banned before the [WIP] tag and it still is if someone accidentally didn't add it to their private/WIP mods

What are you talking about? This must be some development I've never heard of and was certainly even considered as an option when the whole mod system was originally designed a long time ago. There was never even a concept of "stealing mods" nor "private" mods because the whole system was designed that everything is public from the ground up, which is an amazing system. If someone claimed something else later in development then that's news to me. You can go and ask the original dev that created the project if you like, "private" mods was never a feature. I understand the addition of the WIP to not clog people's caches with mods that are in development and constantly changing, but its intention is not to gatekeep any mods and make them "private" as you say.

>  your idea that there ever was a mod distribution system is wrong and mod distribution always was handled via discord and more recently the coop mod sites

Untrue. The whole mod ecosystem especially in the beginning pretty much solely relied on mods being distributed via the game itself and people modifying them and making new ones. As there was no mod site, and discord activity wasn't as high as the game activity was, mod development mostly happened outside of it. Many mods had their compatibility fixed with newer versions by someone else than the original dev when the original dev was no longer around, features added to create more expansive mods from a simple base, in a collaborative manner, that was and is the whole ecosystem.

What seems to me that you're trying to implement a system to gatekeep mods which was not only not ever intended but also doesn't exist today unless you abuse the WIP tag. For this entire time mods have existed for the project which is several years now, mods have been networked fully intact with anyone able to re-host or edit them as they please. There is no "theft" in open-source. That whole attitude is toxic to the health and future of the ecosystem. Mod makes come and go and the idea is that anyone can pick up and improve a mod, or maintain it when new breaking changes are introduced, instead of mods breaking and dying when new versions with breaking changes are introduced. It also prevents fracture that "oh i'm staying behind in this older version because mod X broke in the new one and nobody can repair it and the original dev isn't around". But I don't know why I'm even trying to make arguments to justify this since first of all, you clearly aren't listening, and secondly *open-source has always been what the mod distribution system is, since day 1*. It's you who are trying to push your own ideology of "private" mods to a project where it never existed nor was ever intended, going against the whole beautiful system of the idea. It actually is pretty infuriating the entitlement of coming into a several year old open source project and then trying to push your closed source gatekeeping features into it. And even going as far as claiming that it was never intended to be open source, like what on Earth are you talking about that has been literally the design always.

Unfortunate due to the grey-area legality nature of the project the whole project could have not been licensed AGPL like the coopnet library was, otherwise it could have been made so that it would not be just bad intent, but actually illegal to try to implement closed source parts into a project that has always been open source and every contributor to it from the beginning expecting it to stay as such. Not having a license indeed does expose a project to someone being malicious and trying to take over and push their own ideologies and then close parts of it later on. But like always, you tend to think good of people and don't really consider it'd become an issue in the future, until it does. It's just so egregious, trying to put closed source features afterwards into an open source project developed by other people before you. This is precisely what license like AGPL is specifically intended to combat, which is an extension to GPL, that you can't only not modify the project and make it closed source later, but that you can't use the project to host anything closed source either. Which give people the peace of mind to contribute knowing that nobody can attempt to takeover the project later and try to close any parts of it, whether the project itself or any modifications built on top of the project.

EmeraldLoc:
> I think this should be alleviated simply by adding the extensions that can be loaded to the allowlist. Like .glsl .hlsl for shaders, .json, .xml, whatever

The exact issue is it can't just be an allowlist since files that previously were not intended to be networked are now networked, think like a README if someone got the mod from GitHub. That shouldn't be sent over the network since it is unused. Or png files, mods have plain png files that shouldn't be networked because they are converted to `.tex`, that's why a simple fix like that isn't viable. Hopefully that makes sense

> There absolutely was, originally there was no .tmp folder at all and all mods simply got downloaded straight to the mods folder, as mod distribution was always the design

> What are you talking about? This must be some development I've never heard of and was certainly even considered as an option when the whole mod system was originally designed a long time ago. There was never even a concept of "stealing mods" nor "private" mods because the whole system was designed that everything is public from the ground up, which is an amazing system. If someone claimed something else later in development then that's news to me. You can go and ask the original dev that created the project if you like, "private" mods was never a feature. I understand the addition of the WIP to not clog people's caches with mods that are in development and constantly changing, but its intention is not to gatekeep any mods and make them "private" as you say.

First, I had no idea mods used to go directly to the downloads folder, that's cool! Some UI should be built to allow for that same functionality if a mod isn't a work in progress, I may look into that!

For the stuff on private mods, stealing mods, etc. I mostly agree with this

I think things got murky over time, but you're completely right with some of this. When I joined this community in early 2022, the `.tmp` folder was in place. The idea of a "private" mod didn't exist at all. When I hosted work in progress mods I was making, people would steal the mod and host a mod I hadn't finished, which isn't at all good because I don't want that mod public because it's not up to my quality standard and it has a bunch of debug stuff in it. This was happening to other creators as well. So the WIP tag was added for the sole purpose of fixing that, and that it does excellently. The idea of "stealing mods" probably came sometime in that era. We didn't at all have the context behind the original intent, but we knew that work in progress mods were being stolen and hosted without the creators permission, so that's how the idea of stealing mods came to be.

For private mods, over time people *have* wanted to close off mods and not publish them, leaving them for their own servers they host or in 24/7 servers. But the number is very few, so there hasn't been a ton of backlash, and most people don't really care since it hasn't been a big mod that has had that happened. That being said, I agree. If a mod is ready to be released by the standards of the creator, it should be released to the public and the public should be able to use it. There should be no private mods for the sake of being private.

I agree with your message for the most part so long as you understand modders have a right to protect work in progress mods from unwanted distribution, and I think you do.


That all being said, to get to why I am actually here, I'm marking this as draft since we have come to the conclusion that:

> use a mod manifest and allow that to explicitly add/remove files with clean json syntax

is the way to go, and I'll be redoing some of the manifest pr to deal with that

To be clear, this doesn't at all affect the distribution of mods, it only affects the distribution of assets. We don't want to network files that aren't necessary because it is a waste. For raw assets like raw pngs or levels or such, you'll need to go to the mod creator or use decomp tools to decompile them. That's been the case for 4 years now, ever since djui textures were added. We shouldn't network raw assets that aren't use because downloads are already slow enough lol.

Will be marking as draft till the planned implementation is, well, implemented!

anzz1:
> Some UI should be built to allow for that same functionality

I agree, this would make it easier to host mods, without having to move them from the tmp folder, which not everyone knows how to do. There should be a different category for "downloaded" mods though to separate them from your self-installed ones and reduce clutter.

> The exact issue is it can't just be an allowlist since files that previously were not intended to be networked are now networked, think like a README if someone got the mod from GitHub. That shouldn't be sent over the network since it is unused. Or png files, mods have plain png files that shouldn't be networked because they are converted to .tex, that's why a simple fix like that isn't viable. Hopefully that makes sense

Oh, I didn't know this was an issue since I haven't seen any README files and also only the resulting .tex files instead of .png which already seemed to me that exactly the files that are not actually part of the mod weren't being networked. But I just recently cleaned my tmp folder as i sometimes do so I can't confirm this.

> use a mod manifest and allow that to explicitly add/remove files with clean json syntax is the way to go

I don't agree with this, that's extra complication for mod makers. I think what should happen instead is that the game knows what files it needs to run the mod, whatever is included by the mod, those get networked, and whatever isn't used by the mod, is not. So for example it wouldn't network a README file it never uses, nor the .png but only the .tex file. I actually thought this was already a thing and thus didn't realize that unused files were being networked.

xLuigiGamerx:
@anzz1 you said a bunch of stuff and I don't care to reply to it specifically so I'll just tell you this
1. Whatever mod distribution system you think of is gone and the rule i stated about stealing mods has been in effect for a long while, it's literally in the game rules which you would know if you were in the community
2. I am not trying to make a system to private mods, you're just putting words in my mouth, what I'm trying to make is a way for people to organize mods however they like without security risks
3. Your idea for not including unused file is not possible at least with the current system and I don't think it should be changed because you should never network files not used by the mod during runtime, it should only happen on join

If you want to argue more about this then please just get on discord because communicating on github is a pain

Bloxxel64:
yeah, this is clogging up actual pr discussion lol

anzz1:
> Whatever mod distribution system you think of is gone and the rule i stated about stealing mods has been in effect for a long while, it's literally in the game rules which you would know if you were in the community

What are you talking about? It's not gone. Mods are distributed currently with everything intact and they can be re-hosted and edited. I don't know what rules if you are talking about the [mod site](https://mods.sm64coopdx.com/help/rules/) there is no such thing, and that is a site and not related to coopnet. It doesn't even make sense because there is no such thing as "stolen" in an open source project. Only thing that comes to mind could be when DJ banned B3313 due to abundance of caution related to the gigaleak stuff. And that was about proprietary assets, not about mods to an open-source project. 
Even implementing such rule of "no stealing" in an open-source project would be flagrant violation against the intent of the project even though it can't be legally enforced due to it never being put under AGPL like coopnet was, due to already mentioned licensing grey-area. You got it completely inverted, using an open source project and using the work of people that contributed into it and then making a modification and making it closed, *that* is stealing. If you want to make a thing of your own and keep it's source closed, start your own project from scratch, don't use the work of others, and then you can call it yours.

> I am not trying to make a system to private mods, you're just putting words in my mouth, what I'm trying to make is a way for people to organize mods however they like without security risks

If your way to do that includes a possibility to break the mod distribution system so not all assets that are required to run the mod are not cached, resulting in a broken mod being saved to the user, that is effectively exactly what is being done. The result is what matters, not stated intent.

> Your idea for not including unused file is not possible at least with the current system and I don't think it should be changed because you should never network files not used by the mod during runtime, it should only happen on join

This is contradictory, that is exactly what I meant, any files that are not actually used by the mod should not be downloaded. Only the files that the mod uses should be networked. Frankly I thought this was already the case, but it seems not. If its not the case, that should be fixed. 

> If you want to argue more about this then please just get on discord because communicating on github is a pain

Discord sucks because there is essentially no good thread feature and everything is just instant message and gets lost in the backlog of million messages. But I agree that "discussions" would be a better place, than a PR. In any case though, the discussion was relevant to PR if there is attempt to include a possibility to make a mod closed source, which again is against the original intent and frankly a terrible thing to do for a project that was specifically designed to be open-source from day 1. That would be violating against everyone who ever contributed to the project in an assumption that there won't be some takeover of the project by bad actors who would try to close source of things, while using the open source contributors work as the base. This is exactly why those licenses like AGPL exist, to make it illegal to do such thing, but unfortunately this project doesn't include such legal protection. Otherwise someone could copy a project, and stop providing the source to the changes made on top of it. Or in case of making mods to the project, could not give access to the source. Even in the absence of legal protection, it doesn't mean anything going in that direction should be encouraged in any form, so frankly it doesn't even need a "discussion".

xLuigiGamerx:
@anzz1 you're literally wrong, if you want to argue more then stop replying 2 hours later and get on discord so we can have an actual discussion

AgentXLP:
Jesus Christ

Amy54Desu:
why can't we all just love each other

Isaac0-dev:
@anzz1 I've previously asked you to avoid presenting your conclusions as established facts, and I also asked to continue that discussion in Discord. You didn't respond, and the same pattern I was asking about has continued. As I said, strong disagreement is welcome. But presenting personal design judgements and project history interpretations as objective facts, dismissing disagreement as not worth discussing, and using adversarial language is creating an ongoing problem for our discussions and contributors.

This behaviour needs to stop. Your future participation in our community must not continue this pattern.
It's fine to have strong opinions about how the project should be designed, but it needs to be presented as your opinion, not as established fact, and disagreements must remain open to discussion.

This is a formal warning. If this behaviour continues, I will take further moderation action, potentially including removing your ability to participate in the project.

I do believe you have the potential to make useful contributions, and I'm giving you the opportunity to do that. But the current behaviour is making productive collaboration unnecessarily difficult and it needs to stop if you wish to continue participating in our project.

AgentXLP:
I back what Isaac said

anzz1:
I've only now seen this after going through some of my lingering unreads in the notifications.

I will reiterate that I am not "presenting personal design judgements and project history interpretations", you're essentially accusing that I'm trying to rewrite history, which I'm not. When the mod system was originally being designed and then introduced, the whole idea of it was to be an open ecosystem. Like said, there originally was no `.tmp` folder either, you'd join a server to download a mod and then you'd be able to host it yourself and modify as you wish. And even after it was introduced to reduce the amount of clutter, it still didn't change the fact the system was and is still fully open. I don't know if the logs still exist in Discord but there was a dev chat category called "secret" or something like that, where this stuff was discussed back then. I do remember the names Isaac and PeachyPeach coming later on somewhere around or after the lua modding api was released and iirc PeachyPeach made some of the early mods too, I can't remember for certain whether you were added to the group chat. In either case, maybe you weren't around when the early discussions about the design were had since you don't remember this, I'm not trying to rewrite history here, that is the original design, not "my interpretation". On top of that, to be honest, the system being open should already speak for itself that its how it was designed. This whole nonsense of "stealing mods" is something I've never heard back then, as already pointed out, as EmeraldLoc also remembers. There is no stealing in an open source community, it's called sharing, that's the whole point.

And I already said my piece about Discord, its definitely the worse option than GitHub precisely because here there is a much better record of what was said. In Discord, sure there is some history, but it's hard to search and organize, and with channel re-organizations etc. stuff just vanish and then there is no record. This is not a problem with GitHub issues or discussions.

I'm a straight shooter and honest, and I'll speak my mind, this is who I am. I'm of the opinion that people should speak as frankly and honestly as possible, dancing around and speaking vaguely only wastes everyone's time. Speak straight, speak honest, communicate your goals clearly, don't try to hide anything. That's the way I function. I'm not sure why Isaac you're trying to appoint yourself as some czar who controls the discussion. "Formal warning", you're speaking as if you created the whole project. Free speech is the most important achievement of humanity.

In any case though, with all these words from this entire thread can just ultimately be condensed to a simple tl;dr, which is as follows:

**TL;DR;** Do not try to introduce changes that endanger the open-ecosystem design, and there is no problem. Certainly don't try to push some changes without mentioning it clearly that it carries the possibility of transferring partial, broken mods to the player, i.e. ones they can't host themselves or modify when receiving them because they're missing some of the files, breaking the entire system.

That's what its all about. Keeping it open, as originally designed. To keep the mod ecosystem healthy for long time to come.
Although communication is key, so to prevent such disagreements from happening again, probably a contributing guideline should be created to expressly spell this out as it doesn't seem to be so obvious as one would think. You never think you need them until you do. Trying to steer something created as open source to a direction that is more closed, certainly not the first time such thing has happened and not the last one either.

Bloxxel64:
I think I've seen exposure docs shorter than this

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

PR #1376, opened 2026-08-10
Create gLevelValues.disableShadows for better on-time control of actor shadows

Emeraldsniper:
This is part 1 of my overall PR stuff for the lighting engine to control shadows and their distances from lights. To do that, I had to create a struct first in order for it to turn off all actor shadows when the LE is on.

This took longer than it should.. but I'm a total idiot for coding. At least I finally figured it out on my own time. (After maybe too many redos..)

When using `gLevelValues.disableShadows`, it does indeed disable all actor shadows when set to `true`. But does it regardless of the LE being on or not.

So for it to work with how I want it to be, it has to be set to `false` first, then added into a function in order for it to work properly. That way it can only turn off circle shadows when the LE is on.

```lua
gLevelValues.disableShadows = false

function lighting_engine_off()
   le_set_enabled(false)
   gLevelValues.disableShadows = false
   djui_chat_message_create("Lighting Engine: OFF")
   return true
end

function lighting_engine_on()
    le_set_enabled(true)
    gLevelValues.disableShadows = true
    djui_chat_message_create("Lighting Engine: ON")
    return true
end
```

https://github.com/user-attachments/assets/9d339abf-79dc-4510-a81e-2812b69a720e

I first thought I needed another function within the code in order for the level value to even be read properly... but apparently, I don't need anything further than what I already got. So I kinda wasted a large chunk of hours chasing a thread that wasn't even needed. (I'm stupid, I'm trying here..)


src/game/rendering_graph_node.c
```diff
@@ -1285,6 +1285,12 @@ void geo_set_animation_globals(struct AnimInfo *node, s32 hasAnimation) {
  * the floor below it.
  */
 static void geo_process_shadow(struct GraphNodeShadow *node) {
+        if (gLevelValues.disableShadows) {
```
EmeraldLoc: Extra indentation!

EmeraldLoc (changes requested):

src/game/rendering_graph_node.c
Emeraldsniper: Oops. My bad..
That's fixed now.

Emeraldsniper: Or.. Not. I.. must have missed something crucial if it's failing now.

Emeraldsniper: Okay.. pardon my upmost stupidity for reading something completely wrong...
This is all fixed up properly now.

PeachyPeachSM64 (approved):

EmeraldLoc (approved):

PR #1378, opened 2026-08-15
Add `CONTRIBUTING.md` and `BUILDING.md`, cleanup the C code format and conventions, move windows to UCRT

EmeraldLoc:
Quite the number of changes!

- Move building instructions outside the wiki into `BUILDING.md` so people can pr to it and so it generally stays more maintained.
- Create a `CONTRIBUTING.md` file
- Adjusts readme to link to all these correctly
- Move windows to UCRT, recompiled lua 5.3 using UCRT to prevent linker errors (old shell works fine still if you choose to use a deprecated shell)
- Add significantly more to the code format specifications
- Make macOS use normal `cp` instead of `gcp`

The wiki section should probably be changed to redirect to the `BUILDING.md` file


anzz1:
Nonono don't use UCRT its a piece of crap!

What are these "linker errors" you speak of and where did they come from? I'll rather help to fix them, stay far, far, far away from UCRT.
UCRT support was only recently introduced to MinGW after long period of feet-dragging of not wanting to implement it at all, but some people kept needing it so ultimately through gritted teeth it was added. So if you absolutely can not avoid using it for some reason, then you can use it. But if its in any way possible to avoid it, you should!

I'm sure a solution can be found to whatever problem you're having that doesn't require the introduction of UCRT.

Emeraldsniper:
> Nonono don't use UCRT its a piece of crap!
> 
> What are these "linker errors" you speak of and where did they come from? I'll rather help to fix them, stay far, far, far away from UCRT. UCRT support was only recently introduced to MinGW after long period of feet-dragging of not wanting to implement it at all, but some people kept needing it so ultimately through gritted teeth it was added. So if you absolutely can not avoid using it for some reason, then you can use it. But if its in any way possible to avoid it, you should!
> 
> I'm sure a solution can be found to whatever problem you're having that doesn't require the introduction of UCRT.

We're literally switching to this method because of the fact that MINGW was deprecated since March of this year.
https://www.msys2.org/news/#2026-03-15-deprecating-the-mingw64-environment

We can't just stay with MINGW while its deprecated. We needed an alternative. And with UCRT having a more intuitive C runtime on top of it, it was a fair choice trying it out and sticking to the idea. Why be against the idea of moving to UCRT? If you don't like that engine, then what **would** you prefer?

anzz1:
[MinGW](https://www.mingw-w64.org/) is not "deprecated", its in active development. The latest GCC version is 16.2 which was released literally a week ago. Its simply the (bad) decision of the [MSYS2](https://www.msys2.org/) maintainers to drop MinGW in the future, and MSYS2 is a whole different thing, its a development environment that has a toolchain, its not a toolchain in of itself. *They* are dropping MinGW, not anyone else.

> If you don't like that engine, then what would you prefer?

Sticking with MinGW, and ditching MSYS2. There are plenty of [alternatives](https://www.mingw-w64.org/downloads/) available, like simply using the mingw cross-compilation toolchain that already exist in ubuntu packages (the distro which github runners use, although Arch, Debian and Fedora also package it), or if building the windows binary on windows like now, there are plenty of options too like [w64devkit](https://www.mingw-w64.org/downloads/#w64devkit), [prebuilt MinGW-W64-builds](https://www.mingw-w64.org/downloads/#mingw-w64-builds), [WinLibs](https://www.mingw-w64.org/downloads/#winlibscom) or [LLVM-MinGW](https://www.mingw-w64.org/downloads/#llvm-mingw). There is also a standalone MinGW-w64 toolchain available via the [Chocolatey](https://push.chocolatey.org/packages/mingw) windows package manager, which this [setup-mingw](https://github.com/egor-tensin) GitHub Action uses.

Of course, there isn't any rush with any of this since "deprecated" doesn't mean "removed". But it would make sense to make a draft PR and put it on ice to be pulled out when its actually time to "deprecate" MSYS2.

EmeraldLoc:
Packages are no longer being updated on MSYS2 mingw64, they are being updated on MSYS UCRT64. I don't really use windows (it's garbage), but seeing as I was updating the docs I'd figure I'd quickly recompile lua and have the docs point to using UCRT64 instead of the deprecated MSYS2 mingw64 system for future-proofing.

You can still use the deprecated system just fine, everything still works over there, you can compile without msys2 as well, it's just we point to the updated one in the build documentation, and since UCRT64 compiles quicker than mingw64, we changed it in actions as well. For alternatives, I won't bother looking into them since idc about windows, so yea

anzz1:
I kinda figured you probably don't even use Windows considering you'd even suggest UCRT and not knowing what you're getting yourself into. :smile: [This is going to be your future with UCRT](https://www.google.com/search?q=api+ms+win+missing)

It's also not about whether (you) can use the "deprecated" system which isn't "deprecated" as I already mentioned but rather what the windows distribution build should be built with, what the GitHub Action should be, and that is definitely not UCRT its gonna create a dependency hell that MinGW simply doesn't have. You might not be using windows, but the majority of coop players are, and probably 99% of them are not building from the source but using the binary distribution.

I can look into a proper alternative to MSYS2 and making a draft PR to put on ice to be pulled up when its actually necessary, when the "future" actually comes up. The question is whether the windows version should be cross compiled or compiled natively? As if compilation speed is what you're after, as far as I know the linux runners are faster and will compile windows targets faster than compiling them on windows. But then you obviously can't run any tests, but since there currently isn't any that shouldn't be a problem.

PeachyPeachSM64 (approved):

PR #1383, opened 2026-08-21
Use Andrat's Most Recent Luigi Recordings

wbmarioo:
Provides Andrat's most recent recordings of Luigi's voice from Render96, by using Isaac's restore sound generate patch.


wbmarioo:
Will work on this later, I don't think there's an good way to get this working at the moment

PR #1385, opened 2026-08-21
Main menu level rework, add Castle level and Inside the Castle music

EmeraldLoc:
It wasn't long ago I said it's never a good idea to program at 2am in the middle of the night, but here we are

- Redoes the menu level system. Now it originates form a struct containing level data, so it's all done in one centralized place
- Behaviors are prevented from spawning rather than being killed each frame (yw peachy 😁)
- Added a castle level
- Added the castle music
- Make music persist throughout level changes

Seeing as I'm tired as hell, maybe I made a mistake here, maybe not, review to find out!!! (I think I cooked)


src/game/spawn_object.c
```diff
@@ -379,6 +380,20 @@ struct Object *create_object(const BehaviorScript *bhvScript) {
...
+        struct MenuLevel *menuLevel = get_menu_level();
+        const BehaviorScript **bhvToRemove = menuLevel->behaviorsToRemove;
+
+        while (*bhvToRemove != NULL) {
+            if (*bhvToRemove == behavior) {
+                return NULL; // menu level does not want this object spawned
+            }
+
+            bhvToRemove++;
+        }
```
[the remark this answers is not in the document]
EmeraldLoc: Yea, array should be NULL terminated, so max capacity it really more like 7. I wouldn't statically allocate if C syntax would allow it, but it doesn't

src/game/spawn_object.c
```diff
@@ -379,6 +380,20 @@ struct Object *create_object(const BehaviorScript *bhvScript) {
...
 
+    if (gDjuiInMainMenu) {
+        // iterate through behavior scripts
+        struct MenuLevel *menuLevel = get_menu_level();
+        const BehaviorScript **bhvToRemove = menuLevel->behaviorsToRemove;
+
+        while (*bhvToRemove != NULL) {
+            if (*bhvToRemove == behavior) {
+                return NULL; // menu level does not want this object spawned
+            }
```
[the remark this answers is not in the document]
EmeraldLoc: This is in the main menu, this is fine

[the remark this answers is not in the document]
EmeraldLoc: When I say this is in the main menu I mean no custom behaviors exist in the main menu, so the behavior always points to the vanilla behavior script

autogen/lua_definitions/constants.lua
```diff
@@ -3558,6 +3558,9 @@ LEVEL_COUNT            = 39 --- @type LevelNum
 --- | `LEVEL_UNKNOWN_38`
 --- | `LEVEL_COUNT`
 
+--- @type integer
+MAX_BHVS_TO_REMOVE_IN_MENU_LEVEL = 8
```
PeachyPeachSM64: don't expose this

src/game/level_update.c
```diff
@@ -182,6 +181,281 @@ struct CreditsEntry sCreditsSequence[] = {
...
+        .level = LEVEL_CASTLE_GROUNDS,
+        .area = 1,
+        .act = 1,
+        .marioPos = { -1328, 260, 4664 },
+        .updateMarioPos = false,
+        .cameraPos = { -1328, 390, 6364 },
+        .updateCameraPos = true,
+        .marioYaw = 0,
+        .updateMarioYaw = true,
+        .behaviorsToRemove = NULL
```
PeachyPeachSM64: That's confusing. `behaviorsToRemove` is never NULL, it's actually initialized to `{ NULL }`. C allows this, but it's not a very good practice to do that.

src/game/level_update.c
```diff
@@ -182,6 +181,281 @@ struct CreditsEntry sCreditsSequence[] = {
     { LEVEL_NONE, 0, 1, 0, { 0, 0, 0 }, NULL },
 };
 
+struct MenuLevel gMenuLevels[] = {
+    {
+        .name = "Castle Grounds",
```
PeachyPeachSM64: Wouldn't this need to be translated?

src/pc/djui/djui_panel_menu_options.c
```diff
@@ -35,27 +35,7 @@ struct MainMenuSounds gMainMenuSounds[] = {
...
-    "LLL",
-    "THI",
-    "HMC",
-    "CCM",
-    "RR",
-    "BITDW",
-    "PSS",
-    "TTC",
-    "WDW"
+    { "Inside the Castle", SEQ_LEVEL_INSIDE_CASTLE },
```
PeachyPeachSM64: Translating these as well?

PeachyPeachSM64 (changes requested):

src/game/level_update.c
EmeraldLoc: Oh yea no that's a mistake on my part. I had to constantly change this type because C initializers simply couldn't do what I ultimately wanted. I'll fix this

src/game/level_update.c
EmeraldLoc: Look... They weren't being translated before, and these are sm64 stages so it doesn't matter. And the playerlist does not translate them. And uhhhh mods, yea mods don't translate at all too, probably.

Please don't make me translate these 😭

src/pc/djui/djui_panel_menu_options.c
EmeraldLoc: These weren't being translated before, spare me 🙏

src/game/level_update.c
EmeraldLoc: K added lang entries

src/pc/djui/djui_panel_menu_options.c
EmeraldLoc: K added lang entries

PR #1392, opened 2026-08-27
Add Shift+TAB backwards tab completion, command preview, and command keybind

iZePlayzYT:
Split from #954. This is the chat-interaction slice: tab completion, command input, and sent-history QoL. UI restyle and chat settings are separate follow-up PRs.
- **Shift+TAB**: cycle command, subcommand, and player-name completions backwards (TAB still goes forwards)
- **Preview**: the next match is shown as gray suffix text in the input box before you accept it
- **Chat (Command)**: extra keybind that opens chat with `/` already typed, so you can jump straight into a command.
- **Sent history**: ↑/↓ skips consecutive duplicate entries, so repeating the same command does not clog the list
- **`/help`**: header lists all commands and mentions TAB / Shift+TAB, localized in every language


src/pc/djui/djui_chat_box.c
```diff
@@ -170,10 +175,29 @@ static char* get_main_command_from_input(const char* input) {
...
-static bool complete_subcommand(const char* mainCommand, const char* subCommandPrefix) {
+static s32 tab_completion_cycle_index(s32 current, s32 count, bool reverse) {
+    if (count <= 0) { return -1; }
+    if (reverse) {
+        return (current <= 0) ? (count - 1) : (current - 1);
+    }
+    return (current + 1) % count;
+}
+
+static void free_string_list(char** list) {
```
EmeraldLoc: Don't know if I entirely think you need to abstract a function for this, at least in my opinion. If you do want to have this, use it everywhere in this file, you don't use it in a few places Ill mention later in the review

src/pc/djui/djui_chat_box.c
EmeraldLoc: If you want to have `free_string_list`, use it here!

src/pc/djui/djui_chat_box.c
```diff
@@ -170,10 +175,29 @@ static char* get_main_command_from_input(const char* input) {
     return command;
 }
 
-static bool complete_subcommand(const char* mainCommand, const char* subCommandPrefix) {
+static s32 tab_completion_cycle_index(s32 current, s32 count, bool reverse) {
```
EmeraldLoc: Instead pass a pointer in for the current, so `s32 *current`. That way you don't need to do `x = tab_completion_cycle_index(x, ...)` and you can instead do `tab_completion_cycle_index(&x, ...)`

src/pc/djui/djui_chat_box.c
```diff
@@ -170,10 +175,29 @@ static char* get_main_command_from_input(const char* input) {
...
+
+static void free_string_list(char** list) {
+    if (list == NULL) { return; }
+    for (s32 i = 0; list[i] != NULL; i++) {
+        free(list[i]);
+    }
+    free(list);
+}
+
+static bool complete_subcommand(const char* mainCommand, const char* subCommandPrefix, bool reverse) {
```
EmeraldLoc: Am aware this isn't introduced by this pr, but doing `subCommandPrefix` is inconsistent with the rest of the codebase. If you could use `subcommandPrefix` instead, that'd be great :D

src/pc/djui/djui_chat_box.c
EmeraldLoc: Use your free string list func here or alternatively drop it, your call

src/pc/djui/djui_inputbox.c
```diff
@@ -634,6 +634,23 @@ static bool djui_inputbox_render(struct DjuiBase* base) {
         djui_inputbox_render_char(inputbox, c, &drawX, &additionalShift);
         c = djui_unicode_next_char(c);
     }
+
+    bool isChatInput = (gDjuiChatBox != NULL && gDjuiChatBox->chatInput == inputbox);
+    if (isChatInput && djui_interactable_is_input_focus(&inputbox->base) && inputbox->buffer[0] == '/') {
+        char* previewText = get_next_tab_completion_preview(inputbox->buffer);
+        if (previewText != NULL && previewText[0] != '\0') {
+            gDPSetEnvColor(gDisplayListHead++, 128, 128, 128, 180);
```
EmeraldLoc: For respecting djui themes, could you get the current theme and use `theme->interactables.disabledTextColor` for this?

src/pc/commands.c
```diff
@@ -161,6 +161,11 @@ static void chat_construct_player_message(struct NetworkPlayer *np, char *msg) {
 }
 
 static bool command_help(UNUSED const char *message, bool onConsole) {
+    char tabcompletionHint[256];
+    snprintf(tabcompletionHint, sizeof(tabcompletionHint), "\\#ff2020\\%s \\#ffa020\\(%s)\\#ff2020\\:\\#000000\\",
```
EmeraldLoc: First of all, the colors here are awful lol, I don't like the red and orange at all. Secondly, I don't think this is needed, controls like this should be intuitive enough. I'm not against having a message for this but it can't be this color, keep it the normal color

EmeraldLoc (changes requested):
Think I got everything for my review

Generally though, follow [these conventions](https://github.com/coop-deluxe/sm64coopdx/blob/main/docs/c/c.md) in this pr, as right now code here doesn't follow it

This is a much appreciated QoL feature and will make me actually use the autocomplete feature!

src/pc/djui/djui_chat_box.c
iZePlayzYT: Kept free_string_list and now use it for every string-list cleanup.

src/pc/djui/djui_chat_box.c
iZePlayzYT: Done, this now calls free_string_list.

src/pc/djui/djui_chat_box.c
iZePlayzYT: Updated to take s32 *current, so the call sites are now tab_completion_cycle_index(&x, ...).

src/pc/djui/djui_chat_box.c
iZePlayzYT: Renamed to subcommandPrefix

src/pc/djui/djui_chat_box.c
iZePlayzYT: Done, this uses free_string_list as well.

src/pc/djui/djui_inputbox.c
iZePlayzYT: Preview text now uses theme->interactables.disabledTextColor.

src/pc/commands.c
iZePlayzYT: Kept the hint, but dropped the custom colors so it uses the normal info color.

EmeraldLoc (approved):

EmeraldLoc:
Were these translations done using a machine translator or AI? If so, mark it at the end with a ;

See [lang docs](https://github.com/coop-deluxe/sm64coopdx/blob/dev/docs/c/sections/lang.md) for more information

PR #1393, opened 2026-08-27
Redesign chat box layout, message styling, and character counter

iZePlayzYT:
Split from #954. Stacked on the chat-interactions PR https://github.com/coop-deluxe/sm64coopdx/pull/1392
- **Layout**: more gap between the message list and the input, less padding around messages, and a borderless input bar at the bottom
- **Messages**: focused chat uses one shared background instead of a box per line; closed chat still fades messages out
- **Input**: dark, borderless field with white text; selected text stays white so it remains readable
- **Character counter**: live count to the right of the input, shifting color as you approach the limit
- **Follow-up ready**: existing messages can be restyled in place for the later size/opacity settings PR


EmeraldLoc:
Screenshot 2026-08-27 at 12 34 37 AM
Screenshot 2026-08-27 at 12 34 30 AM

This is what it looks like. Can't say I'm a huge fan. As it stands I like the current input box and how it follows the theme, so I don't think I'd mind if you made the chat window also follow the theme. Other than that, I don't know how much I like this, it doesn't really follow djui. Making it one djui panel I'm cool with so long as there are dividers. Honestly though I think the current look is solid and really just needs to be changed to follow the current djui theme, and that's all

iZePlayzYT:
@EmeraldLoc how about putting the design in the chat settings from PR3 as a toggleble flat/basic Look Option?

EmeraldLoc:
Nah I disagree with that. If anything, wait for the Theme pr to be merged and you can add keys to allow for chat customization in the djui theme, that way you can do the colors so the ui can be flat. I think a lot of the settings in PR3 aren't great as well, I'm generally just not the largest fan of these changes. Another developer can chime in, but that's what I think

iZePlayzYT:
> Nah I disagree with that. If anything, wait for the Theme pr to be merged and you can add keys to allow for chat customization in the djui theme, that way you can do the colors so the ui can be flat. I think a lot of the settings in PR3 aren't great as well, I'm generally just not the largest fan of these changes. Another developer can chime in, but that's what I think

Whats with PR 1 thats just an improvement of my old tab completion that got merged a long time ago?

EmeraldLoc:
Oh I am referring to this and pr3, pr1 looked great when I took a glance at it

PR #1395, opened 2026-08-27
Fixed wrong translation

iZePlayzYT:


EmeraldLoc:
Hopefully you're right!

Issue #1243, opened 2026-05-13
[Feature Request] External Socket API


bluelightzero:
The External Socket provides mods **the limited ability to send UDP-like packets** to programs outside of CoopDX in a similar way  to how ModFS lets you write external files.

You would first launch CoopDX with `--external-port PORT` where `PORT` is your external program's port.

A Lua mod can call `external_socket_is_connected()` to check that it is able to send packets.

If the feature is enabled then the Lua Mod can call either `external_socket_send(dataTable)` or `external_socket_send_bytestring(bytestring)` to send a packet of data to the external application.

The Lua mod can also hook an event for `HOOK_ON_EXTERNAL_PACKET_RECEIVE` or `HOOK_ON_EXTERNAL_PACKET_BYTESTRING_RECEIVE`. When the external program sends a packet back to CoopDX, this hook will trigger.

------------------------

To prevent misuse, not only must `--external-port PORT` be set, but the external program that listens on a local port must accept a special packet with a CoopDX header, and respond like-wise with a specific packet.

Once CoopDX receives the correct packet back then mods will be able to send and receive packets.

Unlike plain UDP packets, each packet will have a CoopDX header.

----------------

The whole idea of implementing this is to reduce the need to create hacky analogues using ModFS.

I would be happy to start working on this in the near future.

AgentXLP:
Hmm...

AgentXLP:
I can't give any guarantees, and it's something I would need to speak to the development team to get them on board, but I personally give you the green light and think you should try and implement this. Let's see where it goes!

ThePlayerRolo:
could this feature theorectically allow for stuff like custom leaderboards since you could send that data to the socket?

bluelightzero:
@ThePlayerRolo anything that could be accomplished with this could already be done through a hacky ModFS message system.

Leaderboards could already be done now by saving the data on the host through ModFS, and if you wanted multiple lobbys to access it, then you could have an external application that syncs the ModFS.

bdalmlkalbry880-sketch:
_### _

> **_**### **_**###**_ ### ### **_**_****_**_******_**

__

> _**__******_

PR #1396, opened 2026-08-27
mod_socket implementation

gingerphoenix10:
Implements a basic system for mods to communicate with UDP / TCP sockets from mods, based on the #1243 feature request.

Sockets are disallowed by default, and must be enabled on a per-user basis in `Options -> Misc -> Allow External Requests`. Mods can check this with `mod_socket_allowed()`, and any request to connect to a socket will return false (failure) if disabled. Toggling the option off mid-game, or leaving the lobby, will instantly disconnect from the connected socket.

If sockets are allowed, mods can run `mod_socket_init(IP, Port, Mode)` where Mode is true for TCP and false for UDP, which will attempt to connect to the socket at that address (or point the current IP and Port to it for UDP). Currently this connection is blocking for simplicity, however it'll likely be moved to a hook so reconnection attempts don't freeze the game each time.

Once the socket is connected (which can be checked via `mod_socket_is_connected()` if using TCP, since UDP doesn't have handshakes and must use a heartbeat to track connection state), mods can send string messages to the socket via `mod_socket_send(data)`, or disconnect via `mod_socket_disconnect()`. Messages can also then be received via the `HOOK_SOCKET_RECEIVE` hook, which gives the string received by the socket.

Example of a simple mod that connects to a (non-existant) socket and echos any message it receives back to the server:
```lua
IP = "12.345.67.890"
Port = 12345
TCP = true

connSuccess = mod_socket_init(IP, Port, TCP)
print("Was success: "..tostring(connSuccess))

if mod_socket_is_connected() then
    -- Since mod_socket_init is currently blocking, this will return whether it succeeded or not right after attempting
    print("Connected to external socket!")
end

function rec(data)
    print("Received \""..data.."\". Echoing")
    mod_socket_send(data)
end

hook_event(HOOK_SOCKET_RECEIVE, rec)
```

### Example use cases:
Connecting to a socket on the local machine can allow for developers to create a mod that connects to a custom app built specifically for it, giving a practical method for integrating something like positional data for a voice chat client, or anything else that could need data from in-game

Connecting to an outside socket can allow mods to have something like leaderboards on an external server that can be added to from any co-op lobby, including private lobbies to play solo without anyone being able to join.

Lua can be used to create the WebSocket protocol on top of the TCP this PR adds, allowing for something like Archipelago randomizers to be built fully into the game without needing an external client to forward messages back and forth between the game and the server

### Other info
This is currently in a draft state and not finished, as I want to get the team's thoughts on this earlier rather than later. Even with the requirement of requiring the user to allow it for it to function, I could understand if this could be seen as too much to give in the lua api. I decided to make it as versatile as it currently is so that it could be used to make projects like shown above without needing any hacky solutions with connecting to different things to get it to work, however it could be locked down to be more akin to #1243 if requested.
I'm also fairly new to C / C++, so if there's any bad practices noticed in my code, please point them out to me so I can get them fixed.

Everything this can do could *technically* be done previously via an extremely hacky ModFS system with an external program to read and write from those files, however this system is infinitely better and more reliable than that was, and doesn't cause unnecessary read/write cycles.

The things I currently have planned at minimum before this could reasonably be a finished feature:
- Better handling of multiple mods using sockets (currently there's only 1 possible socket shared between every mod, so whichever connects last gets priority and the other mods have no indication that the owner has changed. Maybe give each mod their own single socket?)
- Give a dialogue when trying to turn on external requests warning players that it's allows mods to use their local network for sockets
- Add functions for checking the current state of the socket i.e. ip, port, mode. (maybe not strictly necessary after multiple mod handling is done since mods can track that themselves, but can't hurt to add)
- Allow hostnames to be used rather than specific IP addresses
- Mentioned above, disable blocking for `mod_socket_init`
- Add a second hook for returning a Bytestring instead of a regular string
- IPv6 support
I'll also probably adjust some of the function names since having things like `mod_socket_allowed` vs mod_socket_***is***_connected doesn't make much sense considering they're both just checking and returning a value


gingerphoenix10:
Force pushed to rename that commit as it didn't specify it was for mod_socket

bluelightzero:
Forgive me if this has already been addressed, but what prevents abuse by a random lobby host from making a unwanted connection once that setting is turned on?

You can't trust that most players are smart enough to know the risks of enabling the "Allow External Requests", and even someone that knows the risks might forget to turn it off when joining a public lobby with a malicious mod.

I specifically made the original suggestion to require that the player must explicitly use a launch argument. This can be used by an external application to launch the game with the correct settings for convenience and only for a very specific endpoint.

I think making this an in game setting is a mistake, and giving this much freedom in the API is also a mistake.

I think allowing Lua mods to specify the destination is also a mistake. People do not expect mods to have these types of privileges.

I do think that it should allow the option to specify the address in the launch command (eg `--external-socket localhost:87263 `)

EmeraldLoc:
Wowie. Ok, great on thinking about the implications of this and locking this behind a setting. I don't think this will be merged though, this has too much power. I'll let another developer chime in, but that's what I think.

Currently there's a list of conventions we have internally, but nothing has been added to main yet, and the latest one hasn't been merged. But you can find a list to the latest conventions [here](https://github.com/EmeraldLoc/sm64coopdx/blob/more-documentation-weewoo/docs/c/c.md), other than that I'd need to issue a full review on the C code to get an idea of how your code looks, but I haven't done that yet, and wouldn't want to waste the time until we decide if this is something we want to merge

> I think allowing Lua mods to specify the destination is also a mistake. People do not expect mods to have these types of privileges.

I disagree with that. If this feature ends up allowed, lua should be able to change the port and ip.

gingerphoenix10:
(replying to @bluelightzero)
Yeah, that's fair. Forgot to mention it in the post but I was also planning to make it only allow connections from the host so that joining a random co-op lobby can't just connect to anything whilst letting it be used for singleplayer sessions or relying on the host for communication (and if the socket really needs data from other players, the host can just pull most data you'd need from everyone else's MarioState)
Having a warning before you can accept external sockets should help reduce the problem of people not knowing what they're too enabling, but I can still see the concerns with people ignoring it or forgetting it's on in the first place (could make it a thing that needs to be enabled per session if it ends up sticking with a config option)

Originally I was going to implement it the way suggested in the issue, just using a launch argument for port and requiring it to be locally hosted (which is also why the current implementation currently just has 1 shared socket), but I felt that requiring a custom app for it to have any function would kinda drive people away from using / making mods that use it due to the extra setup required. The suggestion of allowing external addresses but still via the launch parameter *could* work, but I still think that suffers from the same problem that people aren't gonna want to use a mod that requires them to mess with how they launch co-op, as most mods are just as simple as enable and play.

Having it also be defined by a launch argument would also mean that you're limited to only having a single mod enabled that uses it, so players would just have to pick one or the other in the case that they have multiple they want to use together. In the example use cases in the PR, you'd only be able to pick one of those three mods to use in your lobby, otherwise only the one who's expected IP is the same as the one connected would actually function. It also means that if a mod developer doesn't actually *check* what IP is currently being used, it could end up sending random data to an address that doesn't expect it

But yeah, that's why this is still in a draft state. Getting different opinions before committing to a finished version helps

bluelightzero:
> I disagree with that. If this feature ends up allowed, lua should be able to change the port and ip.

I only agree that it is fine as long as the application the mod is connecting to has to acknowledge that it is connecting to coopdx. Otherwise it has too much room for exploiting.

This will also solve the multiple mods issue.

EmeraldLoc:
> I only agree that it is fine as long as the application the mod is connecting to has to acknowledge that it is connecting to coopdx. Otherwise it has too much room for exploiting.

I think you mean the other way around, coop alerting the user it's trying to connect to a specific ip and port, but yea, I could see that 100%

bluelightzero:
> I think you mean the other way around, coop alerting the user it's trying to connect to a specific ip and port, but yea, I could see that 100%

No. I think user warnings are a terrible solution to the problem. People that just want to play the game will ignore it.

gingerphoenix10:
Could go both ways - Co-op player acknowledging the connection, and the server acknowledging that it's co-op trying to connect, however the second way couldn't really stand on its own very well since a custom server could always be made that acts as a middle-man that sends co-op what it wants to see, and sends requests from co-op to the server after removing any co-op specific headers and such

bluelightzero:
> Could go both ways - Co-op player acknowledging the connection, and the server acknowledging that it's co-op trying to connect, however the second way couldn't really stand on its own very well since a custom server could always be made that acts as a middle-man that sends co-op what it wants to see, and sends requests from co-op to the server after removing any co-op specific headers and such

The point is to prevent a mod from connecting to something that it shouldn't.

By requiring the application to acknowledge, it will be the only thing the mod could ever connect to.

gingerphoenix10:
> People that just want to play the game will ignore it.

I think that's sort of the point though. If anyone who's worried enough about what it's connecting to can notice that and quit the mod or inspect further, but most players who don't really mind can ignore it and continue like any other mod

EmeraldLoc:
Yea, we can't enforce an application needing to acknowledge its goal of connecting to coopnet, so I don't really understand that. I'd be nice, but not guaranteed at all.

> No. I think user warnings are a terrible solution to the problem. People that just want to play the game will ignore it.

In one way I agree, in another way, we can do things to mitigate this. Things like forcing the user to wait 10 seconds to hit Yes are very effective in getting users to read. That'll move responsibility off us onto them

bluelightzero:
> Yea, we can't enforce an application needing to acknowledge its goal of connecting to coopnet, so I don't really understand that. I'd be nice, but not guaranteed at all.

Yes you can? Coop DX sends the initial packet to make the connection and waits for the application to send the correct response before allowing the mod to do anything over the connection.

EmeraldLoc:
Oh my apologies. You're not saying this from an alert on the app perspective but not allowing coopnet to connect to a random server running on your computer. Yea absolutely we should do that, good idea!

I think I read acknowledge and just said oh yea he wants to add an alert to the app instead lol

gingerphoenix10:
Yeah, I forgot about that part actually, that was something that crossed my mind but thought I'd get back to it later.
That could be a local-specific thing, so if you're trying to connect to a local server then it requires that header, but that then just gets confusing that it sometimes requires it, other times doesn't. I feel that even if it's implemented though, there's no saying that even with that header it doesn't mess something up since It'd still *attempt* to connect at the least, even if it can't send specific data to it

bluelightzero:
Making the IP:PORT external (such as using a launch command) allows you to configure it externally, without having to write an entire mechanism to configure it through a mod.

The IP:PORT might be different for different users of the mod.

Maybe coopdx could have a launch config file thing so you can do `./sm64coopdx --config myconfig.json` to cut down on the launch parameters, but that is outside the scope.

PR #1400, opened 2026-08-30
Fix incorrect paths for DynOS builtin tex

GhostlyDark:
sm64coopdx isn't able to pick up the bitdw skybox textures, as sm64ex mistakenly labelled them as `bidw` instead of `bitdw`.
Also fixed the cake screen tiles not being converted.

These are breaking changes but fix the compatability with sm64ex texture packs.


PeachyPeachSM64:
Hi! We can't merge this since, like you said, it breaks existing texture packs.
Ideally both names should be allowed.

GhostlyDark:
> Hi! We can't merge this since, like you said, it breaks existing texture packs. Ideally both names should be allowed.

It's a renmant of https://github.com/coop-deluxe/sm64coopdx/commit/f8230fe484c37e833869f7a7a76d63308ce67d4d where the bitdw skybox name differences went unnoticed.

Is it possible to do these `define_builtin_tex` with fallbacks/variants? If so, how?

PeachyPeachSM64:
> > Hi! We can't merge this since, like you said, it breaks existing texture packs. Ideally both names should be allowed.
> 
> It's a renmant of [f8230fe](https://github.com/coop-deluxe/sm64coopdx/commit/f8230fe484c37e833869f7a7a76d63308ce67d4d) where the bitdw skybox name differences went unnoticed.
> 
> Is it possible to do these `define_builtin_tex` with fallbacks/variants? If so, how?

From what I can see in the source, the `path` field of the structure is only used in `DynOS_Builtin_Tex_GetNameFromFileName`, which is called when generating texture packs, and return a texture name from a filename.
Which means that you should be able to have the same texture entry for different paths. Keep the entries you added, but don't remove the existing ones.

GhostlyDark:
I duplicated things instead and added comments for future reference.

EmeraldLoc:
What? Can we not add in a typo for compatibility with ancient sm64ex packs? I'd much rather whatever packs are broken to be changed, it's not like these packs are already compiled into DynOS, they're just the raw files so you can change them. Of course it sucks the bitdw skybox will be broken for many of these packs, but I don't think we should intentionally add a typo when it's been 4 years since these packs were able to be used in this way. That's just my opinion though

Cake screen tiles are cool though, so that can def stay :D

GhostlyDark:
That typo wasn't meant to be fixed and I believe the texture mappings are supposed to be whatever sm64ex and pretty much all its forks are using. The packs aren't broken, the DynOS conversion is.

Not changing anything requires duplicating textures (as renaming breaks it the other way) for **current** sm64ex packs, which I would very much like to avoid for everybody involved, myself included.

paradox9955:
The typo fix was very much intentional, see commit #521b0ab . While my opinion is irrelevant, and I'm all for comparability btw, I have to agree with Emerald. We shouldn't really have to add back the typo for compatibility because of some mistake the sm64ex team made. Plus, the texture packs can and should just be updated to the correct system anyways because almost no one even uses sm64ex anymore.

PeachyPeachSM64:
> What? Can we not add in a typo for compatibility with ancient sm64ex packs? I'd much rather whatever packs are broken to be changed, it's not like these packs are already compiled into DynOS, they're just the raw files so you can change them. Of course it sucks the bitdw skybox will be broken for many of these packs, but I don't think we should intentionally add a typo when it's been 4 years since these packs were able to be used in this way. That's just my opinion though
> 
> Cake screen tiles are cool though, so that can def stay :D

Honestly there is no cost in adding those paths, since DynOS seems to allow multiple paths for the same texture. No risk to break the existing behavior and no need for users to manually fix the old packs (which probably aren't maintained anymore).

AgentXLP:
I agree with Peachy

PR #1406, opened 2026-09-01
Fix Windows console colors on conhost

Cooliokid956:
Before:
image

After:
image

Credit to https://stackoverflow.com/a/79686178/15417580 stack overflow always got you

### DIFF
--- a/src/pc/pc_main.c
+++ b/src/pc/pc_main.c
@@ -520,6 +520,10 @@ int main(int argc, char *argv[]) {
         console = pcount > 1;
     }
     if (console) {
+        HANDLE out = GetStdHandle(STD_OUTPUT_HANDLE);
+        DWORD mode = 0;
+        GetConsoleMode(out, &mode);
+        SetConsoleMode(out, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
         SetConsoleOutputCP(CP_UTF8);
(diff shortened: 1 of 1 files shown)


EmeraldLoc:
> stack overflow always got you

Until it doesn't :(

src/pc/pc_main.c
```diff
@@ -520,6 +520,10 @@ int main(int argc, char *argv[]) {
         console = pcount > 1;
     }
     if (console) {
+        HANDLE out = GetStdHandle(STD_OUTPUT_HANDLE);
```
EmeraldLoc, line 523: This can return `NULL`/`INVALID_HANDLE_VALUE`, may be worth checking if it does since `GetConsoleMode` will from what I can tell crash if it takes in a null output handle.

The result of `GetConsoleMode` can also be validated, as a nonzero return indicates success, but that isn't strictly required and shouldn't cause any horrendous issues if it fails. If it fails something is wrong already so

EmeraldLoc (changes requested):

src/pc/pc_main.c
EmeraldLoc: No, this review is false, never mind. I got it confused with the second param

EmeraldLoc (approved):

PR #1407, opened 2026-09-01
Implement FastDL (optional mod download over HTTP)

PiantaMK:
## FastDL: optional HTTP(S) mod downloading with MD5 verification

I came up with the idea due to having to wait multiple minutes while downloading bigger mods when joining servers for the first time.

The host can advertise an HTTP(S) URL, and clients
download missing remote mod files from it instead of pulling every byte
through the in-game chunk transfer. This makes hosting modded lobbies far
more practical, especially for large mods - any static file server (nginx, a CDN, GitHub Pages, ...)
can serve the files, and the in-game transfer only handles leftovers.

### How it works

- If the host has set a Fast Download URL in Host Settings, it is advertised
  to clients in `PACKET_MOD_LIST`.
- The client fetches each file the mod cache doesn't already cover.
- Files stream to a `.part` file while an MD5 is computed on the fly and
  compared against the hash the host already advertises per file. Only HTTP
  200 is accepted, the transferred size must match the advertised size
  exactly (0-byte files included), and only on a full match is the file
  renamed into place and registered in mod.cache - so it is skipped on
  future joins like any other cached file. WIP mods keep their existing
  never-cached behavior.
- If HTTP was unavailable or some files failed, everything that didn't
  verify falls back to the regular in-game chunk transfer and the join
  continues as before - FastDL is strictly an accelerator, never a
  requirement.

### Player-facing behavior

- New "Fast Downloads (HTTP)" setting in Misc: Always Off / Let Me Choose
  (default) / Always On. Progress is shown in the existing download screen
  with an ETA.
- In Let Me Choose, a popup asks once per connection attempt before the
  first file (showing the URL and total size). The answer sticks across the
  join flow's reconnect cycles and is only reset on a full disconnect.
- Server-level failures (DNS failure, connection refused, timeout, e.g. the
  HTTP server not being started) are detected fast and return back to the
  in-game path entirely. File-level failures (404, size/hash mismatch)
  fall back per file with a count notice.
- Windows uses WinINet (mirroring update_checker.c) with short timeouts and
  internal connect retries disabled; other platforms use libcurl.

### Implementation notes

- Downloads run on one detached worker thread that is never joined: a
  generation counter aborts stale workers, and the main thread owns all
  result application and cleanup (contexts freed on the main thread only,
  via a done flag + pending-free slot), so shutdown can never race the
  worker or leak a transfer.
- All main-thread integration is three hooks: URL read in
  `network_receive_mod_list`, the takeover gate in
  `network_receive_mod_list_done`, and `fastdl_update`/`fastdl_shutdown` in
  network.c.

### Testing

- Compiled and ran on Windows (WinINet path) and Linux (libcurl path).
- Manually tested: FastDL join from a local HTTP server, unreachable /
  offline HTTP server fallback, per-file fallback, ask-popup flow, host
  settings input, progress + ETA display.

### DIFF
--- a/lang/Czech.ini
+++ b/lang/Czech.ini
@@ -30,6 +30,8 @@ LOBBY_HOST = "host lobby"
 UPDATE_AVAILABLE = "Nová aktualizace!"
 LATEST_VERSION = "Nejnovější verze"
 YOUR_VERSION = "Vaše verze"
+FASTDL_FAILED = "\\#ffa0a0\\FastDL:\\#\\ @ file(s) failed, using in-game transfer"
+FASTDL_UNREACHABLE = "\\#ffa0a0\\FastDL:\\#\\ could not connect to '@', using in-game transfer"
 
 [CHAT]
 KICKING = "Vyhazování '@'!"
@@ -259,6 +261,7 @@ BOUNCY_BOUNDS_ON = "Zapnuto"
 BOUNCY_BOUNDS_OFF = "Vypnuto"
 BOUNCY_LEVEL_BOUNDS = "Omezení odražejícího úrovně"
 AMOUNT_OF_PLAYERS = "Počet hráčů"
+FASTDL_URL = "Fast Download URL"
 PAUSE_ANYWHERE = "Pozastavit kdekoli"
 
 [HOST]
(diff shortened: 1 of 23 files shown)


Cooliokid956:
@coderabbitai try review

Cooliokid956:
we are Not doing this

EmeraldLoc:
Hi

This is quite cool, but this is also quite a large change and quite a pain for people hosting to take the time to implement.

I've gone ahead and improved the download speeds in <https://github.com/coop-deluxe/sm64coopdx/pull/1426>. It's not perfect, so if you still think you can improve it once it's merged, feel free to (it's limitation is maintaining server fps while sending out files, finding a way to optimize that would lead to download speeds up to 67MB/s to be maintained. As it stands in the PR it's just put on a budget depending on the target fps), however it downloads plenty quick enough now.

Either way, I believe the current system in place can be improved and be used perfectly well, and I don't think we need to resort to a separate more convoluted system for improved download speeds.

PiantaMK:
I understand. I originally wrote this because i was tired of waiting 3+ mins when joining some servers for the first time for some reason (I have over 300mbps download speeds). I'm sure your PR has already pretty much resolved the issue, so I see no reason to try to implement this method as of now.

PR #1410, opened 2026-09-05
Document growing arrays

EmeraldLoc:
Documents growing arrays! I specifically documented initialization, allocating to the array, removing elements (swapping and popping), and cleanup by freeing the array

### DIFF
--- a/docs/c/sections/growing_arrays.md
+++ b/docs/c/sections/growing_arrays.md
@@ -1,0 +1,66 @@
+## [:rewind: C Reference](../c.md)
+
+# Growing Arrays
+
+Growing arrays are a custom type designed to handle dynamically scaling arrays in C. It allocates on the heap and only allocates what it needs, with no limit to how much it can allocate if necessary.
+
+This won't go over every function, but it will go over the most important ones. For a complete view on all the functions, head to `src/game/memory.c`, and head over to the growing array section (starts at line 193).
+
+## Setting Up
+
+Define a pointer to a `GrowingArray` struct:
+
+```c
+struct GrowingArray *gGrowingArrayExample = NULL;
+```
+
+You then need to find a spot to initialize this growing array. Look for an initialization function that works for the code you are modifying and throw the initialization for the array there.
+
+You can initialize a growing array using the function
+
+```c
+struct GrowingArray *growing_array_init(struct GrowingArray *array, u32 capacity, GrowingArrayAllocFunc alloc, GrowingArrayFreeFunc free);
+```
+
+The first param is the array it points to. If you ever need to reinitialize a growing array, you'd insert your `gGrowingArrayExample`. Technically, we don't need to do that for a first time initialization, so we could pass in `NULL`, however it's recommended to pass in your growing array for clarity.
+
+The `capacity` argument is the initial capacity. You generally want to avoid reallocations as reallocations are slow, so keep this number as a good average for how much capacity your array would typically need, but don't make it too large to effectively remove the use case of the growing array for 99% of users.
+
+Then there's an `alloc` function and a `free` function. In 99.999% of cases, you will pass in `malloc` for the allocation function and `free` for the free function. Note that while `malloc` usually doesn't zero-initialize the block of memory allocated, the growing array will zero-initialize the block, so all allocated memory is zero-initialized.
+
+As an example initialization:
+
+```c
+gGrowingArrayExample = growing_array_init(gGrowingArrayExample, 16, malloc, free);
+```
+
+## Allocating to the Growing Array
+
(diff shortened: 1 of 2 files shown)


docs/c/sections/growing_arrays.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Allocating to the Growing Array
+
+Each growing array has a `count` and the `buffer`. The `count` is the number of active elements in the `buffer`. The `count` does **not** represent the `capacity` of the array.
+
+To allocate data to the growing array, use the `growing_array_alloc` function.
+
+```c
+void *growing_array_alloc(struct GrowingArray *array, u32 size)
```
PeachyPeachSM64: missing ;

docs/c/sections/growing_arrays.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Cleanup
+
+If you ever need to cleanup a growing array and remove it entirely, use the `growing_array_free` function:
+
+```c
+void growing_array_free(struct GrowingArray **array);
+```
+
+This function is quite simple. It just takes in your growing array and frees every element inside it, every allocated pointer, and the passed in array itself. It also `NULL`'s the passed in array.
```
PeachyPeachSM64: It's important to note that you pass the address of the pointer to the growing array, not the growing array itself: here, it would be `&gGrowingArrayExample`

PeachyPeachSM64 (changes requested):
I don't mind not going into details for specific functions like `growing_array_sort` or `growing_array_bsearch`, but I think it would be good to have `growing_array_for_each_` documented for iterating over growing array elements.

EmeraldLoc:
> but I think it would be good to have growing_array_for_each_ documented for iterating over growing array elements.

I added it, and also went over standard for loop iteration because `growing_array_for_each_` isn't applicable in all cases (explained in docs)

PeachyPeachSM64 (approved):

docs/c/sections/growing_arrays.md
```diff
@@ -0,0 +1,107 @@
+## [:rewind: C Reference](../c.md)
+
+# Growing Arrays
+
+Growing arrays are a custom type designed to handle dynamically scaling arrays in C. It allocates on the heap and only allocates what it needs, with no limit to how much it can allocate if necessary.
```
Isaac0-dev: why "if necessary"? Makes it sound a little more confusing imo.

```suggestion
Growing arrays are a custom type designed to handle dynamically scaling arrays in C. It allocates on the heap and only allocates what it needs, with no limit to how much it can allocate.
```

docs/c/sections/growing_arrays.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Setting Up
+
+Define a pointer to a `GrowingArray` struct:
+
+```c
+struct GrowingArray *gGrowingArrayExample = NULL;
+```
+
+You then need to find a spot to initialize this growing array. Look for an initialization function that works for the code you are modifying and throw the initialization for the array there.
```
Isaac0-dev: "initialization function" is not perfectly clear. it could refer to an internal api function or their own code.
Perhaps:

```suggestion
You then need to find a spot to initialize this growing array. Look for an appropriate place in your code to init the array, and call `growing_array_init` there.
```

docs/c/sections/growing_arrays.md
```diff
@@ -0,0 +1,107 @@
...
+
+As an example initialization:
+
+```c
+gGrowingArrayExample = growing_array_init(gGrowingArrayExample, 16, malloc, free);
+```
+
+## Allocating to the Growing Array
+
+Each growing array has a `count` and the `buffer`. The `count` is the number of active elements in the `buffer`. The `count` does **not** represent the `capacity` of the array.
```
Isaac0-dev, line 39:
```suggestion
Each growing array has a `count` and the `buffer`. The `count` is the number of active/allocated elements in the `buffer`. The `count` does **not** represent the `capacity` of the array.
```

Isaac0-dev (commented):
here are some nitpick suggestions to improve clarity

docs/c/sections/growing_arrays.md
PeachyPeachSM64: "allocated" is kind of misleading here. The number of allocated elements (i.e. not NULL) is always greater or equal than count. "active" is perfectly fine here.

Isaac0-dev: fair enough. "allocated" is more intuitive but implies pointers are not held when popped. Ignore this suggestion.

Isaac0-dev (approved):

PR #1412, opened 2026-09-05
Allow autogen to generate enums from includes

PeachyPeachSM64:
Allow autogen to generate enums from includes

Define your enum in a separate file `my_enum.inl`:
```c
IDENTIFIER(NAME, value, ...)
```

Then, in `my_enum.h`:
```c
enum MyEnum {
#define IDENTIFIER(name, value, ...) name = value,
#include "my_enum.inl"
#undef IDENTIFIER
};
```

In `autogen/exposed_lists.py`:
```py
constants_enums_with_include = {
    "my_enum.h": {
        "MyEnum": {
            "IDENTIFIER": [index_of_name, index_of_value], # 1-indexed, 0 is IDENTIFIER
        }
    },
}
```

Bonus:
+ Remove self-links (useless)
+ Remove files that have no constant/function to document (wasting space)


autogen/convert_constants.py
```diff
@@ -63,19 +97,28 @@ def saw_constant(identifier, inIfBlock):
         seen_constants.append(identifier)
         return False
 
+def eval_constant(filename, constants, ident, val):
+    try:
+        set_to_val = int(eval(val, {}, {}))
```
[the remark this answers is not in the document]
PeachyPeachSM64: It was already done this way in `process_enum` before.
Besides, this tool is ran only by developers to generate documentation, it's not part of the actual game, and any change to the documentation system is verified before being pushed to avoid such issues.

EmeraldLoc (approved):
Someone else should probably review this since I'm awful at reading python, but it looks fine to me!

src/engine/graph_node_types.inl
```diff
@@ -0,0 +1,24 @@
+GRAPH_NODE_TYPE(GRAPH_NODE_TYPE_ROOT,                 0x001,                              GraphNodeRoot,                LOT_GRAPHNODEROOT)
```
Isaac0-dev, line 1: new way to access the root node fields via Lua. Not bad (since you could already via `geo_get_current_root`), just observing.
`GRAPH_NODE_TYPE_ROOT` used to map to `LOT_GRAPHNODE` in the old `graphNodeLots`.

src/engine/graph_node_types.inl
Isaac0-dev, line 1: `GRAPH_NODE_TYPE_FUNCTIONAL` missing. doesn't seem like a bad thing

Isaac0-dev (approved):

PR #1415, opened 2026-09-08
Re-issue the viewport and scissor when the 4:3 offsets change

the8thbit:
gfx_start_frame recalculates the 4:3 offsets every frame but never sets rdp.viewport_or_scissor_changed, and the offset comparisons in gfx_sp_tri1 sit inside that guard. So if the offsets change on a frame that doesn't emit a viewport or scissor command, the rendering API keeps the previous coordinates.

In practice G_MV_VIEWPORT calls gfx_calc_and_set_viewport at least once per frame, so the flag is normally set anyway and the worst case is a single odd frame after a window resize or a Force 4:3 toggle. This predates the letterbox change I worked on and came up in review there: https://github.com/coop-deluxe/sm64coopdx/pull/1414#discussion_r3954716658

This PR sets the flag at the end of gfx_start_frame. I set it unconditionally rather than only when the offsets differ, because the memcmp and previous-value checks in gfx_sp_tri1 already gate the actual work, so this costs one store per frame and covers both the x and y offsets without caching the previous values in a second place.

Rendering behavior is otherwise unchanged.

### DIFF
--- a/src/pc/gfx/gfx_pc.c
+++ b/src/pc/gfx/gfx_pc.c
@@ -2096,6 +2096,7 @@ void gfx_start_frame(void) {
     } else { gfx_current_dimensions.x_adjust_4by3 = 0; }
     gfx_current_dimensions.aspect_ratio = ((float)gfx_current_dimensions.width / (float)gfx_current_dimensions.height);
     gfx_current_dimensions.x_adjust_ratio = (4.0f / 3.0f) / gfx_current_dimensions.aspect_ratio;
+    rdp.viewport_or_scissor_changed = true;
 }
 
 void gfx_run(Gfx *commands) {


EmeraldLoc:
I don't think this is necessary, but someone may feel free to correct

If you look at `geo_process_root`, the *very* first displaylist command is updating the viewport already, which means frame 1 draw 1 the viewport will be updated, so when it comes time to draw, `rdp.viewport_or_scissor_changed ` will be set appropriately

the8thbit:
@EmeraldLoc 

> I don't think this is necessary, but someone may feel free to correct
> 
> If you look at `geo_process_root`, the _very_ first displaylist command is updating the viewport already, which means frame 1 draw 1 the viewport will be updated, so when it comes time to draw, `rdp.viewport_or_scissor_changed ` will be set appropriately

I don't have strong opinions on the PR, I was just doing what the bot was asking for, so feel free to reject! Really just cared about the other one, since I had to make that change to make my nucleus coop 2 player setup look good.

EmeraldLoc:
Alright, I'll close this then

PR #1417, opened 2026-09-10
fixes since the window manager merge + vsync input delay fix

Isaac0-dev:
- Fixed an input delay issue when VSync was enabled. I accidentally gave `shouldDelay` two meanings. It meant "not uncapped framerate, so frames should be timed in a grid-like pattern", and it also meant "enable our internal frame pacing".
VSync would have `shouldDelay` as false (because SDL does the frame pacing itself). This meant that we were using `curTime` in the calculation for the frame delta, so VSync always appeared to interpolate incorrectly ~50% of a frame behind for 60fps.
- After fixing this, it was discovered that a recently merged PR (specifically #1337) removed the code responsible for updating the VSync state in SDL. This was fixed, and I noticed one other issue with that PR which was worth fixing:
- `configGraphicsBackend` allowed setting the headless backend in the config file, and `gfxBackendChoices` misleadingly used an unrelated constant that happened to have the correct value as the buffer size.

### DIFF
--- a/src/pc/gfx/gfx_window_manager.c
+++ b/src/pc/gfx/gfx_window_manager.c
@@ -27,7 +27,6 @@ static struct GfxWindowBackendAPI *sBackends[GFX_WINDOW_BACKEND_COUNT] = {
 #if defined(_WIN32)
     [GFX_WINDOW_BACKEND_DIRECTX] = &gfx_window_dxgi,
 #endif
-    [GFX_WINDOW_BACKEND_DUMMY] = &gfx_window_dummy,
 };
 
 // TODO: figure out how to switch the backend without restarting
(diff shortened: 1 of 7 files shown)


src/pc/gfx/gfx_window_manager.c
```diff
@@ -112,7 +117,12 @@ void gfx_wm_init(const char *window_title) {
 #else
     currBackend = configGraphicsBackend;
 #endif
-    sBackends[currBackend]->init(window_title);
+    if (currBackend != GFX_WINDOW_BACKEND_DUMMY &&
+        (currBackend < GFX_WINDOW_BACKEND_OPENGL || currBackend > GFX_WINDOW_BACKEND_MAX)
```
EmeraldLoc: Should probably be `currBackend < 0`. If the order ever changes for render apis, this code will break. Using `< 0` guarantees that it always works

EmeraldLoc (changes requested):

src/pc/gfx/gfx_window_opengl.c
```diff
@@ -135,6 +136,7 @@ static void gfx_window_opengl_init(const char *window_title) {
 
     gfx_wm_set_window(sSdlWindow);
     gfx_window_opengl_set_vsync(configWindow.vsync);
+    sAppliedVsync = configWindow.vsync;
```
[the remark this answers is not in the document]
Isaac0-dev: good suggestion, `gfx_window_opengl_reset_dimension_and_pos` is intentionally not supposed to update the cache.

src/pc/gfx/gfx_window_manager.c
Isaac0-dev: good point, thanks

EmeraldLoc (approved):

PR #1420, opened 2026-09-12
Improve chat tab completion and input feedback

iZePlayzYT:
### Tab completion

- Make command, subcommand, and player-name matching case-insensitive
- Complete commands using their canonical casing
- Match and sort player names without color codes
- Preserve player-name formatting when applying completions
- Add player-name previews without formatting
- Suggest and cycle through players after `@`
- Restore the original input when pressing Backspace after cycling completions

### Input feedback

- Show incomplete commands and subcommands in orange
- Show incorrectly cased commands and subcommands in yellow
- Show unknown commands and subcommands in red
- Render visible color codes using the color they represent
- Keep the actual player name in the normal input color

### DIFF
--- a/src/pc/djui/djui_chat_box.c
+++ b/src/pc/djui/djui_chat_box.c
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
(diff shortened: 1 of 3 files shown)


iZePlayzYT:
Could you now review please @EmeraldLoc

PR #1424, opened 2026-09-14
Increase MAX_MOD_SIZE from 35MB to 512MB

iZePlayzYT:
Allow bigger Mods (A lot of romhacks converted with https://isaac0-dev.github.io/rom-decomp-64/ exceed the current limited)


EmeraldLoc:
No. Download times will take 9 trillion years. Optimize your mod to fit within range. As it stands, this will not be increased. It will be increased when the devs decide to as it's been discussed multiple times (with all times it being concluded that the limit should not be raised)

iZePlayzYT:
@EmeraldLoc i meant it for singleplayer stuff especially...

EmeraldLoc:
That's fine, but you'd need to separate it out, not raise the limit on everything

Isaac0-dev:
this is a bug in rom-decomp-64, so it should be fixed there. This project shouldn't need to account for my bad code lol

PR #1425, opened 2026-09-15
Implement delta time

georgegabrdev:
This adds delta time by using HOOK_UPDATE_DELTA_TIME and HOOK_ON_RENDER_DELTA_TIME.
Good for timers and other stuff


Cooliokid956:
this... doesn't add anything that you can't already do in the game? you can just use `clock_elapsed_f64` in Lua anyway, and I'm fairly certain the render hook- oh. Either way it still doesn't look like there's a use case for this hook... mind adding a test mod and providing use cases?

EmilyEmmi:
Assuming the render hook works correctly, it would allow rendering at higher framerates, which might be useful. The HOOK_UPDATE_DELTA_TIME hook doesn't have a use case atm since you can just call calculate that time manually.

Not sure if allowing lua calls at higher framerates should be possible... I recall a similar PR that was rejected for related concerns?

georgegabrdev:
I was too lazy to put a example but ok
```lua
local t = 0

function render_delta(dt)
    t = t + dt
end

hook_event(HOOK_ON_RENDER_DELTA_TIME, render_delta)
hook_event(HOOK_UPDATE_DELTA_TIME, render_delta) -- you can do the same thing, its just at 30 fps but useful for stuff

function render()
    djui_hud_print_text(string.format("%.3f", t), 20, 20, 1) -- timer example
end

hook_event(HOOK_ON_HUD_RENDER, render)
```

PeachyPeachSM64:
> I was too lazy to put a example but ok
> 
> ```lua
> local t = 0
> 
> function render_delta(dt)
>     t = t + dt
> end
> 
> hook_event(HOOK_ON_RENDER_DELTA_TIME, render_delta)
> hook_event(HOOK_UPDATE_DELTA_TIME, render_delta) -- you can do the same thing, its just at 30 fps but useful for stuff
> 
> function render()
>     djui_hud_print_text(string.format("%.3f", t), 20, 20, 1) -- timer example
> end
> 
> hook_event(HOOK_ON_HUD_RENDER, render)
> ```

So... you're trying to add something you have no need for?

Anyway, as Coolio and Emily mentioned, `HOOK_UPDATE_DELTA_TIME` is just `HOOK_UPDATE` with a parameter you can already calculate with the `clock_elapsed` functions.

I can see the need for `HOOK_ON_RENDER_DELTA_TIME`, but it was really a missed opportunity to pass something useful like `delta` instead of an (inaccurate) delta time we can already compute with (again) the `clock_elapsed` functions.

PR #1432, opened 2026-09-20
Make djui_chat_box.c not override Home/End key behavior

FluffyKn1ght:
pageup/pagedown literally do the exact same thing but with a config setting check for how much to scroll. as a seasoned home/end key user it is REALLY inconvenient to press home for it to scroll the history instead of going to the start of the chatbox.


FluffyKn1ght:
oopsie daisy i made this pr with the base set to main instead of dev-

PeachyPeachSM64 (approved):

EmeraldLoc (approved):

